PR #13, opened 2026-03-02
Fix overwrites in parallel file parsing

ChouraquiBen:
## Motivation

Two errors could be encountered:
- Scan crash because of a race condition
- Each different scan on the same folder could lead to a different number of vulnerabilities 

This is all because of shared variables across threads that should not be.

## Changes

- `ResolvedFiles` is now processed per file for Yaml and Json files
- Each thread ignores comments using its own ignore structure
- Terraform files' `InputVars` is now per `fullPath` to avoid issues on files having the same name

## Author Checklist

## QA Instruction

- [x] No errors on the ci
- [x] The scans on a single repo should always give the same number of findings

## Blast Radius

`iac-scanning`

## Additional Notes

### DIFF
--- a/pkg/model/comment_yaml.go
+++ b/pkg/model/comment_yaml.go
@@ -8,7 +8,6 @@ package model
 import (
 	"reflect"
 	"strings"
-	"sync"
 
 	"gopkg.in/yaml.v3"
 )
@@ -20,20 +19,15 @@ type comment string
 type Ignore struct {
 	// Lines is the lines to ignore
 	Lines []int
-	mu    sync.Mutex
 }
 
 // build builds the ignore struct
 func (i *Ignore) build(lines []int) {
-	i.mu.Lock()
-	defer i.mu.Unlock()
 	i.Lines = append(i.Lines, lines...)
 }
 
 // GetLines returns the lines to ignore
 func (i *Ignore) GetLines() []int {
-	i.mu.Lock()
-	defer i.mu.Unlock()
 	return RemoveDuplicates(i.Lines)
 }
 
--- a/pkg/model/comment_yaml.go
+++ b/pkg/model/comment_yaml.go
@@ -23,10 +23,6 @@ type Ignore struct {
 	mu    sync.Mutex
 }
 
-var (
-	NewIgnore = &Ignore{}
-)
-
 // build builds the ignore struct
 func (i *Ignore) build(lines []int) {
 	i.mu.Lock()
@@ -39,13 +35,6 @@ func (i *Ignore) GetLines() []int {
 	i.mu.Lock()
 	defer i.mu.Unlock()
 	return RemoveDuplicates(i.Lines)
-}
-
-// Reset resets the ignore struct
-func (i *Ignore) Reset() {
-	i.mu.Lock()
-	defer i.mu.Unlock()
-	i.Lines = make([]int, 0)
 }
 
 // ignoreCommentsYAML sets the lines to ignore for a yaml file
--- a/pkg/model/comment_yaml.go
+++ b/pkg/model/comment_yaml.go
@@ -20,41 +20,41 @@ type comment string
 type Ignore struct {
 	// Lines is the lines to ignore
 	Lines []int
+	mu    sync.Mutex
 }
 
 var (
 	NewIgnore = &Ignore{}
-	mu        sync.Mutex
 )
 
 // build builds the ignore struct
 func (i *Ignore) build(lines []int) {
-	mu.Lock()
-	defer mu.Unlock()
+	i.mu.Lock()
+	defer i.mu.Unlock()
 	i.Lines = append(i.Lines, lines...)
 }
 
 // GetLines returns the lines to ignore
 func (i *Ignore) GetLines() []int {
-	mu.Lock()
-	defer mu.Unlock()
+	i.mu.Lock()
+	defer i.mu.Unlock()
 	return RemoveDuplicates(i.Lines)
 }
 
 // Reset resets the ignore struct
 func (i *Ignore) Reset() {
-	mu.Lock()
-	defer mu.Unlock()
+	i.mu.Lock()
+	defer i.mu.Unlock()
 	i.Lines = make([]int, 0)
 }
 
 // ignoreCommentsYAML sets the lines to ignore for a yaml file
-func ignoreCommentsYAML(node *yaml.Node) {
+func (ignore *Ignore) ignoreCommentsYAML(node *yaml.Node) {
 	linesIgnore := make([]int, 0)
 	if node.HeadComment != "" {
 		// Squence Node - Head Comment comes in root node
 		linesIgnore = append(linesIgnore, processCommentYAML((*comment)(&node.HeadComment), 0, node, node.Kind, false)...)
-		NewIgnore.build(linesIgnore)
+		ignore.build(linesIgnore)
 		return
 	}
 	// check if comment is in the content
@@ -68,7 +68,7 @@ func ignoreCommentsYAML(node *yaml.Node) {
 		linesIgnore = append(linesIgnore, processCommentYAML((*comment)(&content.HeadComment), i, node, node.Kind, false)...)
 	}
 
-	NewIgnore.build(linesIgnore)
+	ignore.build(linesIgnore)
 }
 
 // processCommentYAML returns the lines to ignore
--- a/pkg/model/comment_yaml.go
+++ b/pkg/model/comment_yaml.go
@@ -6,7 +6,6 @@
 package model
 
 import (
-	"context"
 	"reflect"
 	"strings"
 	"sync"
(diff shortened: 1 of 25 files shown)


pkg/parser/ansible/ini/config/parser.go
```diff
@@ -25,8 +25,6 @@ func (p *Parser) Resolve(ctx context.Context, fileContent []byte, _ string, _ bo
 
 // Parse parses .cfg/.conf file and returns it as a Document
 func (p *Parser) Parse(ctx context.Context, filePath string, fileContent []byte) ([]model.Document, []int, error) {
-	model.NewIgnore.Reset()
```
ChouraquiBen, line 28: No need to reset since the state is not shared

pkg/model/comment_yaml.go
```diff
@@ -20,42 +21,59 @@ type comment string
...
 )
 
+func getIgnoreFromContext(ctx context.Context) *Ignore {
+	if ctx == nil {
+		return NewIgnore
+	}
+	if ignore, ok := ctx.Value(ContextKeyIgnore{}).(*Ignore); ok {
+		return ignore
+	}
+	return NewIgnore
```
ChouraquiBen: This is a global shared state used as fallback. This explains why the mutex are still needed

jacobotb:
Using the `context` to pass objects around is smelly (I'm talking about the `Ignore` in `pkg/model/comment_yaml.go`.)

It looks to me like, in the original code, each `parse` function essentially clears `NewIgnore` and then populates and uses it. So, in effect, each 'parse' function starts with a "new" ignore object, so we should be able to remove `NewIgnore` completely and, instead, create and pass `Ignore` objects explicitly and not in the `context`. This should make the code easier to follow.

pkg/parser/json/parser.go
```diff
@@ -20,7 +20,7 @@ import (
 type Parser struct {
```
jacobotb, line 19: Is `pkg.parser.json.Parser` used anywhere?

ChouraquiBen: The json parser is not used anywhere currently. This has been made on purpose so that we don't start scanning terraform plans, CDK and Cloudformation CDK when this hasn't been tested yet

ChouraquiBen:
> It looks to me like, in the original code, each `parse` function essentially clears `NewIgnore` and then populates and uses it.

I did not know it was not good practice, thanks for telling me. I am curious to know why if you have more informations.

In the mean time, I agree with ignores in each parser. I previously tried to create a new parser each time in order to fix the issue of inconsistent vulnerability number, in vain.

This time, knowing that there was also the NewIgnore that was to be distributed, I tried and it seems to work.

I tried, enabling every single platform that our scanner aims to support in the end along with adding the json parser. The number of vulnerability seems consistent accross the runs now 🙏

ChouraquiBen:
Tests are failing because pkg/model/model_yaml_test.go:646:31 does not have enough arguments to call UnmarshalYAML. This file is not that long and actually, the only call to UnmarshalYAML has the said arguments
edit: this was because of edits in the main branch, it is solved after merging

pkg/model/comment_yaml.go
```diff
@@ -20,42 +20,41 @@ type comment string
 type Ignore struct {
 	// Lines is the lines to ignore
 	Lines []int
+	mu    sync.Mutex
 }
 
 var (
-	// NewIgnore is the ignore struct
 	NewIgnore = &Ignore{}
```
jacobotb: Ok, there are two things here.

The first is the `NewIgnore` problem. The original code has a global `NewIgnore` object and the rest of the code reuses it, resetting it when needed. This is not good code; this is the kind of thing you only do when you are desperate for performance and you have exhausted _everything_ and you are willing to mess up your source code in exchange for one or two CPU cycles.

The problem in this case is that we got the "mess up your source code" part but not the "get one or two CPU cycles in exchange".

This code would have made a little sense if creating an `Ignore` instance had high overhead, but it doesn't. The `Ignore` struct only contains a slice, so the only overhead would be in allocating that slice. _But_ the `Reset` method works by allocating a new slice and replacing the old one with the newly allocated one, so keeping a global `Ignore` instance and reusing it and calling `Reset` between uses produces absolutely _zero_ savings compared to creating a new `Ignore` instance when needed, at the cost of making the code more fragile and harder to maintain.

So, what you need to do is remove the `NewIgnore` variable and the `Reset` method. We are not going to have a global `Ignore` instance; instead, we are going to create a new one whenever `Reset` would have been called.

_Except..._

The second thing is that the `Ignore` instance is only ever used in `pkg/parser/yaml/parser.go`. The reason I know that is because I see that it's the only place where `Ignore.GetLines()` is called. There are other functions that call `Ignore.Reset()` but they don't call `Ignore.GetLines()`, so at the end of the day they haven't done anything.

So we can remove the `NewIgnore.Reset()` calls everywhere except in `pkg/parser/yaml/parser.go:Parser.Parse()`, where we replace it with something like `ignore := model.Ignore{}`.

Then we have to update `doc.UnmarshalYAML` to take a `*model.Ignore` (you already did this in your PR) and do a nil check in `unmarshalWithDepth()` and do nothing if `ignore==nil` (currently you assign `NewIgnore` to it, but that's gone).

And, of course, in the other usage of `UnmarshalYAML` in `pkg/analyzer/analyzer.go`, we have to pass `nil` for the `*model.Ignore` parameter.

pkg/parser/ansible/ini/hosts/parser.go
```diff
@@ -142,3 +142,9 @@ func (p *Parser) StringifyContent(content []byte) (string, error) {
 func emptyDocument() *model.Document {
 	return &model.Document{}
 }
+
+func (p *Parser) Clone() any {
```
jacobotb: I don't think you need this `Clone` method since `ignore` is not really used in this parser, and it is only used per call to `Parse` in the YAML parser, so it doesn't need to be a member.

pkg/model/comment_yaml.go
ChouraquiBen: Thanks a lot for this detailed and clear explanation, creating a new instance of ignore seems indeed to be the cleanest way to go, solving all the issues creating only minimal changes. I do feel a bit dumb, having tried to create something more complex 😓

jacobotb: I've done way worse, and that was last year :-)

pkg/parser/ansible/ini/hosts/parser.go
ChouraquiBen: The scan.go file adds both ansible parsers:
```
		Add(&ansibleConfigParser.Parser{}).
		Add(&ansibleHostsParser.Parser{}).
```
They need to be `kindParser` and I added this `Clone` function to the interface, therefore the `Clone` function here as well

pkg/model/model_yaml.go
```diff
@@ -40,27 +40,30 @@ func (m *Document) UnmarshalYAML(ctx context.Context, value *yaml.Node) error {
...
 		return nil
 	}
 	visited[val] = true
 	defer func() { delete(visited, val) }()
 	tmp := make(map[string]interface{})
-	ignoreCommentsYAML(val)
+	if ignore == nil {
+		ignore = &Ignore{}
+	}
+	ignore.ignoreCommentsYAML(val)
```
jacobotb:
```suggestion
	if ignore != nil {
		ignore.ignoreCommentsYAML(val)
	}
```

pkg/parser/ansible/ini/config/parser.go
```diff
@@ -17,6 +17,7 @@ import (
 
 // Parser defines a parser type
 type Parser struct {
+	ignore *model.Ignore
```
jacobotb: As far as I can tell, this is not used (`GetLines()` is not called on it), so you can remove it.

pkg/parser/ansible/ini/hosts/parser.go
jacobotb: Ok, but why did you add the `Clone` function?

Answering myself: because the yaml parser object is stateful; it contains a `ResolvedFiles` map so whenever we use a parser in `pkg/parser/parser.go:Parse` we want a fresh object that won't share its state with any other parser object.

I don't think you need if if you do a little change in the code. Quite often, when you are trying to parallelize code, the best solution when you find shared state is to remove it, not to try to manage it.

The shared state in question is the `resolvedFiles` map, which is only touched by `Resolve()` and `GetResolvedFiles()`, so let's see how they work:

The `pkg/parser/parser.go:Parse` function calls `Resolve`. This function creates a resolver and calls `res.Resolve()` on it, and then overwrites `p.resolvedFiles` with the content of `res.ResolvedFiles` and returns the "resolved file contents".

Then `pkg/parser/parser.go` calls `p.Parse()` (which does _not_ look at `p.resolvedFiles`) and finally calls `GetResolvedFiles()` to get the resolved files that had been previously returned by the resolver.

So you can simplify this by removing `Parser.resolvedFiles` (and its mutex) and `Parser.GetResolvedFiles`, and changing `Parser.Resolve()` to return three values: the `[]byte` it currently returns, the content of `res.ResolvedFiles`, and the error.

Then you can change `pkg/parser/parser.go:Parse` to do:

```golang
resolved, resolvedFiles, err := c.parsers.Resolve(...)
// ...
return ParsedDocument {
  // ...
  ResolvedFiles: resolvedFiles,
```

With this, the parsers will be stateless and you won't need a `Clone()` method because you can reuse the same parser over and over.

ChouraquiBen: Thanks a lot for this explanation, I made the necessary changes and I feel a lot more confident with my understanding of good code related to parallelisation 🙏

pkg/parser/ansible/ini/hosts/parser.go
```diff
@@ -19,18 +19,18 @@ import (
...
+	resolved []byte,
+	documents []model.Document,
+	ignoreLines []int,
+	resolvedFiles map[string]model.ResolvedFile,
+	error error) {
 	inventoryReader := strings.NewReader(string(fileContent))
 	var inventory, err = aini.Parse(inventoryReader)
 	if err != nil {
-		return nil, nil, err
+		return []byte{}, nil, []int{}, map[string]model.ResolvedFile{}, err
```
jacobotb: Slices and maps are nilable, so you can use `nil` for them when you are returning an error:

```suggestion
		return nil, nil, nil, nil, err
```

And here is an inconsistency in Go: a `nil` slice is a valid slice with length 0 (so you can return it in the "not error" case if you are returning an empty slice), but a `nil` map is not a valid map (so you still have to initialize it in the "not error" case if you are returning an empty map.)

pkg/parser/parser.go
```diff
@@ -22,10 +22,8 @@ type kindParser interface {
 	GetCommentToken() string
 	SupportedExtensions() []string
 	SupportedTypes() map[string]bool
-	Parse(ctx context.Context, filePath string, fileContent []byte) ([]model.Document, []int, error)
-	Resolve(ctx context.Context, fileContent []byte, filename string, _ bool, _ int) ([]byte, error)
+	Parse(ctx context.Context, fileContent []byte, filename string, _ bool, _ int) ([]byte, []model.Document, []int, map[string]model.ResolvedFile, error)
```
jacobotb: I would have left `Parse` and `Resolve` as two different functions because then each function only would do one thing and their interfaces would be simpler. It doesn't matter if one of the functions is trivial in 4 out of 5 implementations, as the alternative is to have 2 unused arguments and 2 trivial return values in 4 out of 5 implementations...

But it's done now and I've had you making changes for a week, so let's merge this so we can keep going.

jacobotb (dismissed):

pkg/parser/ansible/ini/hosts/parser.go
ChouraquiBen: Very good to know, I'll change this right now!

jacobotb (approved):

PR #14, opened 2026-03-02
Implement hclsyntax.RelativeTraversalExpr in HCL parser

whitemerch:
## Motivation

Logs indicate that `hclsyntax.RelativeTraversalExpr` is not implemented in the parser, causing expressions like `jsondecode(var.config).host` or `list[var.i].name` to be unhandled across multiple dispatch sites.

## Changes

- Added `RelativeTraversalExpr` case to `ExpToString` in `pkg/builder/engine/engine.go`
- Added `RelativeTraversalExpr` case to `expressionToAST` in `pkg/engine/inspector.go`
- Added `RelativeTraversalExpr` case to `convertExpression` and `convertStringPart` in `pkg/parser/terraform/converter/default.go`
- Added `RelativeTraversalExpr` case to `resolveExpr` in `pkg/parser/terraform/modules/modules.go`
- Added 14 test cases covering happy paths, error propagation, template integration, and index traversal steps

## Author Checklist

## QA Instruction

Run the test suites for the modified packages:
```
go test ./pkg/builder/engine/ ./pkg/engine/ ./pkg/parser/terraform/converter/ ./pkg/parser/terraform/modules/ -v -count=1
```
Verify all `RelativeTraversalExpr` tests pass and no regressions in existing tests.

## Blast Radius

DataDog IaC Scanner only, HCL expression parsing/evaluation

## Additional Notes

### DIFF
--- a/pkg/engine/inspector.go
+++ b/pkg/engine/inspector.go
@@ -47,6 +47,7 @@ const (
 	DefaultQueryDescriptionID   = "Undefined"
 	DefaultQueryURI             = "https://github.com/DataDog/datadog-iac-scanner/"
 	DefaultIssueType            = model.IssueTypeIncorrectValue
+	unresolvedPlaceholder       = "__UNRESOLVED__"
 
 	regoQuery = `result = data.Cx.CxPolicy`
 )
@@ -789,7 +790,7 @@ func expressionToAST(expr hclsyntax.Expression) (ast.Value, error) { //nolint:go
 		for _, item := range e.Exprs {
 			v, err := expressionToAST(item)
 			if err != nil {
-				v = ast.String("__UNRESOLVED__")
+				v = ast.String(unresolvedPlaceholder)
 			}
 			terms = append(terms, ast.NewTerm(v))
 		}
@@ -809,7 +810,7 @@ func expressionToAST(expr hclsyntax.Expression) (ast.Value, error) { //nolint:go
 			}
 			valVal, err := expressionToAST(item.ValueExpr)
 			if err != nil {
-				valVal = ast.String("__UNRESOLVED__")
+				valVal = ast.String(unresolvedPlaceholder)
 			}
 			obj.Insert(ast.NewTerm(strKey), ast.NewTerm(valVal))
 		}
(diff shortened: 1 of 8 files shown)


pkg/engine/inspector.go
```diff
@@ -822,20 +823,41 @@ func expressionToAST(expr hclsyntax.Expression) (ast.Value, error) { //nolint:go
...
+			return ast.String(unresolvedPlaceholder), nil
 		}
 		collStr := astValueToSimpleString(collV)
 		keyStr := astValueToSimpleString(keyV)
 		return ast.String(collStr + "[" + keyStr + "]"), nil
 
+	case *hclsyntax.RelativeTraversalExpr:
+		sourceVal, err := expressionToAST(e.Source)
+		if err != nil {
+			return ast.String(unresolvedPlaceholder), nil
```
ChouraquiBen, line 835: Maybe we should log a warn in case this is not resolved

pkg/parser/terraform/modules/modules.go
```diff
@@ -241,15 +246,43 @@ func resolveExpr(expr hclsyntax.Expression, locals, vars map[string]string) stri
...
+		return resolveExprDefault(expr)
+	}
+}
+
+func resolveRelativeTraversalExpr(e *hclsyntax.RelativeTraversalExpr, locals, vars map[string]string) string {
+	sourceStr := resolveExpr(e.Source, locals, vars)
+	if strings.HasPrefix(sourceStr, "__") {
+		return unresolvedPlaceholder
+	}
+	for _, step := range e.Traversal {
```
ChouraquiBen, line 262: Since it is used more than once, would it be useful to add it somewhere in the utils pkg?

ChouraquiBen (approved):
Thanks a lot for the change!

pkg/engine/inspector.go
whitemerch: Good point, however `expressionToAST` is a context-free function with no access to a logger from what I saw. The existing error-recovery paths in the same function (TupleConsExp, IndexExpr...) all silently return `unresolvedPlaceholder` without logging. I'd prefer to address observability for all paths in a follow-up rather than creating an asymmetry

pkg/parser/terraform/modules/modules.go
whitemerch: Again agreed this is duplicated, the same loop exists in engine.go as relativeTraversalToString, inspector.go, and modules.go. I'll track this for a cleanup pass that consolidates all the shared HCL traversal helpers (this loop, unresolvedPlaceholder const, etc.) into a proper shared package, because i dont want to mix utils generic instances with hcl specific utils

PR #15, opened 2026-03-03
Support non-decimal integers in YAML parser

whitemerch:
## Motivation

YAML scalar values tagged as `!!int` were parsed with `strconv.Atoi`, which only accepts base-10. Values in hex (`0x1A`), octal (`0o755`), binary (`0b1010`), or with underscores (`1_000_000`), which can happen with YAML, failed and were left as strings, causing repeated "failed to convert integer in yaml parser" errors when scanning IaC that uses those formats.

## Changes

- Replaced manual `!!int` / `!!bool` / `!!null` handling in `scalarNodeResolver` with a single call to `val.Decode(&resolved)`, delegating to the YAML library. This handles all integer formats (hex, octal, binary, underscores), overflow (int64 when out of int range), floats, bools, and null. On decode error, log the failing value and return the raw string.
- Added tests for scalar integer formats (hex, octal, binary, underscores), float, bool/null, and an end-to-end test that parses YAML bytes

## Author Checklist

## QA Instruction

- Run `go test ./pkg/model/... -v -run TestDocument_UnmarshalYAML` and `TestDocument_UnmarshalYAML_FromBytes`. All should pass.
- Optionally scan a YAML file that uses e.g. `port: 0x1A` or `mode: 0o755` and confirm no "failed to convert integer" logs and that values are parsed as numbers.

## Blast Radius

IaC Scanner YAML parsing only

### DIFF
--- a/pkg/model/model_yaml.go
+++ b/pkg/model/model_yaml.go
@@ -9,7 +9,6 @@ import (
 	"context"
 	json "encoding/json"
 	"errors"
-	"strconv"
 
 	"github.com/DataDog/datadog-iac-scanner/pkg/logger"
 	"github.com/DataDog/datadog-iac-scanner/pkg/utils"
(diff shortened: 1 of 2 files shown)


MikaYuoadas (approved):
LGTM, also far simpler 👍

PR #26, opened 2026-03-09
HCL expressions central package refactor

whitemerch:
> [!TIP]

## Motivation

Adding or changing an HCL expression type required updating four separate type switches (engine, inspector, modules, converter), which caused drift and bugs. This PR introduces a single `Visitor` + `Dispatch` and central `Unwrap` so all expression handling goes through one place and stays in sync.

## Changes

- Add `pkg/hclexpr`: `Visitor[T]` interface, `Dispatch(expr, visitor)`, and `Unwrap(expr)` for TemplateWrapExpr, ParenthesesExpr, ObjectConsKeyExpr.
- Migrate engine, inspector, modules, and converter to use `hclexpr.Dispatch` + Unwrap instead of local type switches.
- Add ParenthesesExpr unwrap in converter `convertExpression`.
- Fix ScopeTraversalExpr in inspector to return full traversal path (not only root name).

## Author Checklist

- [ x] All new and existing tests pass.

## QA Instruction

Run `go test ./pkg/hclexpr/... ./pkg/builder/engine/... ./pkg/engine/... ./pkg/parser/terraform/...`. Terraform scans with expressions (traversals, templates, index, conditional, etc.) should behave as before; inspector findings that depend on full traversal paths should be correct.

## Blast Radius

DataDog IaC Scanner only (HCL expression handling in engine, inspector, modules, converter).

## Additional Notes

Stacked on top of `chakib.hamie/hcl_expr_consistency_and_refactor`. Refactor only.

### DIFF
--- a/pkg/builder/engine/engine.go
+++ b/pkg/builder/engine/engine.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/DataDog/datadog-iac-scanner/pkg/hclexpr"
 	"github.com/DataDog/datadog-iac-scanner/pkg/logger"
 	"github.com/hashicorp/hcl/v2"
 	"github.com/hashicorp/hcl/v2/hclsyntax"
@@ -24,17 +25,12 @@ type Engine struct {
 // ExpToString converts an expression into a string
 func (e *Engine) ExpToString(ctx context.Context, expr hclsyntax.Expression) (string, error) {
 	contextLogger := logger.FromContext(ctx)
+	expr = hclexpr.Unwrap(expr)
 	switch t := expr.(type) {
 	case *hclsyntax.LiteralValueExpr:
 		return e.expToStringLiteralValue(t)
 	case *hclsyntax.TemplateExpr:
 		return e.expToStringTemplateExpr(ctx, t)
-	case *hclsyntax.TemplateWrapExpr:
-		return e.ExpToString(ctx, t.Wrapped)
-	case *hclsyntax.ParenthesesExpr:
-		return e.ExpToString(ctx, t.Expression)
-	case *hclsyntax.ObjectConsKeyExpr:
-		return e.ExpToString(ctx, t.Wrapped)
 	case *hclsyntax.ScopeTraversalExpr:
 		return e.expToStringScopeTraversal(t), nil
 	case *hclsyntax.IndexExpr:
--- a/pkg/builder/engine/engine.go
+++ b/pkg/builder/engine/engine.go
@@ -24,31 +24,49 @@ type Engine struct {
 
 // ExpToString converts an expression into a string
 func (e *Engine) ExpToString(ctx context.Context, expr hclsyntax.Expression) (string, error) {
-	contextLogger := logger.FromContext(ctx)
-	expr = hclexpr.Unwrap(expr)
-	switch t := expr.(type) {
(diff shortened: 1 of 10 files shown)


pkg/builder/engine/engine.go
whitemerch, line 1: Overall, curious to have thoughts here on this approach. It's a tradeoff between safety, but having to redefine the methods each time for each implementation sooo

pkg/hclexpr/visitor_test.go
```diff
@@ -0,0 +1,164 @@
+/*
```
MikaYuoadas, line 1: This file is named dispatch_test.go, shouldn’t it be "visitor_test.go" instead?

pkg/hclexpr/visitor_test.go
```diff
@@ -0,0 +1,164 @@
...
+	}{
+		{"LiteralValue", `42`, "LiteralValue"},
+		{"ScopeTraversal", `var.x`, "ScopeTraversal"},
+		{"TemplateExpr", `"hello ${var.x} world"`, "TemplateExpr"},
+		{"IndexExpr", `list[var.i]`, "IndexExpr"},
+		{"RelativeTraversal", `list[var.i].name`, "RelativeTraversal"},
+		{"FunctionCall", `upper("x")`, "FunctionCall"},
+		{"Conditional", `true ? 1 : 2`, "Conditional"},
+		{"TupleCons", `[1, 2]`, "TupleCons"},
+		{"ObjectCons", `{a = 1}`, "ObjectCons"},
```
MikaYuoadas, line 88: There is no test case for TemplateJoin, is it an oversight?

MikaYuoadas (commented):
LGTM 👍 

Left a few nitpick comments, but otherwise I really like the visitor pattern use here. Even though it’s indeed more verbose I think having the compiler checking that none of the implementation is missing is worth it and will pay dividend in the future.

Another small nitpick on the PR itself: the TraverseIndex fix should have been squashed with the first commit about ScopeTraversal and not bundled with the Visitor + Dispatch commit as they aren’t related. (don’t bother rebasing this PR, just pointing this out as best practice for future PRs)

MikaYuoadas (dismissed):
Now it’s really approved 😬

pkg/hclexpr/visitor_test.go
whitemerch: Indeed

pkg/parser/terraform/modules/modules.go
```diff
@@ -64,12 +64,13 @@ func isTerraformFile(filePath string) bool {
...
-	stringRegistry        = "registry"
-	stringPrivate         = "private"
-	unresolvedPlaceholder = "__UNRESOLVED__"
+	stringLocal                 = "local"
+	stringUnknown               = "unknown"
+	stringPublic                = "public"
+	stringRegistry              = "registry"
+	stringPrivate               = "private"
+	unresolvedPlaceholder       = "__UNRESOLVED__"
+	invalidTraversalPlaceholder = "__INVALID_TRAVERSAL__"
```
whitemerch, line 73: go ci was failing, I don't know it started after the rebase though??

MikaYuoadas: 🪄

MikaYuoadas (approved):

PR #32, opened 2026-03-10
Ansible normalize resourceType

whitemerch:
## Motivation

Ansible rules reported inconsistent or wrong `resourceType` values: (1) risky_file_permissions used the map value (boolean) instead of the module name; (2) rules output either FQCN (e.g. `community.aws.ec2_instance`) or short name (`ec2_instance`) depending on the set. We fix the bug and normalize Ansible resourceType in Go so all findings expose the content name (last segment after `.`) without changing every rego rule.

## Changes

- **Fix risky_file_permissions:** Third CxPolicy block used `resourceType: modules[m]` (map value = boolean), which means it reports true/false for the resourceType haha. Now uses `m` (module name).
- **Normalize Ansible resourceType in Go:** Added `utils.NormalizeAnsibleResourceType(resourceType)` (returns last segment after `.`, or unchanged if no dot). In `vulnerability_builder.go`, when platform is Ansible we normalize before setting `ResourceType`. Rego rules are unchanged; normalization is applied once at build time. Includes unit tests for the util.

## Author Checklist

## QA Instruction

Run scanner on Ansible playbooks; confirm findings show `resourceType` as short names (e.g. `ec2_instance`, `user`, `ansible_config`) not FQCNs. Run `go test ./pkg/utils/ -run TestNormalizeAnsibleResourceType` and `go test ./pkg/engine/...`.

## Blast Radius

Ansible findings only: `ResourceType` field in vulnerability model (and thus in SARIF, storage, API). No rego changes; no other platforms affected.

## Additional Notes

### DIFF
--- a/assets/queries/ansible/general/risky_file_permissions/query.rego
+++ b/assets/queries/ansible/general/risky_file_permissions/query.rego
@@ -73,7 +73,7 @@ CxPolicy[result] {
 
 	result := {
 		"documentId": id,
-		"resourceType": modules[m],
+		"resourceType": m,
 		"resourceName": task.name,
 		"searchKey": sprintf("name={{%s}}.{{%s}}", [task.name, m]),
 		"issueType": "IncorrectValue",
--- a/pkg/engine/vulnerability_builder.go
+++ b/pkg/engine/vulnerability_builder.go
@@ -17,6 +17,7 @@ import (
 	"github.com/DataDog/datadog-iac-scanner/pkg/featureflags"
 	"github.com/DataDog/datadog-iac-scanner/pkg/logger"
 	"github.com/DataDog/datadog-iac-scanner/pkg/model"
+	"github.com/DataDog/datadog-iac-scanner/pkg/utils"
 	"github.com/pkg/errors"
 	"github.com/rs/zerolog"
 )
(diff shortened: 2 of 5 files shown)


assets/libraries/ansible.rego
```diff
@@ -50,6 +50,15 @@ validGroup("always") = true
 
 validGroup("rescue") = true
 
+# Returns canonical resource type as "provider.short_module" (e.g. aws.ec2_instance).
+# module: FQCN or short name (e.g. community.aws.ec2_instance or ec2_instance).
+# provider: aws, azure, gcp, or general.
+canonical_resource_type(module, provider) = canonical {
+	parts := split(module, ".")
+	short := parts[minus(count(parts), 1)]
+	canonical := sprintf("%s.%s", [provider, short])
```
ChouraquiBen: I think the other resource are written like `<provider>_<type>`, using a `_` a separator, but it's the only thing I would change

whitemerch: Indeed, but the semantic for ansible is the `.` between each other, so I was more trying to be coherent inside Ansible, not with the other platforms

valeryjuli: I agree with Jeff, I think we should align with the separators of other platforms for consistency

valeryjuli (commented):
LGTM just left same comment as @ChouraquiBen

ChouraquiBen (approved):
LGTM

valeryjuli (approved):

PR #37, opened 2026-03-12
Ansible resourceName: align with module identifiers and manifest

whitemerch:
## Motivation

Ansible findings should report the **resource identifier** (e.g. bucket name, `db_instance_identifier`) instead of only the task name, so users can locate the exact resource in their playbooks. This PR ensures every Ansible rule that sets `resourceName` uses the correct module identifier from official Ansible docs, and keeps the manifest and verification doc in sync.

## Changes

- Two rules were reporting wrong searchKeys, making the findings being dropped because no resource lines reported
- Corrected identifier fields and variable names where they didn’t match Ansible module parameters (e.g. `aws_s3` → `bucket`, `opensearch` → `domain_name`, `iam_policy` → `policy_name`/`iam_name`, `iam_managed_policy` variable renames, account-scoped IAM password policy → `task.name` only).`cos_node_image_not_used` → `gcp_container_node_pool`; added config/hosts rules with literal identifiers).

## Author Checklist

## QA Instruction

- Run a scan on the Ansible samples of the rules and generate SARIF. In the output, confirm that reported `resourceName` values match the real identifiers from the example tasks (e.g. bucket name, `db_instance_identifier`, task name where appropriate).
- Review the diff to ensure only intended Rego identifier fields, manifest rows, and verification doc updates are changed.

## Blast Radius

Ansible query results only: `resourceName` in findings may change to the real resource id (e.g. bucket name) where it was previously wrong or task-only. No API or scanner behavior change beyond that.

## Additional Notes

### DIFF
--- a/assets/queries/ansible/aws/public_lambda_via_api_gateway/query.rego
+++ b/assets/queries/ansible/aws/public_lambda_via_api_gateway/query.rego
@@ -5,17 +5,17 @@ import data.generic.ansible as ansLib
 CxPolicy[result] {
 	task := ansLib.tasks[id][t]
 	modules := {"community.aws.lambda_policy", "lambda_policy"}
-	lambda := task[modules[m]]
-	ansLib.checkState(lambda)
+	lambdaPolicy := task[modules[m]]
+	ansLib.checkState(lambdaPolicy)
 
-	lambdaAction(lambda.action)
-	principalAllowAPIGateway(lambda.principal)
-	re_match("/\\*/\\*$", lambda.source_arn)
+	lambdaAction(lambdaPolicy.action)
+	principalAllowAPIGateway(lambdaPolicy.principal)
+	re_match("/\\*/\\*$", lambdaPolicy.source_arn)
 
 	result := {
 		"documentId": id,
 		"resourceType": modules[m],
-		"resourceName": object.get(lambda, "function_name", task.name),
+		"resourceName": object.get(lambdaPolicy, "function_name", task.name),
 		"searchKey": sprintf("name={{%s}}.{{%s}}.source_arn", [task.name, modules[m]]),
 		"issueType": "IncorrectValue",
 		"keyExpectedValue": "lambda_policy.source_arn should not equal to '/*/*'",
--- a/assets/queries/ansible/aws/public_lambda_via_api_gateway/query.rego
+++ b/assets/queries/ansible/aws/public_lambda_via_api_gateway/query.rego
@@ -15,7 +15,7 @@ CxPolicy[result] {
 	result := {
 		"documentId": id,
 		"resourceType": modules[m],
-		"resourceName": task.name,
+		"resourceName": object.get(lambda, "function_name", task.name),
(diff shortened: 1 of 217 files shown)


assets/queries/ansible/aws/public_lambda_via_api_gateway/query.rego
```diff
@@ -15,7 +15,7 @@ CxPolicy[result] {
 	result := {
 		"documentId": id,
 		"resourceType": modules[m],
-		"resourceName": task.name,
+		"resourceName": object.get(lambda, "function_name", task.name),
```
ChouraquiBen: lambda and lambdaPolicy refer to two different cloud resources, right? Here is a nit if so, to change `lambda` for `lambda_policy`

assets/queries/ansible/aws/rds_associated_with_public_subnet/query.rego
```diff
@@ -34,7 +34,7 @@ CxPolicy[result] {
 	result := {
 		"documentId": id,
 		"resourceType": rds[r],
-		"resourceName": task.name,
+		"resourceName": object.get(instance, "db_instance_identifier", task.name),
```
ChouraquiBen: Same kind of nit here, to change the name for `rds_instance`

assets/queries/ansible/gcp/bigquery_dataset_is_public/query.rego
```diff
@@ -14,7 +14,7 @@ CxPolicy[result] {
 	result := {
 		"documentId": id,
 		"resourceType": modules[m],
-		"resourceName": task.name,
+		"resourceName": object.get(bigquery_dataset, "name", object.get(object.get(bigquery_dataset, "dataset_reference", {}), "dataset_id", task.name)),
```
ChouraquiBen, line 17: Quite the fallback

assets/queries/ansible/gcp/cloud_sql_instance_with_cross_db_ownership_chaining_on/query.rego
```diff
@@ -14,7 +14,7 @@ CxPolicy[result] {
 	result := {
 		"documentId": id,
 		"resourceType": modules[m],
-		"resourceName": task.name,
+		"resourceName": object.get(instance, "name", task.name),
```
ChouraquiBen, line 17: Same nit here but at this point I am wondering if this is necessary: each rule is to be read in its context, not as I do here reading mostly the `resourceName` line.

assets/queries/ansible/general/risky_file_permissions/query.rego
ChouraquiBen, line 30: This may be other things than copy or template, right? Will dest below be enough in this case?

ChouraquiBen (dismissed):
I tried my best not to get lost but despite a few comments, it looks good to me

ChouraquiBen (approved):
LGTM, thanks!

PR #40, opened 2026-03-12
Make E2E tests less fragile so they won't break when we add a platform or provider

jacobotb:
## Motivation

The E2E tests verify the exact output of several help commands. Whenever we add a platform, the help output changes and the E2E tests start failing.

This code changes how those tests so they'll only search for a particular text fragment, not for an exact match.

## Changes

Added an `ExpectedOutputFunc` that takes a function, and switched `ExpectedOut` to that. Removed the files with the expected output.

## Author Checklist

## QA Instruction

## Blast Radius

## Additional Notes


e2e/testcases/general.go
```diff
@@ -20,8 +20,9 @@ type ResultsValidation struct {
 }
 
 type args struct {
-	Args                    []cmdArgs // args to pass to the scanner binary
-	ExpectedOut             []string  // path to file with expected output
+	Args                    []cmdArgs    // args to pass to the scanner binary
+	ExpectedOut             []string     // path to file with expected output
```
ChouraquiBen, line 24: Does this become unused? Should we remove it?

e2e/testcases/e2e-cli-001_help_text.go
```diff
@@ -9,7 +11,11 @@ func init() { //nolint
 			Args: []cmdArgs{
 				[]string{"--help"},
 			},
-			ExpectedOut: []string{"E2E_CLI_001"},
+			ExpectedOutputFunc: []Validation{
```
ChouraquiBen, line 14: Question: in this case, it is better to have a function because it gives freedom to add other tests using other conditions than the contains used in the following files?

ChouraquiBen (approved):
LGTM

e2e/testcases/general.go
jacobotb: Yes, it's unused now. We may remove it, but probably in a separate PR.

e2e/testcases/e2e-cli-001_help_text.go
jacobotb: The problem is that it was comparing the output to the full content of the E2E_CLI_001 file (or, rather, to what it rendered to as a template), and the output changes whenever we make changes in flags, which happens even if you add platforms. So the function lets you do a more targeted test (and you can even use regexps for example.)

PR #1, opened 2026-02-18
Put the head_ref in an envvar to avoid shell injection

jacobotb:
## Motivation

Original code just inserted the content of the head_ref in a command-line, which made it possible to insert shell commands. By putting it in an environment variable first, we ensure it goes within the quotes.

## Author Checklist

- [X] I have reviewed my own PR.
- [X] I have added or updated relevant unit tests where necessary. If no tests are added, I've explained why.
- [X] All new and existing tests pass.
- [ ] I have tested my changes on staging (if applicable).
- [ ] I have updated any relevant documentation (if applicable).

## QA Instruction

Tell the reviewer how they can make sure the PR is indeed working as intended. Tell them what they should be looking for.

## Blast Radius

What services will this change impact. Is it only DataDog IaC Scanner, internal services, the documentation or anything else?

## Additional Notes

If you need to share anything else along with your PR, please do it here.

I submit this contribution under the Apache-2.0 license.

PR #42, opened 2026-03-16
[CICD Coverage #1] Create specific CICD parsing

ChouraquiBen:
## Motivation

There are only few rules to scan github-actions, which are not enough to cover the flaws hackers may use to attack a codebase. This series of PRs aims to extend the CICD scan coverage.

This first PR changes the go code to add the required features for the rules to be added:
- Adding comment to the documents sent to be scanned by the rego rules
- Adding bash expressions parsing
- Adding specific github-actions parsing
- Enabling dependabot configuration files to be scanned on the cicd platform.

To do so, other changes had to be performed:

- Enabled CGO to have tree-sitter parsing that requires it
- As a side effect of enabling CGO, removed vendor usage that does not support CGO

**Not supported**

- Powershell
- Heredocs
- Functions / loops

## Author Checklist

## Blast Radius

iac-scanning - CICD findings

## Additional Notes


.github/workflows/go-e2e.yaml
```diff
@@ -38,15 +38,15 @@ jobs:
             ${{ runner.OS }}-build-${{ env.cache-name }}
             ${{ runner.OS }}-build-
             ${{ runner.OS }}-
-      - name: Get Modules
-        run: go mod vendor
       - name: Build datadog-iac-scanner
         run: |
+          CGO_ENABLED=1
```
ChouraquiBen, line 43: Tree-sitter uses c packages, thus the need of CGO enabled

pkg/analyzer/analyzer.go
```diff
@@ -70,6 +70,9 @@ var (
 	cicdOnRegex                                     = regexp.MustCompile(`\s*on:\s*`)
 	cicdJobsRegex                                   = regexp.MustCompile(`\s*jobs:\s*`)
 	cicdStepsRegex                                  = regexp.MustCompile(`\s*steps:\s*`)
+	dependabotVersionRegex                          = regexp.MustCompile(`\s*version:\s*`)
```
ChouraquiBen, line 74: Check dependabot config files usual patterns to be able to notice one

pkg/analyzer/analyzer.go
```diff
@@ -532,9 +544,13 @@ func (a *analyzerInfo) checkContent(ctx context.Context, results, unwanted chan<
 }
 
 func checkReturnType(ctx context.Context, path, returnType, ext string, content []byte) string {
+	fmt.Printf("returnType: %s, ext: %s\n", returnType, ext)
 	if returnType != "" {
-		if returnType == "cdkTf" {
+		switch returnType {
+		case "cdkTf":
 			return terraform
+		case dependabot:
```
ChouraquiBen, line 552: Return dependabot config files as part of the cicd platform. This is a choice made not to confuse customers as dependabot can be considered a part of cicd although, it is not a github action

.github/workflows/go-ci.yml
```diff
@@ -68,9 +68,6 @@ jobs:
             ${{ runner.OS }}-build-${{ env.cache-name }}
             ${{ runner.OS }}-build-
             ${{ runner.OS }}-
-      - name: Get Modules
-        run: |
-          go mod vendor
```
ChouraquiBen, line 73: mod vendor does not support CGO dependencies

pkg/parser/yaml/cicd/expression_parser.go
```diff
@@ -0,0 +1,328 @@
+/*
```
ChouraquiBen, line 1: This parser aims to parse expressions specific to github actions

pkg/parser/yaml/cicd/parser.go
```diff
@@ -0,0 +1,362 @@
+/*
```
ChouraquiBen, line 1: Based on the YAML parser, it adds the expressions parsed for the shell parser and the expression parser

pkg/parser/yaml/cicd/shell_parser.go
```diff
@@ -0,0 +1,348 @@
+/*
```
ChouraquiBen, line 1: Parses shell expressions, only bash, zsh, sh, no powershell for now

pkg/parser/yaml/cicd/shell_parser.go
```diff
@@ -0,0 +1,348 @@
...
+	// Normalize shell name
+	normalized := normalizeShell(shell)
+
+	// Select appropriate parser based on shell type
+	var lang *sitter.Language
+	switch normalized {
+	case bash_shell, sh, zsh:
+		lang = p.bashLang
+	case pwsh, powershell:
+		// PowerShell not yet supported - Go bindings not available
```
ChouraquiBen, line 68: I might have found a dependency but I'll look later into adding it

pkg/parser/yaml/parser.go
```diff
@@ -139,7 +139,6 @@ func (p *Parser) SupportedTypes() map[string]bool {
 		"googledeploymentmanager": true,
 		"pulumi":                  true,
 		"serverlessfw":            true,
-		"cicd":                    true,
```
ChouraquiBen, line 142: Transfer the cicd parsing ownership to cicd parser

pkg/parser/yaml/cicd/expression_parser.go
```diff
@@ -0,0 +1,328 @@
...
+	fullMatchLen = 4
+)
+
+// ExpressionParser parses GitHub Actions expressions using tree-sitter
+type ExpressionParser struct {
+	ghactionsLang *sitter.Language
+}
+
+// getExpressionParser returns the expression parser, initializing it if needed
+func (p *Parser) getExpressionParser() *ExpressionParser {
```
ChouraquiBen: Returning expression parser that is stateless, so will not be facing race conditions or shared states. Do you think it may lead to mistakes? Should I just create a new expression parser every time?

pkg/parser/yaml/cicd/parser.go
ChouraquiBen, line 1: Should have copied the parser in a first commit and then create all the changes related to cicd. I will point them out with comments

pkg/parser/yaml/cicd/parser.go
```diff
@@ -0,0 +1,362 @@
...
+)
+
+// Parser defines a parser type
+type Parser struct {
+	shellParser      *ShellScriptParser
+	expressionParser *ExpressionParser
+}
+
+// ParsedCommand represents a parsed shell command
+type ParsedCommand struct {
```
ChouraquiBen, line 21: Create all the structures related to gh-actions and shell parsing

pkg/parser/yaml/cicd/parser.go
```diff
@@ -0,0 +1,362 @@
...
+
+	if len(documents) == 0 {
+		return nil, nil, nil, nil, errors.New("no documents found in yaml file")
+	}
+
+	// Convert keys to string format
+	documents = convertKeysToString(documents)
+
+	// Enhance documents with parsed run blocks
+	p.enhanceWithParsedRuns(ctx, documents)
```
ChouraquiBen, line 110: Add shell parsing

pkg/parser/yaml/cicd/parser.go
```diff
@@ -0,0 +1,362 @@
...
+
+	// Enhance documents with parsed expressions
+	p.enhanceWithParsedExpressions(ctx, documents)
```
ChouraquiBen, line 113: Add gh action Expression parsing

pkg/parser/yaml/cicd/parser.go
```diff
@@ -0,0 +1,362 @@
...
+func (p *Parser) GetCommentToken() string {
+	return "#"
+}
+
+// StringifyContent converts original content into string formatted version
+func (p *Parser) StringifyContent(content []byte) (string, error) {
+	return string(content), nil
+}
+
+// getShellParser returns the shell parser, initializing it if needed
```
ChouraquiBen: Back to parsing from here

pkg/parser/yaml/cicd/expression_parser.go
jacobotb: It depends on the parser.

If it's stateless and the same for every Parser, you could just create one at startup and then use it every time:

```
var ExpressionParser = &ExpressionParser{
	ghactionsLang: sitter.NewLanguage(ghactions.Language()),
}
```

then you wouldn't even need `NewExpressionParser` or `Parser.getExpressionParser`.

But at that point, do you even need an `ExpressionParser` object? You could just have the `ParseExpression` and `ExtractExpressionsFromString` functions. This is not Java, you don't need a class for everything :-)

ChouraquiBen: That's... absolutely right x'). I'll change this

pkg/parser/yaml/parser.go
```diff
@@ -23,32 +23,29 @@ type Parser struct {
 }
```
MikaYuoadas: 💡 **Suggestion**: I think this Parser is no longer used anywhere (as well as its `GetKind()`, `GetCommentToken()` and `StringifyContent()` methods) as we’re now using the yaml Parser from `pkg/parser/yaml/default` inside scan.go.

pkg/parser/yaml/default/parser.go
```diff
@@ -0,0 +1,127 @@
...
+		return
+	}
+	for _, value := range mapResources {
+		mapValue, ok := value.(map[string]interface{})
+		if !ok {
+			continue
+		}
+		processElements(ctx, mapValue, filePath)
+	}
+}
```
MikaYuoadas: 💡 **Suggestion**: Why not reuse `yamlParser.AddExtraInfo()` instead?


```suggestion
	// UnmarshalYAML already adds line tracking, so we can use documents directly
	return resolved, yamlParser.ConvertKeysToString(yamlParser.AddExtraInfo(ctx, documents, filePath)), ignoreLines, resolvedFiles, nil
}

// SupportedExtensions returns extensions supported by this parser, which are yaml and yml extension
func (p *Parser) SupportedExtensions() []string {
	return []string{".yaml", ".yml"}
}

// SupportedTypes returns types supported by this parser, which are ansible, cloudFormation, k8s
func (p *Parser) SupportedTypes() map[string]bool {
	return map[string]bool{
		"ansible":                 true,
		"cloudformation":          true,
		"kubernetes":              true,
		"crossplane":              true,
		"knative":                 true,
		"openapi":                 true,
		"googledeploymentmanager": true,
		"pulumi":                  true,
		"serverlessfw":            true,
	}
}

// GetKind returns YAML constant kind
func (p *Parser) GetKind() model.FileKind {
	return model.KindYAML
}
```

pkg/parser/yaml/cicd/parser.go
```diff
@@ -0,0 +1,271 @@
...
+	return "#"
+}
+
+// StringifyContent converts original content into string formatted version
+func (p *Parser) StringifyContent(content []byte) (string, error) {
+	return string(content), nil
+}
+
+// enhanceWithParsedRuns walks through documents and parses run blocks with tree-sitter
+func (p *Parser) enhanceWithParsedRuns(ctx context.Context, documents []model.Document) {
```
MikaYuoadas, line 147: 💡 **Suggestion**: This only walks `jobs[].steps[].run`, [composite actions](https://docs.github.com/en/actions/tutorials/create-actions/create-a-composite-action) are not supported (`runs.steps[].run`).
I’m not expecting you to grow even more the scope of this PR, but could you maybe create a ticket so we can address that if we find time in the future?

MikaYuoadas (dismissed):
Thanks for addressing all my previous comments 🙏 
It looks good with the composition, we’re not duplicating as much code anymore 👍 

Just left a couple comments for remaining cleanup I found, but otherwise I approved the PR.

MikaYuoadas (approved):
LGTM 👍

PR #43, opened 2026-03-16
Ansible: canonical resource type/name and regos fixes

whitemerch:
> [!TIP]

## Motivation

Unify how Ansible queries report **resourceType** and **resourceName** in findings, and avoid duplicated module/variant logic across all of Ansible queries. The engine expects consistent CxPolicy output; queries were using ad-hoc module names and manual name resolution. Some queries are using deprecated modules, some are using outdated ones, and some are mixing both, which is giving completely wrong regos

## Changes

- Added a map in [Add map canonical module => variants + identifier accessor](https://github.com/DataDog/datadog-iac-scanner/pull/43/commits/fdfe7a46826da36f9b0099fccdab9c015e6a6a61), that for each canonical resource type, that is represented by the last segment of the most up to date module name, lists its variants, the ones that are up to date, and the ones in transition. It also lists for each the accessor that gets us the resource name. I have spent quite some time checking one module by module, so I am confident on the map.

- Edit the Ansible queries to use this map for both the resource name and types, which made me realize some regos were completely wrong, so refining some of them, which these are commented to help the reviewer with it.  Now all of the regos report the correct resource type and name, and more importantly catch security issues on the most up to date modules in Ansible.

- Edit the tests after these changes

## Author Checklist

## QA Instruction

- Run Ansible query tests: `go test ./test/ -run 'TestQueries/ansible' -v`. All should pass.
- Run full test suite to ensure no regressions. Optionally run the scanner on sample Ansible playbooks and confirm findings show sensible `resourceType` and `resourceName`.

## Blast Radius

Ansible Regos.

## Additional Notes


assets/libraries/ansible.rego
whitemerch, line 1: I am almost 100% sure of this mapping

assets/queries/ansible/aws/ec2_instance_has_public_ip/query.rego
```diff
@@ -3,74 +3,46 @@ package Cx
 import data.generic.ansible as ansLib
 import data.generic.common as common_lib
 
+# ec2_launch_template
 CxPolicy[result] {
 	task := ansLib.tasks[id][t]
-	modules := {"amazon.aws.ec2", "ec2"}
```
whitemerch, line 8: `amazon.aws.ec2` does not even exist, so that first block was targeting a non-existent FQCN. It was essentially dead code

assets/queries/ansible/aws/ecs_service_without_running_tasks/query.rego
```diff
@@ -3,47 +3,49 @@ package Cx
...
-		"keyExpectedValue": sprintf("%s.deployment_configuration should have at least 1 task running", [modules[m]]),
-		"keyActualValue": sprintf("%&s.deployment_configuration must have at least 1 task running", [modules[m]]),
+		"keyExpectedValue": sprintf("%s.deployment_configuration should have at least 1 task running", [variant]),
+		"keyActualValue": sprintf("%s.deployment_configuration must have at least 1 task running", [variant]),
 	}
 }
 
 checkContent(deploymentConfiguration) {
-    common_lib.valid_key(deploymentConfiguration, "maximum_percent")
+	common_lib.valid_key(deploymentConfiguration, "maximum_percent")
```
whitemerch, line 47: Inconsistent space

assets/queries/ansible/aws/elb_using_insecure_protocols/query.rego
whitemerch, line 1: Each block now reports the correct canonical resource type for the specific module it matched, rather than mixing them in one set. No logic was added or removed, it's the same 3 checks, just duplicated per canonical so we can report the right type and name

assets/queries/ansible/aws/elb_using_weak_ciphers/query.rego
whitemerch, line 1: Same

assets/queries/ansible/aws/instance_uses_metadata_service_IMDSv1/query.rego
whitemerch, line 1: Same

assets/queries/ansible/aws/stack_without_template/query.rego
whitemerch, line 1: So the old code was actually wrong for `cloudformation_stack_set`: it was looking for stack_name in a module that uses name. Now each canonical gets the right name_key from the library. That's a bugfix on top of the refactor, where each block handles each module, same logic as the comments above

assets/queries/ansible/azure/aks_network_policy_misconfigured/query.rego
```diff
@@ -44,26 +45,27 @@ CxPolicy[result] {
...
+		"searchKey": sprintf("name={{%s}}.{{%s}}", [task.name, variant]),
 		"issueType": "MissingAttribute",
 		"keyExpectedValue": "Azure AKS cluster network policy should be defined",
 		"keyActualValue": "Azure AKS cluster network policy is undefined",
 	}
 }
 
 isValidNetworkPolicy(policy) {
-	policy = "calico"
+	policy == "calico"
```
whitemerch, line 66: bahaha

assets/queries/ansible/azure/azure_instance_using_basic_authentication/query.rego
whitemerch, line 1: The old rule accessed input.document directly instead of using ansLib.tasks, so it missed tasks inside blocks and only matched the short module name. Refactored to use the standard library pattern, which adds FQCN support (azure.azcollection.azure_rm_virtualmachine), block/rescue/always task handling, and checkState to skip absent resources. Detection logic is unchanged

assets/queries/ansible/general/risky_file_permissions/query.rego
whitemerch, line 1: I think that's the rule that changed the most:

- Rule 1 (preserve check):
    - **Before**: Matched any module key on the task (action := task[m]), then excluded copy/template by name. This was overly broad, it'd match non-module keys too.
    - **After**: Iterates over a set of specific canonicals (archive, assemble, file, get_url, lineinfile, replace) using get_variants. Cleaner and only matches real modules

- Rule 2 (missing mode on file-creating modules):
    - **Before**: Hardcoded array of 12 variant strings.
    - **After**: Set of 6 canonicals (archive, assemble, copy, file, get_url, template) with get_variants. Same check, but each canonical gets proper resourceType and resourceName (via name_key, these modules use dest or path).

- Rule 3 (create=true but no mode):
    - Before: One block with a map of 8 variant strings → default booleans. Does the job but mixed all modules in one body
    - **After**: 4 separate blocks, one per canonical (blockinfile, htpasswd, ini_file, lineinfile), each with its own create_default_* constant. More verbose but each reports the correct canonical resource type and uses the right name_key from the library (path for all four)

Detection logic is identical across all three rules. The expansion is because each canonical needs its own block to report the right resourceType

whitemerch (commented):
Notes for reviewer

assets/libraries/ansible.rego
MikaYuoadas: I am almost 100% sure I’m confident to say may I’m certain this mapping looks correct 👍 
My only nit is that it’s mixing tab and spaces for indentation 😨

assets/queries/ansible/general/risky_file_permissions/query.rego
```diff
@@ -3,89 +3,144 @@ package Cx
 import data.generic.ansible as ansLib
 import data.generic.common as common_lib
 
+# Rule 1: mode == "preserve" not allowed for modules other than copy/template
+# One body per canonical that can have mode but is not copy/template
+preserve_check_canonicals := {"archive", "assemble", "file", "get_url", "lineinfile", "replace"}
 
 CxPolicy[result] {
+	some canonical
```
MikaYuoadas: nit: this line isn’t necessary as the following one is already acting as an iterator that binds values to canonical. (same for the other "some canonicals" below

assets/queries/ansible/aws/iam_access_key_is_exposed/query.rego
```diff
@@ -2,22 +2,24 @@ package Cx
...
+		"resourceType": canonical,
+		"resourceName": ansLib.get_resource_name(resource, canonical, task),
+		"searchKey": sprintf("name={{%s}}.{{%s}}.access_key_state", [task.name, variant]),
 		"issueType": "IncorrectValue",
-		"keyExpectedValue": "iam.name should be 'root' for an active access key",
-		"keyActualValue": sprintf("iam.name is '%s' for an active access key", [iamuser.name]),
+		"keyExpectedValue": "iam_access_key should not be active for root user",
+		"keyActualValue": sprintf("iam_access_key is active for user '%s' (root)", [resource.user_name]),
 	}
 }
```
MikaYuoadas, line 28: I don’t understand the change in logic, it seems to now just be a worse (doesn’t check for `active != absent` field) copy of `root_account_has_active_access_keys` checking if a root user has active `iam_access_key` where the original intent was to check if a non-root user had an active `iam_access_key`.
Either change it to match the old intent or remove it to not overlap `root_account_has_active_access_keys`.

assets/queries/ansible/aws/ec2_instance_has_public_ip/query.rego
MikaYuoadas: Is it though? This module [used to exists until ansible v6](https://docs.ansible.com/projects/ansible/6/collections/amazon/aws/ec2_module.html).

If we remove this we drop support for customer ansible playbook that used older ansible version.
More generally this pose a question of what do we support ? Only the latest version of providers and/or modules? If not how far back do we look? X years? X versions?
I think that’s an interesting product question to discuss in our backroom.

Same applies to other rules (I’ve seen some dropping support for ec2_instance_has_public_ip, ec2_not_ebs_optimized, ec2_instance_using_default_vpc, ec2_instance_using_default_security_group, instance_with_no_vpc...)

assets/queries/ansible/aws/ebs_volume_encryption_disabled/query.rego
```diff
@@ -25,24 +26,19 @@ CxPolicy[result] {
 
 CxPolicy[result] {
 	task := ansLib.tasks[id][t]
-	ec2_vol := task[modules[m]]
-	checkState(ec2_vol)
+	variant := ansLib.get_variants(canonical)[_]
+	ec2_vol := task[variant]
+	ansLib.checkState(ec2_vol)
```
MikaYuoadas, line 32: Now it only check for `state != "absent"` while it used to also check for `state != "list"` is that intended or a regression?

assets/queries/ansible/aws/db_security_group_open_to_large_scope/query.rego
```diff
@@ -1,11 +1,14 @@
 package Cx
 
 import data.generic.ansible as ansLib
+import data.generic.common as common_lib
```
MikaYuoadas: nit: it is imported but not used

MikaYuoadas (changes requested):
Globally LGTM, great work.

I got a couple minor nitpick.
One rule seems wrong and another need a discussion on what support for historical version of providers/modules... we want to provide (more of a product question really).

assets/queries/ansible/aws/iam_access_key_is_exposed/query.rego
whitemerch: Yes my bad

assets/queries/ansible/aws/ec2_instance_has_public_ip/query.rego
whitemerch: You know I searched for it on the browser [here](https://www.google.com/search?q=amazon.aws.ec2+ansible&sca_esv=f818ef0c6d37c692&rlz=1C5GCCM_en&sxsrf=ANbL-n7eVwB00dz84vhf0KmTUboaHC0k9A%3A1773936444847&ei=PB-8admpM7qakdUPmeLjyQE&biw=1512&bih=738&ved=0ahUKEwjZt5idrKyTAxU6TaQEHRnxOBkQ4dUDCBE&uact=5&oq=amazon.aws.ec2+ansible&gs_lp=Egxnd3Mtd2l6LXNlcnAiFmFtYXpvbi5hd3MuZWMyIGFuc2libGUyBhAAGBYYHjIGEAAYFhgeMgYQABgWGB4yBhAAGBYYHjIGEAAYFhgeMgYQABgWGB4yBhAAGBYYHjIGEAAYFhgeMggQABgWGAoYHjIIEAAYFhgKGB5I8whQjAFY9gdwAXgAkAEAmAGeAaABwgaqAQMzLjW4AQPIAQD4AQGYAgmgAtcGwgIJEAAYsAMYBxgewgILEAAYsAMYBxgKGB7CAggQABiABBiwA8ICBBAAGB6YAwCIBgGQBgqSBwM0LjWgB4BDsgcDMy41uAfUBsIHAzAuOcgHD4AIAA&sclient=gws-wiz-serp), and didn't find it, I assumed it didn't exist

whitemerch: IMO, we can try to keep up the direct previous version of a module, which in general is still supported at this moment but just act as a redirect until some time. We can't say let's just report new modules as this would break customer that are still transitioning, but at the same time, checking each module by itself, to check if one is deprecated or not is a bit of an effort. I guess it doesn't hurt to still have a couple of modules for the modules that exist in two forms, until a certain time where we can assume we can clean them up. 

As for the ec2, as I said above, I completely missed this, but looking at it, it has been removed 4 years ago, so I'll remove its support, and fully rely on the new corresponding instances

MikaYuoadas: Well it existed quite a long time ago. I agree with your point, we can’t support everything that ever existed.
At a product level what’s the garanties we gives our customers in term of support, if any?
You don’t really need to change this PRs but I wanted to take the opportunity to raise that point.
I think we should discuss it next week with Valery and Michelle to have a clear stance on this so we don’t guesstimate when it’s ok to drop support in the future. :)

whitemerch: Indeed, you are right. I guess there is a clear difference between the modules that will be deprecated, but still serve as a redirect to the new modules, which in this case IMO we should support, and the ones that are completely dropped, which we shouldn't support

MikaYuoadas (approved):

PR #45, opened 2026-03-17
[CICD Coverage #2] HIGH Severity rules

ChouraquiBen:
> [!TIP]
> To review the rules, use the positive and negative cases to grasp a general idea of the rule

## Motivation

There are only few rules to scan github-actions, which are not enough to cover the flaws hackers may use to attack a codebase. This series of PRs aims to extend the CICD scan coverage.

This second PR adds HIGH Severity rules to the scan.

## Changes

- Bot conditions
- Cache Poisoning
- Dangerous Triggering
- Dependabot Execution
- Github environment
- Hardcoded Container credentials
- Unpinned images
- Unsecured commands
- Unsound Conditions
- Unsound Contains

## Author Checklist

## QA Instruction

- To review the rules, use the positive and negative cases to understand what should be flagged and what should not be.
- Review the attributes returned in the results of the rules: ensure that the values are all there, that they explain stuff related to the issue.
- Take a look at the logic to make sure their is no flaw.

## Blast Radius

IaC findings for CICD platform

## Additional Notes


assets/queries/cicd/github/bot_conditions/test/negative.yaml
```diff
@@ -0,0 +1,80 @@
+name: Secure Bot Check
```
ChouraquiBen: [This article](https://org.boostsecurity.io/blog/weaponizing-dependabot-pwn-request-at-its-finest) gives some insights about what is vulnerable. Look for _Confusable or Forgeable Contexts_

assets/queries/cicd/github/bot_conditions/query.rego
```diff
@@ -0,0 +1,328 @@
+package Cx
```
ChouraquiBen: This rule is very long and very complex. I have tried my best but could not make it simpler

ChouraquiBen: The goal is to check if users are checking both ids and names when they are checking one of them

assets/queries/cicd/github/bot_conditions/query.rego
```diff
@@ -0,0 +1,328 @@
...
+}
+
+# Parse and check for bot conditions in an if expression
+# Returns: {is_vulnerable: bool, context: string}
+check_bot_condition_in_expression(expr) = has_bot_check {
+	# Remove ${{ }} wrapper if present
+	cleaned := trim_space(trim_prefix(trim_suffix(trim_space(expr), "}}"), "${{"))
+
+	# Skip if entire expression is negated with ! operator
+	not startswith(cleaned, "!")
```
ChouraquiBen: supposes that users won't write checks such as `!(id != bot_id)`. Not ideal, but works as first draft of the rule. Added a tack to tackle this

assets/queries/cicd/github/bot_conditions/query.rego
```diff
@@ -0,0 +1,328 @@
...
+
+	# Check for OR with bot condition (dominating)
+	contains(cleaned, "||")
```
ChouraquiBen: First tackle the or: if there is an or, assumes that the whole expression might be vulnerable

assets/queries/cicd/github/bot_conditions/query.rego
```diff
@@ -0,0 +1,328 @@
...
+	has_bot_check := check_or_expression(cleaned)
+} else = has_bot_check {
+	# Remove ${{ }} wrapper if present
+	cleaned := trim_space(trim_prefix(trim_suffix(trim_space(expr), "}}"), "${{"))
+
+	# Skip if entire expression is negated with ! operator
+	not startswith(cleaned, "!")
+
+	# Check for AND with bot condition (non-dominating)
+	contains(cleaned, "&&")
```
ChouraquiBen: else it may have an and, so specific checks will be performed to see if the expression contains the necessary secure parts to make it non-vulnerable or else if it contains the parts that make it vulnerable.

assets/queries/cicd/github/bot_conditions/query.rego
```diff
@@ -0,0 +1,328 @@
...
+	# Remove ${{ }} wrapper if present
+	cleaned := trim_space(trim_prefix(trim_suffix(trim_space(expr), "}}"), "${{"))
+
+	# Skip if entire expression is negated with ! operator
+	not startswith(cleaned, "!")
+
+	# Simple equality check without logical operators (dominating)
+	not contains(cleaned, "||")
+	not contains(cleaned, "&&")
+	has_bot_check := check_equality_expression(cleaned)
```
ChouraquiBen: If it does not contain any of those, it is only checking on equals. Not taking into account reverse negative and checking on for example `id != dependabot_id` is not considered vulnerable

assets/queries/cicd/github/bot_conditions/query.rego
```diff
@@ -0,0 +1,328 @@
...
+} else = {"is_vulnerable": false, "context": ""}
+
+# Check OR expression parts for bot conditions
+check_or_expression(expr) = result {
```
ChouraquiBen: To check or, you check every equal and see if at least one of them is vulnerable

assets/queries/cicd/github/bot_conditions/query.rego
ChouraquiBen: This is not a good rule to implement within rego: it would be better to recursively check the AST tree to check the conditions from the most deeply nested to the surface. Getting the AST tree from the Parser can be done but rego does not support recursion

assets/queries/cicd/github/bot_conditions/query.rego
```diff
@@ -0,0 +1,328 @@
...
+	parts := split(expr, "||")
+	vulnerables = [parts[i] | bot_check := check_equality_expression(trim_space(parts[i])); bot_check.is_vulnerable]
+	count(vulnerables) > 0
+	string = concat(" || ", vulnerables)
+	result := {"is_vulnerable": true, "context": string}
+} else = {"is_vulnerable": false, "context": ""}
+
+# Check AND expression parts for bot conditions
+check_and_expression(expr) = result {
```
ChouraquiBen: To check and, you check if the secure pattern exists and then if not, if at least one part is vulnerable

assets/queries/cicd/github/bot_conditions/query.rego
```diff
@@ -0,0 +1,328 @@
...
+		bot_check := check_equality_expression(trim_space(parts[i]));
+		bot_check.is_vulnerable
+	]
+	count(vulnerable_parts) > 0
+	string = concat(" && ", vulnerable_parts)
+	result := {"is_vulnerable": true, "context": string}
+} else = {"is_vulnerable": false, "context": ""}
+
+# Check if an AND expression is a secure pattern (name + ID verification)
+is_secure_and_pattern(expr) {
```
ChouraquiBen: To check if a pattern is secure, check if at least one id is given and at least one name and check if they match

assets/queries/cicd/github/bot_conditions/query.rego
```diff
@@ -0,0 +1,328 @@
...
+	lower_name := lower(name_ctx)
+	lower_id := lower(id_ctx)
+
+	# github.event.pull_request.sender.login with github.event.pull_request.sender.id
+	lower_name == "github.event.pull_request.sender.login"
+	lower_id == "github.event.pull_request.sender.id"
+}
+
+# Check a single equality expression for bot conditions
+check_equality_expression(expr) = result {
```
ChouraquiBen: To check equality expression you check that both parts of the equal match bot name or bot id

assets/queries/cicd/github/bot_conditions/query.rego
```diff
@@ -0,0 +1,328 @@
...
+# Handles: github.actor, github['actor'], GITHUB.actor, GitHub['ACTOR']
+extract_context_path(raw) = result {
+	# Remove any brackets and quotes to normalize
+	cleaned := replace(replace(replace(raw, "[", "."), "]", ""), "'", "")
+	cleaned2 := replace(cleaned, "\"", "")
+	result := lower(cleaned2)
+}
+
+# Job-level bot condition check
+CxPolicy[result] {
```
ChouraquiBen: Check at job or step level

assets/queries/cicd/github/cache_poisoning/query.rego
```diff
@@ -0,0 +1,195 @@
+package Cx
```
ChouraquiBen, line 1: The rule works with a list of caches known as vulnerable and actions that would publish artifactes

assets/queries/cicd/github/cache_poisoning/query.rego
```diff
@@ -0,0 +1,195 @@
...
+	"Azure/functions-action",
+	"Azure/sql-action",
+	"cloudflare/wrangler-action",
+	"google-github-actions/deploy-appengine",
+	"google-github-actions/deploy-cloudrun",
+	"google-github-actions/deploy-cloud-functions",
+}
+
+# Check if the workflow trigger is used for publishing artifacts
+is_publishing_trigger(trigger) {
```
ChouraquiBen: Two ways of publishing artifacts: either the publishers up there or publishing triggers

assets/queries/cicd/github/cache_poisoning/query.rego
```diff
@@ -0,0 +1,195 @@
...
+	startswith(step.uses, action)
+}
+
+# Check if action uses match (handles versions)
+action_matches(uses, pattern) {
+	startswith(uses, pattern)
+}
+
+# Check if a step uses a cache-aware action and if caching is enabled
+uses_cache_with_config(step) = result {
```
ChouraquiBen, line 95: A cache action might be used but configured so that the cache is disabled

assets/queries/cicd/github/dangerous_triggers/query.rego
```diff
@@ -0,0 +1,127 @@
+package Cx
+
+import data.generic.common as common_lib
+
+# Check for pull_request_target trigger
+CxPolicy[result] {
+	doc := input.document[i]
+
+	# Check if pull_request_target is in the triggers
+	doc.on == "pull_request_target"
```
ChouraquiBen: Either the trigger is a string

assets/queries/cicd/github/dangerous_triggers/query.rego
```diff
@@ -0,0 +1,127 @@
...
+	}
+}
+
+# Check for pull_request_target in array format
+CxPolicy[result] {
+	doc := input.document[i]
+
+	# Check if on is an array and contains pull_request_target
+	is_array(doc.on)
+	doc.on[_] == "pull_request_target"
```
ChouraquiBen: Or a item in an array

assets/queries/cicd/github/dangerous_triggers/query.rego
```diff
@@ -0,0 +1,127 @@
...
+	}
+}
+
+# Check for pull_request_target in object format
+CxPolicy[result] {
+	doc := input.document[i]
+
+	# Check if on is an object and contains pull_request_target
+	is_object(doc.on[key])
+	key == "pull_request_target"
```
ChouraquiBen: Or a key

assets/queries/cicd/github/github_env/query.rego
```diff
@@ -0,0 +1,118 @@
...
+		"issueType": "IncorrectValue",
+		"keyExpectedValue": "Step should not write to GITHUB_ENV or GITHUB_PATH with potentially unsafe data",
+		"keyActualValue": "Step writes to GITHUB_ENV or GITHUB_PATH which may allow code execution",
+		"searchLine": common_lib.build_search_line(["jobs", j, "steps", s, "run"], []),
+	}
+}
+
+
+# Check if workflow has dangerous triggers
+has_dangerous_trigger(doc) {
```
ChouraquiBen: Check for dangerous triggers

assets/queries/cicd/github/github_env/query.rego
```diff
@@ -0,0 +1,118 @@
...
+	trigger == ["pull_request_target", "workflow_run"][_]
+}
+
+has_dangerous_trigger(doc) {
+	is_object(doc.on[trigger])
+	trigger == ["pull_request_target", "workflow_run"][_]
+}
+
+# Detect writes to GITHUB_ENV or GITHUB_PATH using tree-sitter parsed data (bash/sh/zsh)
+writes_to_env_file(step, run_block) {
```
ChouraquiBen, line 34: Check if the command writes to the env and if that comes from potential user input

assets/queries/cicd/github/unpinned_images/query.rego
```diff
@@ -0,0 +1,137 @@
+package Cx
```
ChouraquiBen, line 1: Checks for each kind of service:
- if it has a pinned sha
- or is an expression in. which case
  - checks the matrix which allows users to loop in a way
  - if the matrix holds a string that contains a pinned image
  - if the matrix holds a array that contain positive images

assets/queries/cicd/github/unsound_conditions/test/positive.yaml
```diff
@@ -0,0 +1,41 @@
+name: Unsound Conditions
+on: push
+
+jobs:
+  # Case 1: Job-level condition with literal block scalar
+  job_with_unsound_condition:
+    runs-on: ubuntu-latest
+    if: |
```
ChouraquiBen, line 8: The `|` makes the whole expression a multiline string that won't be turned into a boolean

assets/queries/cicd/github/unsound_conditions/test/positive.yaml
```diff
@@ -0,0 +1,41 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Unsound literal block
+        if: |
+          ${{ github.event_name == 'pull_request' }}
+        run: echo "This will always run!"
+
+      # Case 3: Folded block scalar
+      - name: Unsound folded block
+        if: >
```
ChouraquiBen, line 24: Same here

assets/queries/cicd/github/unsound_conditions/test/positive.yaml
```diff
@@ -0,0 +1,41 @@
...
+          ${{ false }}
+        run: echo "This will also always run!"
+
+      # Case 4: Multiple spaces (excessive whitespace)
+      - name: Excessive leading spaces
+        if: "  ${{ github.ref == 'refs/heads/main' }}"
```
ChouraquiBen, line 30: Before the expression, there is a whitespace that will prevent the expression to be treated as a boolean

assets/queries/cicd/github/unsound_contains_with_controllable_input/test/positive.yaml
```diff
@@ -0,0 +1,49 @@
+name: Unsound Contains Usage
+on: pull_request
+
+jobs:
+  # Case 1: Job-level condition with space-separated branches
+  test_branches:
+    runs-on: ubuntu-latest
+    if: ${{ contains('refs/heads/main refs/heads/develop', github.ref) }}
```
ChouraquiBen, line 8: Allows `refs/heads/main/malicious` to run

assets/queries/cicd/github/unsound_contains_with_controllable_input/query.rego
```diff
@@ -0,0 +1,216 @@
+package Cx
+
+import data.generic.common as common_lib
+
+# User-controllable contexts that make this HIGH severity
+user_controllable_contexts := {
```
ChouraquiBen, line 6: Uses user controllable inputs to determine how big the threat is

assets/queries/cicd/github/unsound_contains/query.rego
```diff
@@ -0,0 +1,216 @@
...
+
+	# Only process successfully parsed expressions
+	parsed_expr.parse_ok == true
+
+	# Find vulnerable contains() calls in the AST
+	vulnerable_call := find_vulnerable_contains(parsed_expr.ast)
+
+	# Determine severity based on context
+	is_user_controllable := check_user_controllable_in_ast(vulnerable_call.second_arg)
+	# Severity is not actually controllable through IaC Scanning
```
ChouraquiBen: I let a comment here: The SDLC team said that some rule severity depended on the context. Following Zizmor on the rule, here is an example of how the severity could be determined

assets/queries/cicd/github/unsound_contains_with_controllable_input/query.rego
```diff
@@ -0,0 +1,216 @@
...
+		"keyActualValue": sprintf("Step '%s' contains() condition can be bypassed (%s severity)", [step_name, severity]),
+		"searchLine": common_lib.build_search_line(["jobs", j, "steps", s, "if"], []),
+		"resourceType": "github_action",
+		"resourceName": step_name
+	}
+}
+
+# Find vulnerable contains() calls in the AST
+# Returns an object with first_arg and second_arg if vulnerable pattern found
+find_vulnerable_contains(ast) = result {
```
ChouraquiBen, line 91: to check if the contains is vulnerable, the nodes are checked, even the nested ones in case of `foo(contains(${vulnerable_string}))`

If the first argument of `contains` contains a whitespace or a pipe, the contains is unsound

assets/queries/cicd/github/unsound_contains/query.rego
MikaYuoadas: 💡 **Suggestion:** Could this actually be two rules with different severity?
Like what is done for example with `ansible/aws/iam_access_key_is_exposed` and `ansible/aws/root_account_has_active_access_keys` which are identical except one test for root with `HIGH` severity and the other for non-root with `MEDIUM` severity.

assets/queries/cicd/github/unsound_contains_with_controllable_input/test/positive.yaml
MikaYuoadas: ❓ **Question:** I’m confused, isn’t it substring that would be allowed to run (e.g. `refs/heads/ma` or `refs/heads/mai`)?

assets/queries/cicd/github/unsound_contains/test/negative.yaml
```diff
@@ -0,0 +1,56 @@
...
+    if: ${{ endsWith(github.ref, '/main') }}
+    steps:
+      - run: echo "Main branch"
+
+  # Case 7: Contains with non-user-controllable context
+  test_safe_context:
+    runs-on: ubuntu-latest
+    if: ${{ contains('push pull_request', github.event_name) }}
+    steps:
+      - run: echo "This is informational, not high severity"
```
MikaYuoadas: ❓ **Question:** Won’t this case produce a finding (making it a positive instead of a negative)?
IIUC it only check that first arg is a string with a delimiter (space or pipe).
Being user controllable (like here) doesn’t affect matching, only the severity.

assets/queries/cicd/github/unsound_contains/query.rego
```diff
@@ -0,0 +1,216 @@
...
+
+	first_arg := args[0]
+	second_arg := args[1]
+
+	# First argument must be a string literal
+	is_string_literal(first_arg)
+
+	# First argument must contain spaces or delimiters (indicating multiple values)
+	string_value := get_string_value(first_arg)
+	has_delimiter(string_value)
```
MikaYuoadas: 💡 **Suggestion:** Why do we check if the string has a delimiter?
I would say any contains with string literal pattern should be flagged.
For example `if: ${{ contains('main', github.ref_name) }}` would still allow an attacker to run on any substring branch (e.g. `ai`, `in` or `ma` branches).
Looking at [zizmor](https://github.com/zizmorcore/zizmor/blob/main/crates/zizmor/src/audit/unsound_contains.rs) equivalent rule, that’s what they are actually doing.

assets/queries/cicd/github/unsound_contains/query.rego
```diff
@@ -0,0 +1,216 @@
...
+	# Collect great-great-grandchildren (should be deep enough for most expressions)
+	gggrandchildren := {gggchild |
+		child := node.children[_]
+		grandchild := child.children[_]
+		ggchild := grandchild.children[_]
+		gggchild := ggchild.children[_]
+	}
+
+	nodes := children | grandchildren | great_grandchildren | gggrandchildren
+}
```
MikaYuoadas: 💡 **Suggestion:** Couldn’t we simply use [walk](https://www.openpolicyagent.org/docs/policy-reference/builtins/graph#builtin-graph-walk) instead of these?

i.e. in `find_vulnerable_contains`:
```diff
	# Get all nodes from the AST (root and all descendants)
-	all_nodes := get_all_nodes(ast)
-	node := all_nodes[_]
+	walk(ast, [_, node])
```

assets/queries/cicd/github/unsound_conditions/query.rego
```diff
@@ -0,0 +1,90 @@
...
+	# Get parsed expressions for the if condition
+	parsed_exprs := job_or_step._parsed_expressions_if[_]
+	parsed_exprs.parse_ok == true
+
+	# Get the raw if condition
+	condition := job_or_step["if"]
+
+	# Check if condition has newlines (indicates block scalar style)
+	contains(condition, "\n")
+}
```
MikaYuoadas: ❓ **Question:** Won’t this also incorrectly flag multi-line `|-`/`>-`?
Isn’t internal \n in expression like this one allowed?
```yaml
  if: |-
    ${{ github.event_name == 'push'
      && github.ref == 'main' }}
```

assets/queries/cicd/github/unsound_conditions/query.rego
```diff
@@ -0,0 +1,90 @@
...
+
+has_leading_or_trailing_space(str) {
+	# Check for trailing whitespace beyond a single space
+	endswith(str, "  ")
+}
+
+has_leading_or_trailing_space(str) {
+	# Check for tab characters
+	contains(str, "\t")
+}
```
MikaYuoadas: ❓ **Question:** Is non-trailing tab (i.e. inside the condition expression) really an issue?

MikaYuoadas: 💡 **Suggestion:** Couldn’t we more simply check like zizmor is doing for any extra surrounding space around a fenced expression? (still not sure why your implementation was allowing a single trailing/leading space)
```suggestion
check_unsound_condition(job_or_step) {
	# Get parsed expressions for the if condition
	parsed_exprs := job_or_step._parsed_expressions_if[_]
	parsed_exprs.parse_ok == true

	# Get the raw if condition
	condition := job_or_step["if"]

	#
	trimmed := trim_space(condition)
	startswith(trimmed, "${{")
	endswith(trimmed, "}}")
	condition != trimmed
}
```

MikaYuoadas (changes requested):
First comments on:
* `unsound_contains`
* `unsound_conditions`

assets/queries/cicd/github/unpinned_images/query.rego
```diff
@@ -0,0 +1,137 @@
...
+import data.generic.common as common_lib
+
+# Check for unpinned container images in job container
+CxPolicy[result] {
+	doc := input.document[i]
+	job := doc.jobs[j]
+	container := job.container
+
+	# Check if container has an image
+	image := container.image
```
MikaYuoadas, line 38: ⛔ **must fix:** Github actions also [allow omitting image key if you only set an image](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-running-a-job-within-a-container).
e.g. both syntax are possible:
```yaml
jobs:
  job-a:
    runs-on: ubuntu-latest
    container:
      image: node:18
  job-b:
    runs-on: ubuntu-latest
    container: node:18
```
You’ll be missing the second one with this rule.

assets/queries/cicd/github/unpinned_images/query.rego
```diff
@@ -0,0 +1,137 @@
...
+	is_string(matrix_value)
+	contains(matrix_value, "@sha256:")
+} else {
+	job.strategy
+	job.strategy.matrix
+	matrix_value := job.strategy.matrix[_]
+	is_array(matrix_value)
+	matrix_value[_]
+	contains(matrix_value[_], "@sha256:")
+}
```
MikaYuoadas: ⛔ **must fix:** Here we’re looking at any matrix key, we should scope to only the actual matrix referenced by our image expression to avoid false positive match.

assets/queries/cicd/github/unpinned_images/query.rego
```diff
@@ -0,0 +1,137 @@
...
+# Case 1: Expression references matrix but matrix has no pinned images
+check_unpinned_expression(job, container_or_service) {
+	parsed_exprs := container_or_service._parsed_expressions_image[_]
+	parsed_exprs.parse_ok == true
+
+	# Check if expression references matrix
+	contains(parsed_exprs.raw, "matrix.")
+
+	# Look for pinned images in strategy.matrix values
+	not has_pinned_in_matrix(job)
```
MikaYuoadas: ⛔ **must fix:** Here (and below) we’re only checking that at least one element has a pinned value.
We should require all values from the list to be pinned for the rule to not produce any finding.

assets/queries/cicd/github/hardcoded_container_credentials/test/negative.yaml
```diff
@@ -0,0 +1,13 @@
...
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    container:
+      image: private.registry.com/image:latest
+      credentials:
+        username: ${{ secrets.REGISTRY_USER }}
+        password: ${{ secrets.REGISTRY_PASSWORD }}
+    steps:
+      - run: echo "Running in container"
```
MikaYuoadas, line 13: 💡 **Suggestion:** Could you also add a negative test for service credentials?

assets/queries/cicd/github/hardcoded_container_credentials/test/positive.yaml
```diff
@@ -0,0 +1,13 @@
...
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    container:
+      image: private.registry.com/image:latest
+      credentials:
+        username: myuser
+        password: mypassword123
+    steps:
+      - run: echo "Running in container"
```
MikaYuoadas, line 13: 💡 **Suggestion:** Could you also add a positive test for service credentials?

assets/queries/cicd/github/github_env/query.rego
```diff
@@ -0,0 +1,118 @@
...
+	# PowerShell redirection: >> $env:GITHUB_ENV
+	regex.match(`(?i)>>\s*["']?\$\{?env:(GITHUB_ENV|GITHUB_PATH)\}?["']?`, run_block)
+}
+
+writes_to_env_file(step, run_block) {
+	not step._parsed_run.parse_ok
+
+	# CMD/batch patterns: >> %GITHUB_ENV%
+	regex.match(`(?i)>>\s*["']?%(GITHUB_ENV|GITHUB_PATH)%["']?`, run_block)
+}
```
MikaYuoadas, line 91: 💬 **Remark:** Having regex fallback for now is fine, but maybe you could create a ticket so we remember to come back to address this more cleanly later (if we ever find time to prioritize this 😁)

assets/queries/cicd/github/github_env/query.rego
MikaYuoadas: ❓ **Question:** Here we don’t check for the command before the redirection (zizmor specifically only allows `echo`).
Shouldn’t we also restrict the set of allowed command?
I think while `echo "static" >> $GITHUB_ENV` would be safe, some things like `cat static_file >> $GITHUB_ENV` could still be dangerous provided the "static_file" is in some way coming from user input (say a `echo $DANGEROUS_USER_INPT` >> static_file` step just above).

MikaYuoadas (changes requested):
Continuing review:
* `unpinned_images`
* `hardcoded_container_credentials`
* `github_env`

assets/queries/cicd/github/dangerous_triggers/test/positive1.yaml
```diff
@@ -0,0 +1,14 @@
+name: Dangerous Workflow
+on: 
+  pull_request_target:
+    types: [opened, synchronize]
```
MikaYuoadas, line 4: ❓ **Question:**Shouldn’t this case be
```suggestion
    types: workflow_run
```

assets/queries/cicd/github/dangerous_triggers/test/positive.yaml
```diff
@@ -0,0 +1,11 @@
+name: Dangerous Workflow
+on: pull_request_target
```
MikaYuoadas, line 2: 💡 **Suggestion:** We should also add test cases for array and object triggers.

assets/queries/cicd/github/dangerous_triggers/query.rego
```diff
@@ -0,0 +1,127 @@
...
+
+import data.generic.common as common_lib
+
+# Check for pull_request_target trigger
+CxPolicy[result] {
+	doc := input.document[i]
+
+	# Check if pull_request_target is in the triggers
+	doc.on == "pull_request_target"
+	doc_name := object.get(doc, "name", sprintf("document", []))
```
MikaYuoadas: 💡 **Suggestion:** `sprintf` without formatting verbs or values passed is effectively a noop. Consider just using a string litteral here:
```suggestion
	doc_name := object.get(doc, "name", "document")
```

assets/queries/cicd/github/dangerous_triggers/query.rego
```diff
@@ -0,0 +1,127 @@
...
+
+# Check for pull_request_target in array format
+CxPolicy[result] {
+	doc := input.document[i]
+
+	# Check if on is an array and contains pull_request_target
+	is_array(doc.on)
+	doc.on[_] == "pull_request_target"
+
+	doc_name := object.get(doc, "name", sprintf("document", []))
```
MikaYuoadas: 💡 **Suggestion:**
```suggestion
	doc_name := object.get(doc, "name", "document"
```

assets/queries/cicd/github/dangerous_triggers/query.rego
```diff
@@ -0,0 +1,127 @@
...
+
+# Check for pull_request_target in object format
+CxPolicy[result] {
+	doc := input.document[i]
+
+	# Check if on is an object and contains pull_request_target
+	is_object(doc.on[key])
+	key == "pull_request_target"
+
+	doc_name := object.get(doc, "name", sprintf("document", []))
```
MikaYuoadas: 💡 **Suggestion:**
```suggestion
	doc_name := object.get(doc, "name", "document"
```

assets/queries/cicd/github/dangerous_triggers/query.rego
```diff
@@ -0,0 +1,127 @@
...
+}
+
+# Check for workflow_run trigger
+CxPolicy[result] {
+	doc := input.document[i]
+
+	# Check if workflow_run is in the triggers
+	doc.on == "workflow_run"
+
+	doc_name := object.get(doc, "name", sprintf("document", []))
```
MikaYuoadas: 💡 **Suggestion:**
```suggestion
	doc_name := object.get(doc, "name", "document"
```

assets/queries/cicd/github/dangerous_triggers/query.rego
```diff
@@ -0,0 +1,127 @@
...
+
+# Check for workflow_run in array format
+CxPolicy[result] {
+	doc := input.document[i]
+
+	# Check if on is an array and contains workflow_run
+	is_array(doc.on)
+	doc.on[_] == "workflow_run"
+
+	doc_name := object.get(doc, "name", sprintf("document", []))
```
MikaYuoadas: 💡 **Suggestion:**
```suggestion
	doc_name := object.get(doc, "name", "document"
```

assets/queries/cicd/github/dangerous_triggers/query.rego
```diff
@@ -0,0 +1,127 @@
...
+
+# Check for workflow_run in object format
+CxPolicy[result] {
+	doc := input.document[i]
+
+	# Check if on is an object and contains workflow_run
+	is_object(doc.on[name])
+	name == "workflow_run"
+
+	doc_name := object.get(doc, "name", sprintf("document", []))
```
MikaYuoadas: 💡 **Suggestion:**
```suggestion
	doc_name := object.get(doc, "name", "document"
```

assets/queries/cicd/github/dangerous_triggers/test/negative.yaml
```diff
@@ -0,0 +1,11 @@
+name: Safe Workflow
+on:
+  pull_request:
```
MikaYuoadas, line 3: 💡 **Suggestion:** To be thorough we could also add negative test cases for string litteral and object triggers

assets/queries/cicd/github/unsound_contains/test/negative.yaml
ChouraquiBen: It should have, but the rule was broken so it did not. It is fixed and flagged in a separate rule

assets/queries/cicd/github/dangerous_triggers/query.rego
```diff
@@ -0,0 +1,127 @@
...
+		"documentId": doc.id,
+		"searchKey": "on",
+		"issueType": "IncorrectValue",
+		"keyExpectedValue": "Workflow should not use workflow_run trigger",
+		"keyActualValue": "Workflow uses workflow_run trigger which is almost always used insecurely",
+		"searchLine": common_lib.build_search_line(["on"], []),
+		"resourceType": "github_action",
+		"resourceName": doc_name
+	}
+}
```
MikaYuoadas: 💡 **Suggestion:** A lot of code duplication could be avoided by using a single policy with some helper functions for the "generic" (like handling different types).

🤏 **nitpick:** Also the way it’s written will generate two findings if a workflow has both `pull_request_target` and `workflow_run` triggers.

```suggestion
  dangerous_triggers := {"pull_request_target", "workflow_run"}                                                                                      
                                                                                                                                                   
  # Detect a dangerous trigger regardless of YAML format (string, array, object key)                                                                 
  has_trigger(doc, trigger) {                                                                                                                      
        doc.on == trigger                                                                                                                            
  }                                                                                                                                                

  has_trigger(doc, trigger) {
        is_array(doc.on)
        trigger in doc.on                                                                                                                            
  }
                                                                                                                                                     
  has_trigger(doc, trigger) {                                                                                                                      
        doc.on[trigger]
  }

  # Collect all dangerous triggers found in the workflow                                                                                             
  found_triggers(doc) = triggers {
        triggers := {t | t := dangerous_triggers[_]; has_trigger(doc, t)}                                                                            
  }                                                                                                                                                  
  
  CxPolicy[result] {                                                                                                                                 
        doc := input.document[i]                                                                                                                   
        triggers := found_triggers(doc)                                                                                                              
        count(triggers) > 0
                                                                                                                                                     
        doc_name := object.get(doc, "name", "document")                                                                                              
        trigger_list := concat(", ", triggers)
                                                                                                                                                     
        result := {                                                                                                                                
                "documentId": doc.id,
                "searchKey": "on",                                                                                                                   
                "issueType": "IncorrectValue",
                "keyExpectedValue": "Workflow should not use dangerous triggers",                                                                    
                "keyActualValue": sprintf("Workflow uses dangerous trigger(s): %s", [trigger_list]),                                               
                "searchLine": common_lib.build_search_line(["on"], []),                                                                              
                "resourceType": "github_action",
                "resourceName": doc_name,                                                                                                            
        }                                                                                                                                            
  }
```

MikaYuoadas (changes requested):
Continuing the review:
* `dangerous_triggers`
* `dependabot_execution`

assets/queries/cicd/github/cache_poisoning/query.rego
```diff
@@ -0,0 +1,195 @@
+package Cx
+
+import data.generic.common as common_lib
+import future.keywords.in
+
+# Known cache-aware actions that can be exploited in cache poisoning attacks
+# Format: {"action": "uses_pattern", "disable_field": "field_name", "disable_value": value, "type": "opt_in|opt_out"}
```
MikaYuoadas: 🤏 **nitpick:** Format in comment doesn’t exactly match actual data structure below. Was it an outdated comment from a previous implementation? If so, could you update it?

assets/queries/cicd/github/cache_poisoning/query.rego
```diff
@@ -0,0 +1,195 @@
...
+uses_publisher_action(step) {
+	step.uses
+	some action in publisher_actions
+	startswith(step.uses, action)
+}
+
+# Check if action uses match (handles versions)
+action_matches(uses, pattern) {
+	startswith(uses, pattern)
+}
```
MikaYuoadas, line 92: 🤏 **nitpick:** Not sure of the value of this empty wrapper to `startswith()` 🤔

assets/queries/cicd/github/cache_poisoning/test/positive.yaml
```diff
@@ -0,0 +1,22 @@
...
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-node@v4
+        name: vulnerable code
+        with:
+          node-version: '18'
+          cache: 'npm'
```
MikaYuoadas, line 17: 💡 **Suggestion:** I think we could have a lot more test cases.
Here we’re only testing one cache-aware action on a publishing workflow, the second policy (known publisher action) is completely untested.
I’d recommend adding at least one test case for the alternate policy and at least one per config type (`opt_in`, `opt_out`, `opt_out_string` and `not_configurable`) to test all branches.

assets/queries/cicd/github/cache_poisoning/query.rego
```diff
@@ -0,0 +1,195 @@
...
+import data.generic.common as common_lib
+import future.keywords.in
+
+# Known cache-aware actions that can be exploited in cache poisoning attacks
+# Format: {"action": "uses_pattern", "disable_field": "field_name", "disable_value": value, "type": "opt_in|opt_out"}
+cache_aware_actions := {
+	"actions/cache": {"disable_field": "lookup-only", "disable_value": true, "type": "opt_out"},
+	"actions/setup-java": {"disable_field": "cache", "disable_value": "", "type": "opt_in_string"},
+	"actions/setup-go": {"disable_field": "cache", "disable_value": false, "type": "opt_in"},
+	"actions/setup-node": {"disable_field": "cache", "disable_value": "", "type": "opt_in_string"},
```
MikaYuoadas: ⛔ **Must fix:** Not covering the [package-manager-cache introduced in v5](https://github.com/actions/setup-node?tab=readme-ov-file#breaking-changes-in-v5) for `action/setup-node`

assets/queries/cicd/github/cache_poisoning/query.rego
```diff
@@ -0,0 +1,195 @@
...
+# Known publisher actions that indicate artifacts are being published
+publisher_actions := {
+	"pypa/gh-action-pypi-publish",
+	"rubygems/release-gem",
+	"jreleaser/release-action",
+	"goreleaser/goreleaser-action",
+	"softprops/action-gh-release",
+	"release-drafter/release-drafter",
+	"googleapis/release-please-action",
+	"docker/build-push-action",
```
MikaYuoadas, line 41: 🤏 **Nitpick:** If `push: false` is set this wouldn’t really be a publisher actions (and zizmor isn’t actually reporting as one in this case).

MikaYuoadas (changes requested):
Added review comments for:
* `cache_poisoning`

assets/queries/cicd/github/bot_conditions/query.rego
```diff
@@ -0,0 +1,333 @@
...
+	doc := input.document[i]
+	job := doc.jobs[j].if
```
MikaYuoadas: 🤏 **Nitpick:** Naming it `job` is confusing, we would expect the whole "job" dict but instead it’s only the its child `if` condition.
Consider:
```suggestion
	job_if := doc.jobs[j].if
```

assets/queries/cicd/github/bot_conditions/query.rego
```diff
@@ -0,0 +1,333 @@
...
+
+	bot_check := check_bot_condition_in_expression(job)
+	bot_check.is_vulnerable
+
+	result := {
+		"documentId": doc.id,
+		"searchKey": sprintf("jobs.%s.if={{%v}}", [j, bot_check]),
```
MikaYuoadas: ⛔ **Must fix:** I think you’re rendering the whole `bot_check` object here whereas I think you want the raw `if` expression (found inside var `job` or `job_if` if you implement my suggestion above).
```suggestion
		"searchKey": sprintf("jobs.%s.if={{%v}}", [j, job]),
```

assets/queries/cicd/github/bot_conditions/query.rego
```diff
@@ -0,0 +1,333 @@
...
+}
+
+# Spoofable actor ID contexts
+spoofable_actor_id_contexts := {
+	"github.actor_id",
+	"github.event.pull_request.sender.id",
+}
+
+# All spoofable contexts (case-insensitive matching will be done)
+all_spoofable_contexts := spoofable_actor_name_contexts | spoofable_actor_id_contexts
```
MikaYuoadas: ❓ **Question:** I haven’t seen `all_spoofable_contexts` used anywhere, is it forgotten leftover from a previous implementation?

assets/queries/cicd/github/bot_conditions/query.rego
```diff
@@ -0,0 +1,333 @@
...
+	"github.actor",
+	"github.triggering_actor",
+	"github.event.pull_request.sender.login",
+}
+
+# Spoofable actor ID contexts
+spoofable_actor_id_contexts := {
+	"github.actor_id",
+	"github.event.pull_request.sender.id",
+}
```
MikaYuoadas: ❓ **Question:** Honestly I’m confused about this. They seem to be a collection of documented github generated fields (`github.actor`, `github.actor_id` and `github.triggering_actor`) and commonly wrong typo field (`github.event.pull_request.sender.login` and `github.event.pull_request.sender.id`) that you’ll never find in a github payload. 😕 
What are we looking for here?

assets/queries/cicd/github/bot_conditions/query.rego
MikaYuoadas: ⛔ **Must fix:** I’m not sure having both ids and names is any more secure than one or the other.
Instead I think the goal is to nudge the user towards using `github.event.pull_request.user.login` and/or `github.event.pull_request.user.id` (the initial PR author) instead of the often seen `github.actor`, `github.actor_id` and `github.triggering_actor` that are *NOT* the user creating the PR but instead the latest to have acted on it (allowing confused deputy attack by triggering a bot on a attacker-created PR).

Also we have access to the tree-sitter AST, using `walk()` to find conditions using the bad fields (last PR actor instead of PR initial author) will be far sturdier than doing string splitting (which will fail at first nested conditions with parenthesis).

assets/queries/cicd/github/bot_conditions/query.rego
```diff
@@ -0,0 +1,333 @@
...
+
+# Check if a value is a known bot actor ID
+is_bot_id(value) {
+	bot_actor_ids[value]
+}
+
+# Ignore negation edge case supposing that most users won't write check like `!(id != bot_id)`
+skip_negation(value) {
+	not startswith(value, "!")
+}
```
MikaYuoadas: 💬 **Remark:** You’ll always find user doing the wildest thing yyou thought would never happen! 😆

assets/queries/cicd/github/github_env/query.rego
ChouraquiBen: Right! Creating a ticket for this rn!

assets/queries/cicd/github/github_env/query.rego
ChouraquiBen: We cannot make the scan intelligent enough to target the step only if the input may have been influenced by the user, be it in this step or in a previous one. Because of this, I would let as is right now and if customers complain about the noise because the rule is being too conservative, tweak it to allow some commands

assets/queries/cicd/github/dangerous_triggers/query.rego
ChouraquiBen: Actually, I do agree with you a bunch here, I must admit that I sinned out of lack of patience here, but I will even go further and create a lib so that I can reuse the functions through the different rules that require them

assets/queries/cicd/github/dangerous_triggers/test/positive1.yaml
ChouraquiBen: Indeed, there should be a case for workflow run, but no, this is not the one:
This one checks if the scanner is able to capture the `pull_request_target` trigger as an object

assets/queries/cicd/github/cache_poisoning/query.rego
ChouraquiBen: Do you mean that the function is not that useful, given that it only uses `startswith`?
I would say it's clearer having a meaningful name but a comment could also do

MikaYuoadas (changes requested):

assets/queries/cicd/github/bot_conditions/query.rego
ChouraquiBen: Indeed, I forgot to remove it while testing, I'll do it rn

assets/queries/cicd/github/bot_conditions/query.rego
ChouraquiBen: 😭

assets/queries/cicd/github/unsound_contains_with_controllable_input/test/positive.yaml
ChouraquiBen: That's right, my confusion. I corrected the implementation

assets/queries/cicd/github/github_env/query.rego
MikaYuoadas: Fair enough 👍 
Let’s roll with it and see what feedback we get.

assets/queries/cicd/github/unpinned_images/query.rego
```diff
@@ -0,0 +1,168 @@
...
+
+# Case 1: Expression references matrix but matrix has no pinned images
+check_unpinned_expression(job, container_or_service) {
+	parsed_exprs := container_or_service._parsed_expressions_image[_]
+	parsed_exprs.parse_ok == true
+
+	# Check if expression references matrix
+	contains(parsed_exprs.raw, "matrix.")
+
+	matrix := regex.find_n("matrix\\.[a-zA-Z0-9_]+", parsed_exprs.raw, 1)[0]
```
MikaYuoadas: 💡 **Suggestion:** This regex extracts the first `matrix.*`it finds, but for complex expressions like `format('{0}:{1}', matrix.image, matrix.tag)` it would only check one of the two matrix references.

To avoid a complex evaluation of the expression (that’d be hard if not impossible in rego) we could simply anchor with `^` and `$` to restrict this path to simple `matrix.<field>` expressions and let anything more complex fall through to Case 2 (flagged unconditionally), which is the safer default.

```suggestion
	matrix := regex.find_n("^matrix\\.[a-zA-Z0-9_]+$", parsed_exprs.raw, 1)[0]
```

assets/queries/cicd/github/dangerous_triggers/test/positive1.yaml
MikaYuoadas: Hmm, yeah reading back my comment doesn’t make any sense, `workflow_run` isn’t even a `types` but instead a event name like `pull_request_target` and should be a level higher in `on:` map 🤦 

But thanks for understanding my old man rambling anyway and adding positive3 and positive4 to address testing `workflow_run` 🙏

assets/queries/cicd/github/dangerous_triggers/query.rego
MikaYuoadas: 🔥 **Nice:** The new implementation is very clean and readable, thanks 🙏

assets/queries/cicd/github/dangerous_triggers/query.rego
```diff
@@ -0,0 +1,22 @@
...
+	doc := input.document[i]
+
+	cicd_lib.has_dangerous_trigger(doc)
+	doc_name := object.get(doc, "name", "document")
+
+	result := {
+		"documentId": doc.id,
+		"searchKey": "on",
+		"issueType": "IncorrectValue",
+		"keyExpectedValue": "Workflow should not use pull_request_target or worflow_run triggers",
```
MikaYuoadas: 🤏 Nitpick: Small typo in user-facing string, "worflow_run" should be "workflow_run" (missing 'k').
```suggestion
		"keyExpectedValue": "Workflow should not use pull_request_target or workflow_run triggers",
```

assets/queries/cicd/github/cache_poisoning/query.rego
MikaYuoadas: Yes, I mean I personally would have just used `startswith()` when needed but I get your argument of self documenting code with clear name in the policy.
Let’s keep it that way 👍

assets/libraries/cicd.rego
```diff
@@ -1,2 +1,19 @@
...
+	trigger == dangerous_triggers[_]
+}
+
+has_dangerous_trigger(doc) {
+	trigger := doc.on[_]
+	trigger == dangerous_triggers[_]
+}
+
+has_dangerous_trigger(doc) {
+	is_object(doc.on[trigger])
```
MikaYuoadas: ⛔ **Must fix:** The `is_object` guard here is unnecessary and it will miss triggers declared with a null value which I think is a valid syntax (although not sure how frequent it is in real world, I found at least [GitbookIO/gitbook](https://github.com/GitbookIO/gitbook/blob/main/.github/workflows/deploy-preview.yaml#L3) and [rerun-io/rerun](https://github.com/rerun-io/rerun/blob/main/.github/workflows/auto_approve.yml#L4) which are fairly popular repositories using this pattern). e.g.:

```
  on:
    pull_request_target:
jobs:
  ...
```
In this case `doc.on["pull_request_target"]` is `null`, and `is_object(null)` is `false`, so the rule silently skips it.

```suggestion
  doc.on[trigger]
```

*Note: you may want to also add test cases with this syntax to avoid future regression as it’s easy to miss.*

MikaYuoadas (dismissed):
Thanks a lot for addressing all my comments, we’re almost there 😄 

I only have three comments left to address, but I already pre-approve so you can merge as soon as you address them 👍

assets/libraries/cicd.rego
ChouraquiBen: Wow, I would have thought that `object(null)` would be `true`! Nice catch, thanks!

assets/queries/cicd/github/unpinned_images/query.rego
ChouraquiBen: I am not sure about this: indeed, if we fail to parse more complex expressions, we may surface false positives. However, with this proposal, when the rule sees a more complex expression, it will put it under the rug, potentially ending in a false negative

assets/libraries/cicd.rego
MikaYuoadas: Yes, I think that’s a common assumption,I had to double check too.
Honestly it’s kind of a gotcha that the [rego builtin documentation](https://www.openpolicyagent.org/docs/policy-reference/builtins/types#builtin-types-is_object) doesn’t even explicitly address.

assets/queries/cicd/github/unpinned_images/query.rego
MikaYuoadas: Hmm my understanding was flawed, I misread that it would fall through to case 2, but I was wrong as there was a `not contains()` that specifically excluded any matrix references.
So it would indeed cause false negatives. Your new approach is better, it limits the false negative, thank you 👍

MikaYuoadas (approved):

MikaYuoadas (approved):

PR #46, opened 2026-03-18
[Helm] Location dropped issue

whitemerch:
## Motivation

When scanning Helm charts, findings were dropped with **"Invalid resource location for file"** and never appeared in the report. This happened for templates where the finding was about **missing** configuration (e.g. `containers_running_as_root`): the line detector walks the search key path (e.g. `spec.template.spec.containers.securityContext.runAsNonRoot`), and when the **last** key part does not exist(securityContext for eg.) in the template it returned a zeroed location. The SARIF builder then rejected those findings (`Start.Line < 1`).

## Changes

- In `pkg/detector/helm/helm_detect.go`, only update `start`/`end` when the current key part is found. When the final part of the search key is missing (e.g. `runAsNonRoot`), we now keep the last valid location (e.g. the `securityContext:` line) instead of zeroing it, so the finding is no longer dropped.
- Use `Line: i + 1` in `detectCurrentLine` so locations pass SARIF validation (`Start.Line >= 1`) and match the rest of the pipeline for Helm

## Author Checklist

## QA Instruction

1. Run the scanner on a repo that contains any Helm chart where findings target missing keys, e.g. `securityContext.runAsNonRoot`.
2. **Before the fix:** Such findings would log "Invalid resource location for file" and not appear in the SARIF report.
3. **After the fix:** Findings appear in the report with a valid location (file + line) pointing to the last matched key (e.g. the `securityContext:` line).
4. Run `go test ./pkg/detector/helm/... -v` and confirm all tests pass.

## Blast Radius

- **DataDog IaC Scanner only.** Helm line detection and SARIF report generation for Helm charts

## Additional Notes

### DIFF
--- a/pkg/detector/helm/helm_detect.go
+++ b/pkg/detector/helm/helm_detect.go
@@ -71,11 +71,14 @@ func (d DetectKindLine) DetectLine(ctx context.Context, file *model.FileMetadata
 	// Since we are only looking at keys we can ignore the second value passed through '=' and '[]'
 	for _, key := range strings.Split(sanitizedSubstring, ".") {
 		substr1, _ := detector.GenerateSubstrings(ctx, key, extractedString, lines, curLineRes.lineRes)
-		curLineRes, start, end = curLineRes.detectCurrentLine(lines, fmt.Sprintf("%s:", substr1), "", true, file.IDInfo, helmID)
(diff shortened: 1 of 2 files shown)


MikaYuoadas (approved):
LGTM

PR #47, opened 2026-03-18
Fix mutex deadlock in filesystem.go

ChouraquiBen:
## Motivation

Scans were failing, often because of a `context deadline exceeded`. This is due to a deadlock in mutex usage:

If a repo with a folder that matched the terracache regex (`XXX.terraXXX`), the mutex already held by the `checkCondition` function would then be tried to be locked by `AddExcluded` inside the same goroutine, resulting in a deadlock.

The SourceProvider is not stateless. Making it so ended up making the code messy and so the resulting PR is compromising:
- The `excludes` is populated once when creating the source provider
- When sourcing starts, the source provider copies the state and passes it to each function to prevent concurrency.

## Changes

- The exclusion attribute of the filesystem is only used to store the exclusion parameters from the launch of the scan
- This attribute is copied once when the sourcing starts and then the copy is passed as argument through the rest of the sourcing.

## Author Checklist

## QA Instruction

- No more deadlocks in mutex
- The tests should pass 

## Blast Radius

`iac-scanning`

## Additional Notes

### DIFF
--- a/pkg/engine/provider/filesystem.go
+++ b/pkg/engine/provider/filesystem.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"fmt"
 	ioFs "io/fs"
+	"maps"
 	"os"
 	"path/filepath"
 	"regexp"
@@ -26,7 +27,8 @@ import (
 // FileSystemSourceProvider provides a path to be scanned
 // and a list of files which will not be scanned
 type FileSystemSourceProvider struct {
-	paths    []string
+	paths []string
+	// Only used to store the exclusion from the scanParameters
 	excludes map[string][]os.FileInfo
 	mu       sync.RWMutex
 }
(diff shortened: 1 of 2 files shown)


pkg/engine/provider/filesystem.go
```diff
@@ -69,7 +71,7 @@ func NewFileSystemSourceProvider(ctx context.Context, paths, excludes []string)
 }
 
 // AddExcluded add new excluded files to the File System Source Provider
-func (s *FileSystemSourceProvider) AddExcluded(ctx context.Context, excludePaths []string) error {
+func (s *FileSystemSourceProvider) AddExcluded(ctx context.Context, excludes map[string][]os.FileInfo, excludePaths []string) error {
```
jacobotb, line 74: I think this PR is a little too complicated with all the copying. I think it's a bit easier if you note that `AddExcluded` is not used outside of this file, so you can make it private, and then you can say "you always need to hold a lock _before_ you call this function". And then you change the RWMutex to a regular Mutex (RW mutexes are more trouble than they are worth -- everybody recommends you use a regular mutex first and only change to a RW mutex if it turns out you _really_ need it.)

ChouraquiBen: Yeah, I thought it would be cleaner not to use any mutex at all, thus the copy and adding the excludes to all the different functions

PR #54, opened 2026-03-23
Fix Terraform CloudWatch resource name attribution

whitemerch:
## Summary
- Replace hard-coded `\"resourceName\": \"unknown\"` in Terraform AWS CloudWatch alarm-missing rules with deterministic attribution.
- Use resource-count branching (`count([n | resources[n]])`) so documents with existing log metric filters report against a named filter resource; otherwise report against document id.

## Test plan

Test the impacted rules against its tests, and output a SARIF for them to check on the resource names

### DIFF
--- a/assets/queries/terraform/aws/cloudwatch_aws_config_configuration_changes_alarm_missing/query.rego
+++ b/assets/queries/terraform/aws/cloudwatch_aws_config_configuration_changes_alarm_missing/query.rego
@@ -1,6 +1,7 @@
 package Cx
 
-import data.generic.common as commonLib
+import data.generic.common as common_lib
+import data.generic.terraform as tf_lib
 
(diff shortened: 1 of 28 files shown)


MikaYuoadas (dismissed):
LGTM 👍

ChouraquiBen (commented):
I tend to disagree with the changes here: At some point, I tackled a few of the cloudwatch rules and noticed that the way they are written is a lot unclear, difficult to understand, like those and also have the issue of not returning easily the faulty resource name.
I would refactor the rules by unwrapping the checks: rn, the rule tries to check all the resources in one go. Instead, imho, it is better to use the idiomatic rego iteration to check each resource at a time

ChouraquiBen (approved):
LGTM!

PR #59, opened 2026-03-23
Helm Multi Chart resolution

whitemerch:
## Motivation

When a scan root contains **multiple sibling Helm chart directories**, a single global “resolved” flag caused only the first chart to be sent through the Helm resolver. Later charts were skipped for rendering, so their template files could be parsed as plain YAML instead, leading to noisy failures and inconsistent behavior depending on walk order.

## Changes

- Track resolved chart roots as a path list instead of one boolean.
- Skip only paths under an already-rendered chart; sibling charts are still resolved independently.

## Author Checklist

## QA Instruction

- `go test ./pkg/engine/provider/...`
- Optional: run the scanner against a directory that contains **two or more** chart roots (each with `Chart.yaml`) under the same parent and confirm both are handled without YAML parse errors on raw templates when Helm render succeeds, I have some real repository examples I can provide

## Blast Radius

This PR only affects how Helm charts are discovered and excluded during directory walks.

## Additional Notes

### DIFF
--- a/pkg/engine/provider/filesystem.go
+++ b/pkg/engine/provider/filesystem.go
@@ -155,7 +155,7 @@ func (s *FileSystemSourceProvider) GetSources(ctx context.Context,
 			continue
 		}
 
-		err = s.walkDir(ctx, scanPath, false, sink, resolverSink, extensions)
+		err = s.walkDir(ctx, scanPath, sink, resolverSink, extensions)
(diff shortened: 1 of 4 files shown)


MikaYuoadas (approved):
LGTM

ChouraquiBen (approved):
LGTM!

PR #63, opened 2026-03-25
Use sync.Pool to decrease memory usage on the parsing

ChouraquiBen:
## Motivation

IaC scanning tends to use lots of memory scanning big repositories. This PR aims to decrease the memory load by using a sync.Pool

## Changes

- Added a sync.Pool to reuse objects instead of creating new objects each time

## Author Checklist

  - Only added a sync pool that does not change the logic. The tests should pass
  - Logic does not change

## QA Instruction

## Blast Radius

IaC scanning

## Additional Notes

### DIFF
--- a/pkg/kics/service.go
+++ b/pkg/kics/service.go
@@ -67,6 +67,13 @@ type Service struct {
 	MaxFileSize    int
 }
 
+var bufferPool = sync.Pool{
+	New: func() any {
+		b := make([]byte, mbConst)
+		return &b
+	},
+}
+
 // PrepareSources will prepare the sources to be scanned
 func (s *Service) PrepareSources(ctx context.Context,
 	scanID string,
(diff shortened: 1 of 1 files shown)


pkg/kics/service.go
```diff
@@ -79,31 +86,34 @@ func (s *Service) PrepareSources(ctx context.Context,
 	// CxSAST query under review
 	contextLogger.Info().Msgf("Getting sources")
 	var err error
+
+	var sink = func(ctx context.Context, filename string, rc io.ReadCloser) error {
+		data := bufferPool.Get().(*[]byte)
+		if data == nil {
+			dataSlice := make([]byte, mbConst)
+			data = &dataSlice
+		}
```
MikaYuoadas: ⛔ **must fix:** Pretty sure [bufferPool.Get()](https://pkg.go.dev/sync#Pool.Get) can never return a nil (either there’s something to return from the pool or it’ll call bufferPool.New to create a new buffer).
So this is dead code that should be removed.

```suggestion
```

pkg/kics/service.go
```diff
@@ -67,6 +67,13 @@ type Service struct {
 	MaxFileSize    int
 }
 
+var bufferPool = sync.Pool{
+	New: func() interface{} {
```
MikaYuoadas: 🤏 **nitpick:** You could use [any](https://pkg.go.dev/builtin#any) instead of `interface{}` which is strictly equivalent but reads better imho.

```suggestion
	New: func()any {
```

MikaYuoadas (changes requested):
Left a nitpick and a dead code that should be removed.

More generally this PR claims to decrease memory usage: are you sure that’s actually the case rather than just maybe saving some CPU cycles on allocations overhead and GC?
For performance PR I’d ideally lik to have some benchmark result or before/after graphs to highlight the actual gains.

pkg/kics/service.go
ChouraquiBen: Thanks for the comments! I have changed according to them!

PR #65, opened 2026-03-31
Decreased max num workers and make it configurable

ChouraquiBen:
## Motivation

IaC Scanning is using a lot of memory, and the parallel parsing is only going to increase the load: the goal of this PR is to make this impact configurable by allowing to set the number of workers

## Changes

- Added a `MaxNumWorkers` scan parameter
- Added a `max-num-workers` flag

## Author Checklist

## QA Instruction

- Check the default flag value
- Check that the value is indeed being used within the source provider

## Blast Radius

`iac-scanning`

## Additional Notes

### DIFF
--- a/cmd/scanner/scan.go
+++ b/cmd/scanner/scan.go
@@ -73,6 +73,11 @@ var scanAction = &cli.Command{
 			Usage:   "a list of platform types to scan",
 			Value:   GetSupportedPlatforms(),
 		},
+		&cli.IntFlag{
+			Name:  "max-num-workers",
+			Usage: "Max workers running parallel jobs for parallel parsing",
+			Value: 32,
+		},
 		&cli.BoolFlag{
 			Name:   "x-parallelparsing",
 			Hidden: true,
@@ -142,6 +147,7 @@ func runScan(ctx context.Context, c *cli.Command) error {
 		ExcludePaths:               config.ExcludePaths,
 		IncludeQueries:             config.IncludeQueries,
 		DownloadQueriesFromDatadog: c.Bool("x-downloadqueriesfromdatadog"),
+		MaxNumWorkers:              c.Int("max-num-workers"),
 	}
 
 	metadata, err := console.ExecuteScan(ctx, params)
--- a/pkg/engine/provider/filesystem.go
+++ b/pkg/engine/provider/filesystem.go
@@ -26,14 +26,14 @@ import (
 // FileSystemSourceProvider provides a path to be scanned
 // and a list of files which will not be scanned
 type FileSystemSourceProvider struct {
-	paths    []string
-	excludes map[string][]os.FileInfo
-	mu       sync.RWMutex
+	paths         []string
+	excludes      map[string][]os.FileInfo
+	mu            sync.RWMutex
+	maxNumWorkers int
 }
 
 const (
 	minNumWorkers = 4
-	maxNumWorkers = 64
 )
 
 var (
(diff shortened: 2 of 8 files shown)


pkg/engine/provider/filesystem.go
```diff
@@ -293,8 +294,8 @@ func (s *FileSystemSourceProvider) calculateWorkerCount() int {
 	if numWorkers < 1 {
 		numWorkers = minNumWorkers
 	}
-	if numWorkers > maxNumWorkers {
-		numWorkers = maxNumWorkers
+	if numWorkers > s.maxNumWorkers {
+		numWorkers = s.maxNumWorkers
```
MikaYuoadas, line 298: ⛔ **must fix:** invalid value (e.g. negative or 0) will break the scanner. You should either clamp it to `minWorkers` (and log a warning that the user provided value was overridden) or validate the input at the cli level.

If you go the validation route, you can look at [cli lib flag actions](https://cli.urfave.org/v3/examples/flags/advanced/#flag-actions) to do it.

cmd/scanner/scan.go
```diff
@@ -73,6 +73,11 @@ var scanAction = &cli.Command{
 			Usage:   "a list of platform types to scan",
 			Value:   GetSupportedPlatforms(),
 		},
+		&cli.IntFlag{
+			Name:  "max-num-workers",
+			Usage: "Max workers running parallel jobs for parallel parsing",
```
MikaYuoadas, line 78: 🤏 **nitpick:** usage string is a bit redundant

```suggestion
			Usage: "Max number of workers for parallel file parsing",
```

MikaYuoadas (changes requested):
I’m not gonna lie, I’m not a fan of this change in its current form. 🤔 

It’s really confusing for the end user what this flag actually do.
They’ll expect it to set the number of worker used for any steps of the scan to this specific value, but instead what happens:
* Only affect the file processing (reading + parsing). don’t affect other worker pool that are controlled by the hardcoded `ParallelScanFlag` (no current way to set this flag from CLI so it’ll default to 0 -> auto-detect -> `runtime.GOMAXPROCS(-1)` -> CPU count):
  * [File analysis](https://github.com/DataDog/datadog-iac-scanner/blob/31aff7d1dac27b710d38bf1f285b871d850d697a/pkg/analyzer/analyzer.go#L336-L356) which detect file type
  * [Query inspection](https://github.com/DataDog/datadog-iac-scanner/blob/9472e3a8b8fcb695cc8f494cbff98b2ad7629014/pkg/engine/inspector.go#L316-L325) which run the queries against the file contents.
* Don’t actual set the worker pool size: awkwardly just set an upper-bound to the auto-detected value (`runtime.GOMAXPROCS(-1)` -> CPU count) which for `iac-scanning` is [currently set to 13](https://app.datadoghq.com/logs?query=service%3Aiac-scanning%20dd-maxprocs&agg_m=count&agg_m_source=base&agg_t=count&cols=host%2Cservice&messageDisplay=inline&refresh_mode=sliding&storage=flex_tier&stream_sort=desc&viz=stream&from_ts=1774969538213&to_ts=1774973138213&live=true) (the pod milli CPU limit rounded up) so 32 will actually not reduce it further. So I’m not expecting anything to change except if we set it far lower (<13) than the proposed default (32).

More generally before going headfirst in doing memory optimization, I think we should first profile the memory usage of the application to have a better feel of where the actual memory hogs are.
For example:
 * [pprof](https://pkg.go.dev/net/http/pprof) for go builtin tooling
 * [Datadog’s profiler](https://docs.datadoghq.com/profiler/)

PR #78, opened 2026-04-10
fix: extend SNS public access rule to cover aws_sns_topic_policy

valeryjuli:
## Summary
- Extends `sns_topic_is_publicly_accessible` Rego rule to also detect public access on `aws_sns_topic_policy` resources using the dynamic resource set pattern
- Adds positive and negative test cases for the new resource type

### DIFF
--- a/assets/queries/terraform/aws/sns_topic_is_publicly_accessible/query.rego
+++ b/assets/queries/terraform/aws/sns_topic_is_publicly_accessible/query.rego
@@ -3,8 +3,11 @@ package Cx
 import data.generic.common as common_lib
 import data.generic.terraform as tf_lib
 
+resource_type := {"aws_sns_topic", "aws_sns_topic_policy"}
+
 CxPolicy[result] {
-	resource := input.document[i].resource.aws_sns_topic[name]
+	res_type := resource_type[_]
+	resource := input.document[i].resource[res_type][name]
 
 	policy := common_lib.json_unmarshal(resource.policy)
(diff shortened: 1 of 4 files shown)


assets/queries/terraform/aws/sns_topic_is_publicly_accessible/query.rego
whitemerch, line 35: Do you want to update this too? even though `aws_sns_topic_policy` isn't used this much through modules

whitemerch (approved):
Thanks for the fix, good to go!

One process note: when we fix rules like this, we should check the equivalent rule across platforms (CloudFormation, Ansible, etc.) to see if the same gap exists. I'd like us to codify this as a guideline (or agent skill) going forward.

In this case, no cross-platform changes are needed:

- CloudFormation already checks `AWS::SNS::TopicPolicy`, in CFN the policy is always a separate resource, not inline on `AWS::SNS::Topic`

- Ansible checks `community.aws.sns_topic / sns_topic`, AKA there's no separate `sns_topic_policy` module, the policy is always managed within the topic resource

assets/queries/terraform/aws/sns_topic_is_publicly_accessible/query.rego
valeryjuli: `aws_sns_topic_policy` is a resource that references a topic by ARN, so I think it doesn't make sense as a module parameter (claude neither). Module support already is using the `aws_sns_topic`

PR #79, opened 2026-04-10
fix: extend db_security_group rules to support aws_security_group

valeryjuli:
## Summary
- Extended `db_security_group_has_public_interface`, `db_security_group_open_to_large_scope`, and `db_security_group_with_public_scope` rules to also detect `aws_security_group` resources (the modern replacement for the deprecated `aws_db_security_group`)
- Refactor rules to use a shared `resource_types` set
- Fix `resourceType` in `default_security_groups_with_unrestricted_traffic` (`aws_db_security_group` → `aws_default_security_group`)
- Add positive and negative test fixtures for all three rules using `aws_security_group`
- Add `CLAUDE.md` with cross-platform rule consistency guidelines

## Test plan
- [x] All `db_security_group_*` query tests pass
- [x] `default_security_groups_with_unrestricted_traffic` tests pass
- [x] Audited CloudFormation equivalents  already cover both `AWS::EC2::SecurityGroup` and `AWS::RDS::DBSecurityGroup`
image


assets/queries/terraform/aws/db_security_group_has_public_interface/query.rego
```diff
@@ -3,42 +3,117 @@ package Cx
 import data.generic.common as common_lib
 import data.generic.terraform as tf_lib
 
+resource_types := {"aws_db_security_group", "aws_security_group"}
```
whitemerch: Adding `aws_security_group` here will cause overlap with the existing [unrestricted_security_group_ingress](https://github.com/DataDog/datadog-iac-scanner/blob/main/assets/queries/terraform/aws/unrestricted_security_group_ingress/query.rego) rule, which already detects `aws_security_group` resources with `cidr_blocks` containing `0.0.0.0/0` and `ipv6_cidr_blocks` containing `::/0`.

A single aws_security_group with cidr_blocks = ["0.0.0.0/0"] would now fire both this rule and `unrestricted_security_group_ingress`, producing duplicate findings for the user. Is this intentional?

assets/queries/terraform/aws/db_security_group_has_public_interface/query.rego
```diff
@@ -3,42 +3,117 @@ package Cx
...
 	}
 }
 
+# Single ingress with cidr_blocks field (aws_security_group)
 CxPolicy[result] {
-	resource := input.document[i].resource.aws_db_security_group[name]
+	res_type := resource_types[_]
+	resource := input.document[i].resource[res_type][name]
+	cidrValue := public_cidrs[_]
+	contains(resource.ingress.cidr_blocks[j], cidrValue)
```
whitemerch: `public_cidrs` includes `::/0`, but `::/0` will never appear in `cidr_blocks`, it belongs in `ipv6_cidr_blocks`. This means the `::/0` iteration over `cidr_blocks` is a dead code path that can never match. The ipv6_cidr_blocks rules below already handle ::/0 correctly.

Same issue applies to the array variant on line 71.

assets/queries/terraform/aws/db_security_group_has_public_interface/test/positive_expected_result.json
whitemerch, line 1: This file adds tests for `cidr_blocks` in `positive3.tf` and `positive4.tf`, but there are no test fixtures for the two ipv6_cidr_blocks rules (lines 85-101 and 103-119 in the query). Should we add a positive5.tf with:

```
resource "aws_security_group" "positive5" {
  ingress {
    from_port        = 5432
    to_port          = 5432
    protocol         = "tcp"
    ipv6_cidr_blocks = ["::/0"]
  }
}
```

Without this, those two CxPolicy rules have zero test coverage

assets/queries/terraform/aws/db_security_group_with_public_scope/query.rego
whitemerch, line 1: This rule is inconsistent with `db_security_group_has_public_interface`:

- `has_public_interface` handles both `resource.ingress.cidr` (line 14) and `resource.ingress[idx].cidr` (line 33), but this rule only handles the single-ingress form (line 12). An `aws_db_security_group` with multiple ingress blocks won't be caught

- `has_public_interface` has dedicated rules for `ipv6_cidr_blocks`, but this rule doesn't check for `::/0` at all in `ipv6_cidr_blocks`. If we're extending to aws_security_group, IPv6 public scope should also be detected

Additionally, this rule only checks `== "0.0.0.0/0"` but the original `has_public_interface` rule checks both `0.0.0.0/0` and `::/0` via `public_cidrs`. The `cidr` field on `aws_db_security_group` could theoretically hold `::/0`, should this rule also check for it, for parity?

assets/queries/terraform/aws/db_security_group_open_to_large_scope/query.rego
whitemerch, line 1: Two concerns:

- The "open to large scope" check (CIDR prefix <= 24) makes sense for database-specific security groups, where broad network access is almost always wrong. But for generic `aws_security_group`, a /24 CIDR on an ingress rule is very common and often intentional (e.g. allowing a subnet's traffic to a web server). This will produce a lot of false positives. Is this the right threshold for generic SGs, or should this rule stay scoped to `aws_db_security_group` only?

- Like the original single-rule, this doesn't handle `resource.ingress[idx].cidr` for `aws_db_security_group` resources with multiple ingress blocks. The other two rules in this PR have array ingress variants, and this should too

assets/queries/terraform/aws/default_security_groups_with_unrestricted_traffic/query.rego
```diff
@@ -8,7 +8,7 @@ CxPolicy[result] {
 
 	result := {
 		"documentId": input.document[i].id,
-		"resourceType": "aws_db_security_group",
+		"resourceType": "aws_default_security_group",
```
whitemerch, line 11: Niceeee!

CLAUDE.md
```diff
@@ -0,0 +1,41 @@
...
+
+# Build and scan a directory
+make build
+./bin/datadog-iac-scanner scan -p <path> -o <output-dir> -t Terraform
+```
+
+## Writing Rules
+
+- Each rule needs: `query.rego`, `metadata.json`, and test fixtures (`positive*.tf`, `negative*.tf`, `positive_expected_result.json`)
+- Use `contains()` for list fields (e.g., `cidr_blocks`) vs `==` for scalar fields (e.g., `cidr`)
```
whitemerch: I guess that's a bit too specific on this PR, the skill should remain agnostic

CLAUDE.md
whitemerch, line 1: Can we make it an agnostic skill and not only claude please? so it can be picked up by cursor too

whitemerch (changes requested):

assets/queries/terraform/aws/db_security_group_has_public_interface/query.rego
valeryjuli: Good catch, I reverted the `aws_security_group` addition from all three `db_security_group_*` rules and kept them scoped to `aws_db_security_group` only

assets/queries/terraform/aws/db_security_group_has_public_interface/query.rego
valeryjuli: You're right, removed on last version

assets/queries/terraform/aws/db_security_group_has_public_interface/test/positive_expected_result.json
valeryjuli: Idem

assets/queries/terraform/aws/db_security_group_with_public_scope/query.rego
valeryjuli: I've added `resource.ingress[idx].cidr` handling for `aws_db_security_group`. And added the `::/0` check on the `cidr` field for parity with `has_public_interface`

assets/queries/terraform/aws/db_security_group_open_to_large_scope/query.rego
valeryjuli: Agree, I'm keeping this rule scoped to `aws_db_security_group` only. Also added the missing array ingress variant.

CLAUDE.md
valeryjuli: Fair point, I've generalized the writing rules section to keep it agnostic

CLAUDE.md
valeryjuli: I've moved the guidelines into `AGENTS.md` (which Cursor and other tools pick up) and have `CLAUDE.md` reference it so both tools work.

assets/queries/terraform/aws/db_security_group_open_to_large_scope/query.rego
```diff
@@ -13,7 +15,24 @@ CxPolicy[result] {
...
 		"issueType": "IncorrectValue",
-		"keyExpectedValue": "'aws_db_security_group.ingress.cidr' > 24",
-		"keyActualValue": "'aws_db_security_group.ingress.cidr' <= 24",
+		"keyExpectedValue": sprintf("'aws_db_security_group[%s].ingress.cidr' CIDR prefix > 24", [name]),
+		"keyActualValue": sprintf("'aws_db_security_group[%s].ingress.cidr' CIDR prefix <= 24", [name]),
+	}
+}
+
+# Array ingress with cidr field
+CxPolicy[result] {
```
valeryjuli, line 23: added the second rule for the array ingress pattern

assets/queries/terraform/aws/db_security_group_with_public_scope/query.rego
```diff
@@ -3,21 +3,44 @@ package Cx
...
 import data.generic.terraform as tf_lib
 
+public_cidrs := {"0.0.0.0/0", "::/0"}
+
+# Single ingress with cidr field
 CxPolicy[result] {
-	resource := input.document[i].resource.aws_db_security_group[name].ingress
-	resource.cidr == "0.0.0.0/0"
+	resource := input.document[i].resource.aws_db_security_group[name]
+	cidrValue := public_cidrs[_]
```
whitemerch, line 11: The rule now matches `cidrValue := public_cidrs[_]` which includes both "0.0.0.0/0" and "::/0", but the error messages are hardcoded:

```
"keyExpectedValue": sprintf("'aws_db_security_group[%s].ingress.cidr' != 0.0.0.0/0", [name]),
"keyActualValue": sprintf("'aws_db_security_group[%s].ingress.cidr' = 0.0.0.0/0", [name]),
```

When `cidrValue` is ::/0, the user sees "= 0.0.0.0/0" which is wrong. We should maybe interpolate the matched value like `has_public_interface` does:

```
"keyExpectedValue": sprintf("'aws_db_security_group[%s].ingress.cidr' should not be '0.0.0.0/0' or '::/0'", [name]),
"keyActualValue": sprintf("'aws_db_security_group[%s].ingress.cidr' is '%s'", [name, resource.ingress.cidr])
```

Same issue on lines 37-38 for the array variant.

assets/queries/terraform/aws/db_security_group_with_public_scope/query.rego
whitemerch: The module rule still only checks == "0.0.0.0/0" but the resource rules above now check both 0.0.0.0/0 and ::/0 via public_cidrs. Should this also use public_cidrs for parity?

assets/queries/terraform/aws/db_security_group_open_to_large_scope/query.rego
```diff
@@ -1,10 +1,12 @@
 package Cx
 
+import data.generic.common as common_lib
```
whitemerch: nit: `common_lib` is not used, can we remove it?

assets/queries/terraform/aws/db_security_group_with_public_scope/query.rego
valeryjuli: Fixed now says "should not be '0.0.0.0/0' or '::/0'" and `keyActualValue` interpolates the actual matched CIDR via `resource.ingress.cidr` / `resource.ingress[idx].cidr`.

assets/queries/terraform/aws/db_security_group_with_public_scope/query.rego
valeryjuli: Yup, fixed - the module rule now uses `cidrValue := public_cidrs[_]`

assets/queries/terraform/aws/db_security_group_open_to_large_scope/query.rego
valeryjuli: Removed :)

whitemerch (approved):
Let's goooo

PR #80, opened 2026-04-13
Fix Terraform policy line reporting for jsonencode and SARIF regions

whitemerch:
> [!TIP]

## Motivation

Findings on Terraform policies that use `jsonencode()` (or JSON-shaped policy strings) could report the wrong source line: the detector did not always resolve `Statement[i].Principal` inside nested expressions, and SARIF could prefer an empty remediation snippet over the resource span. This change aligns reported lines with the actual policy fields and keeps SARIF regions consistent when no remediation text exists.

## Changes

- SARIF: use remediation insertion locations only when remediation content is non-empty; otherwise keep the resource span.
- Terraform: improve list/array resolution for `Statement` entries inside `jsonencode` and JSON array syntax; small refactors in array-start detection.
- Rego: extend `searchKey` paths for several AWS policy rules so line resolution targets `Statement[idx].Principal` (and related keys) where applicable; update expected test fixtures.

## Author Checklist

## QA Instruction

Run unit tests for affected packages, for example:

`go test ./pkg/report/model/... ./pkg/detector/terraform/...`

Run the scanner on a small Terraform file that uses `jsonencode` for `aws_s3_bucket_policy` (or similar) with `Principal = "*"` and confirm the reported line matches that attribute line and SARIF regions behave as expected.

## Blast Radius

Affects the open-source IaC scanner: SARIF output shape for some findings, Terraform line detection for policy-related rules, and bundled Rego query metadata used for fingerprints and locations

## Additional Notes

### DIFF
--- a/assets/queries/terraform/aws/sqs_queue_exposed/query.rego
+++ b/assets/queries/terraform/aws/sqs_queue_exposed/query.rego
@@ -6,12 +6,7 @@ import data.generic.terraform as tf_lib
 CxPolicy[result] {
 	resource := input.document[i].resource.aws_sqs_queue[name]
 
-	policy := common_lib.json_unmarshal(resource.policy)
-	st := common_lib.get_statement(policy)
-	statement := st[idx]
-
-	common_lib.is_allow_effect(statement)
-	tf_lib.anyPrincipal(statement)
+	idx := exposed_statement_indices(resource.policy)[_]
 
 	result := {
 		"documentId": input.document[i].id,
@@ -29,12 +24,7 @@ CxPolicy[result] {
 	module := input.document[i].module[name]
 	keyToCheck := common_lib.get_module_equivalent_key("aws", module.source, "aws_sqs_queue", "policy")
 
-	policy := common_lib.json_unmarshal(module[keyToCheck])
-	st := common_lib.get_statement(policy)
-	statement := st[idx]
-
-	common_lib.is_allow_effect(statement)
-	tf_lib.anyPrincipal(statement)
+	idx := exposed_statement_indices(module[keyToCheck])[_]
 
 	result := {
 		"documentId": input.document[i].id,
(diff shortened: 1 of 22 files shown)


assets/queries/terraform/aws/sqs_queue_exposed/query.rego
```diff
@@ -6,43 +6,44 @@ import data.generic.terraform as tf_lib
...
+		"searchKey": sprintf("module[%s].%s.Statement[%d].Principal", [name, keyToCheck, idx]),
 		"issueType": "IncorrectValue",
 		"keyExpectedValue": "'policy.Principal' shouldn't get the queue publicly accessible",
 		"keyActualValue": "'policy.Principal' does get the queue publicly accessible",
-		"searchLine": common_lib.build_search_line(["module", name, "policy"], []),
+		"searchLine": common_lib.build_search_line(["module", name, keyToCheck, "Statement", idx, "Principal"], []),
 	}
 }
-
-exposed(policyValue) {
```
ChouraquiBen, line 41: 💡 Suggestion: instead of removing the function altogether, you could've returned a list of the index of the faulty `Principal` statements. This would also make it so that we only report one finding instead of multiple ones

ChouraquiBen (dismissed):
LGTM with suggestion: could you please add some tests alongside the queries with more than one `Statement`, have the second `Principal` flagged for example?

ChouraquiBen (approved):
One last thing comes to my mind and it is that the new test cases may have changed the documentation example if regenerated. Could you please create a follow-up PR? If you don't have the time, do not worry, I will do

whitemerch:
> One last thing comes to my mind and it is that the new test cases may have changed the documentation example if regenerated. Could you please create a follow-up PR? If you don't have the time, do not worry, I will do

Will do! Let me create a ticket for it

PR #86, opened 2026-04-14
Added new workflow to check if the documentation is up to date

ChouraquiBen:
## Motivation

When working on the rules, it is common to create new test cases, or change the description. It is as common to forget and update the documentation accordingly. As the documentation is not automatically generated, forgetting to update is easy. This PR aims to add a workflow checking if the documentation is up to date or not. This allows flexibility:

- If the PR is already big, the user may decide to generate the documentation change in another PR
- If making a slight modification, the user may decide to make the documentation change a part of the same PR

## Changes

- Create a workflow that uses the documentation generation script to check if the documentation is up to date
- Make the documentation up to date
  - Four rules changed capital letters in name
  - The rest of the changes is about the compliant / non-compliant code examples

## Author Checklist

## QA Instruction

- [Example of failing run](https://github.com/DataDog/datadog-iac-scanner/actions/runs/24454659104/job/71452078425?pr=86)

## Blast Radius

datadog-iac-scanner

## Additional Notes

### DIFF
--- a/.github/workflows/check-documentation-up-to-date.yaml
+++ b/.github/workflows/check-documentation-up-to-date.yaml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   check-documentation-up-to-date:
-    runs-on: ubuntu-latest
+    runs-on: macos-latest
     permissions:
       contents: read
     steps:
--- a/.github/workflows/check-documentation-up-to-date.yaml
+++ b/.github/workflows/check-documentation-up-to-date.yaml
@@ -3,6 +3,10 @@ name: check-documentation-up-to-date
 on:
   workflow_dispatch:
   pull_request:
+    paths:
+      - "assets/queries/**"
+      - "documentation/**"
+      - ".github/workflows/check-documentation-up-to-date.yaml"
 
 jobs:
   check-documentation-up-to-date:
--- a/.github/workflows/check-documentation-up-to-date.yaml
+++ b/.github/workflows/check-documentation-up-to-date.yaml
@@ -1,0 +1,58 @@
+name: check-documentation-up-to-date
+
+on:
+  workflow_dispatch:
+  pull_request:
+
+jobs:
+  check-documentation-up-to-date:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+        with:
+          persist-credentials: false
+
+      - name: Setup Python
+        uses: actions/setup-python@7f4fc3e22c37d6ff65e88745f38bd3157c663f7c # v4.9.1
+        with:
+          python-version: '3.x'
+
+      - name: Install dependencies
+        run: pip install pyyaml
+
+      - name: Generate documentation and validate no change
+        run: |
+          cd documentation
+
+          python local_gen.py ../assets/queries/ \
+          --resources-json "resources.json" \
+          --output-dir "rules" \
+          --list-json "list.json" \
+          --frontmatter-yaml "frontmatter.yaml"
+          
+          cd ..
+          if ! git diff --quiet; then
+            echo "ERROR: Documentation is out of date!"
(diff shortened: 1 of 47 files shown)


documentation/rules/ansible/aws/rds_db_instance_publicly_accessible.md
```diff
@@ -78,14 +79,19 @@ Secure example:
     password: "{{ password }}"
     allocated_storage: "{{ allocated_storage }}"
     publicly_accessible: Yes
-- name: Create RDS instance publicly accessible
+- name: community - Basic mysql provisioning example
```
drichards-87:
```suggestion
- name: community - Basic MySQL provisioning example
```

drichards-87 (dismissed):
Looks good! I added one small note and approved the PR.

whitemerch (dismissed):

.github/workflows/check-documentation-up-to-date.yaml
```diff
@@ -0,0 +1,58 @@
+name: check-documentation-up-to-date
+
+on:
+  workflow_dispatch:
+  pull_request:
+
+jobs:
+  check-documentation-up-to-date:
+    runs-on: macos-latest
```
MikaYuoadas, line 13: **❓ Question:** Why though?
I don’t see anything macos specific about this and [macos runners cost 10x more than linux runners](https://docs.github.com/en/billing/concepts/product-billing/github-actions#baseline-minute-costs) and they have had [a long history of poor performance](https://github.com/actions/runner-images/issues/1336) (users have reported anywhere from 2x to 10x slower processing/provisionning time).
If we don’t have a good reason to use `macos-latest` I’d stick with `ubuntu-latest`

.github/workflows/check-documentation-up-to-date.yaml
```diff
@@ -0,0 +1,58 @@
+name: check-documentation-up-to-date
+
+on:
+  workflow_dispatch:
+  pull_request:
```
MikaYuoadas, line 5: **💡 Suggestion:** Why not scope it so it only triggers on relevant changes using `paths`?

```suggestion
on:
  workflow_dispatch:
  pull_request:
    paths:
      - "assets/queries/**"
      - "documentation/**"
      - ".github/workflows/check-documentation-up-to-date.yaml"
```

MikaYuoadas (dismissed):
LGTM, just added two suggestions/questions

.github/workflows/check-documentation-up-to-date.yaml
ChouraquiBen: Yeah, definitely seems to be off the hat 😅... putting MacOS here is a magic trick that I pulled: the script to generate the documentation does not have anything to lock the code examples order: you may have `negative1.yaml` first but then `negative3.yaml` and so on... without this option, the order has to be locked. It is very easy to do, using a simple sort in the script, but that means that hundreds of files would need another review from both us and the documentation team to ensure that the documentation is still good to go.
Therefore this idea that I had to just force the OS. I tried and it does not really take that long, it is a matter of seconds and the PR is locked by other blocking jobs which take much longer.
In addition to that, this workflow is a temporary thing since the rules are moving to another repo, therefore the unwillingness to change the order of the code examples to ask the Docs team for huge pointless reviews.

.github/workflows/check-documentation-up-to-date.yaml
ChouraquiBen: Agreed! I will add this and rebase!

ChouraquiBen:
Will need to regenerate the documentation because of name mismatch in the frontmatter.yaml. Doing this asap

whitemerch (commented):
This macOS job currently runs on every PR, including changes that cannot affect generated docs. Since macos-latest is intentionally used here and is more expensive/slower than Linux runners, this adds CI cost and noise. I’d add pull_request.paths for `assets/queries/**`, `documentation/**`, the workflow file itself, and what's needed. What do u think?

whitemerch (approved):

MikaYuoadas (approved):

PR #88, opened 2026-04-15
Reduce scanner peak memory usage

MikaYuoadas:
## Summary

The scanner was hitting OOM on large repositories. Profiling on a test repo(28,715 files) showed a true peak of **2,549 MB** during OPA query evaluation, with the Go runtime reserving up to **2,909 MB** from the OS.

This PR is a first low impact effort to reduce memory usage. It applies 5 optimizations that together reduce peak memory with no functional regression.

### 1. Stream JSON report export
Replace `json.Marshal` + `json.Compact` + `f.Write` with `json.NewEncoder` streaming directly to file, avoiding keeping the full report in memory.

### 2. Set GOGC=50 for more aggressive garbage collection
The biggest lever here. OPA query evaluation creates a huge amount of transient allocations. Triggering GC at 50% heap growth (instead of the default 100%) reclaims them faster, preventing the heap from growing as large between cycles at the of additional GC cycles, ~4% throughput hit.

### 3. Skip JSON round-trip when building OPA AST payload
The inspector was serializing the combined Documents to JSON, deserializing back to `interface{}`, then converting to OPA AST. Since Documents are already `map[string]interface{}`, we can pass them directly to `ast.InterfaceToValue`, avoiding intermediate copies.

### 4. Use pointer-based FileMetadatas to avoid value copies
`FileMetadatas` was `[]FileMetadata`, causing full struct copies on every append and map insertion. Changed to `[]*FileMetadata` so `SaveFile`, `saveToFile`, `ToMap`, and `Combine` all share pointers. Also removed the unused `Content` field from `FileMetadata` (redundant with `OriginalData` and `LinesOriginalData`.

### 5. Compute ToMap() once and share across query workers
`files.ToMap()` was called inside `performInspection` for every query (~2,600 times). Now computed once before spawning workers and passed as a shared read-only map.

## Benchmarking

Profiled with `runtime.ReadMemStats` snapshots at pipeline stage boundaries plus a background goroutine polling heap usage every 200ms to capture the true in-stage peak (not just the post-GC snapshot).

Target test repo: 28,715 files, ~14,786 violations.

### Per-change incremental impact

| Step | True peak | Delta | heap_sys | Delta |
|---|---:|---:|---:|---:|
| Baseline | 2,549 MB | | 2,909 MB | |
| + Stream export | 2,422 MB | -127 MB | 2,853 MB | -56 MB |
| + GOGC=50 | 1,607 MB | -815 MB | 1,813 MB | -1,040 MB |
| + Skip JSON round-trip | 1,552 MB | -55 MB | 1,785 MB | -28 MB |
| + Pointer FileMetadatas | 1,403 MB | -149 MB | 1,597 MB | -188 MB |
| + ToMap() once | 1,371 MB | -32 MB | 1,565 MB | -32 MB |

### Final results

| Metric | Baseline | Final | Reduction |
|---|---:|---:|---:|
| **True peak** | 2,549 MB | **1,371 MB** | -1,178 MB |
| **heap_sys max** | 2,909 MB | **1,565 MB** | -1,344 MB |
| **Scan time** | 7m05s | 6m20s | -45s |
| **Violations** | 14,786 | 14,786 | 0 |

## Test plan

- [x] All 4,026 unit tests pass after each change
- [x] Violation count verified stable (14,786) across 3 baseline runs and after each strategy
- [x] Benchmarked incrementally on test repo (28,715 files) after each change
- [x] E2E test suite on CI
- [x] Manual staging validation for non-regression


pkg/engine/inspector.go
whitemerch, line 593: With the change from `map[string]model.FileMetadata` to `map[string]*model.FileMetadata`, isn't there nil pointer derefernce here?

pkg/report/commons.go
```diff
@@ -111,22 +109,7 @@ func ExportJSONReport(ctx context.Context, path, filename string, body interface
...
-	err = json.Compact(&minifiedJSON, bodyBytes)
-	if err != nil {
-		contextLogger.Err(err).Msg("Error minifying JSON:")
-		return err
-	}
-
-	_, err = f.Write(minifiedJSON.Bytes())
-
-	return err
+	return json.NewEncoder(f).Encode(body)
```
whitemerch, line 112: Should we be adapt anything related to [this](https://github.com/golang/go/issues/37083)?

cmd/scanner/main.go
```diff
@@ -15,7 +16,13 @@ import (
 
 const defaultFailCode = 126
 
+// gcPercent sets the garbage collection target percentage. A lower value makes
+// the GC run more frequently, reclaiming transient allocations from OPA query
+// evaluation faster and reducing peak memory usage.
+const gcPercent = 50
+
 func main() {
+	debug.SetGCPercent(gcPercent)
```
whitemerch: should we check GOGC before hand?

whitemerch (dismissed):
Lovely

pkg/report/commons.go
MikaYuoadas: Yes, I was actually also concerned when I found out about this "regression".
Most (if not all) json parser shouldn’t be bothered.
I specifically looked at how our downstream workflow read it and they simply `json.Unmarshal()` which will silently ignore any leading/trailing whitespace (per the [JSON RFC](https://datatracker.ietf.org/doc/html/rfc8259#section-2)).
So I’m not too worried but the E2E test in staging will make sure that is indeed the case.

cmd/scanner/main.go
MikaYuoadas: indeed, we probably want to respect user defined GOGC if available. I’ll update it.

MikaYuoadas: Added it 👍

pkg/engine/inspector.go
MikaYuoadas: It shouldn’t really happen in real life as this is set to `"documentId": input.document[i].id,` from the same values that are used to populate `qCtx.Files` map.
But (although unlikely as it would fails the unit tests) we could write a faulty rego policy that set it to something incorrect to trigger it.

I added a nil guard to "belt and suspenders" it anyway.

pkg/model/model.go
```diff
@@ -298,23 +297,23 @@ type Document map[string]interface{}
 func (m FileMetadatas) Combine(ctx context.Context, lineInfo bool) Documents {
```
whitemerch, line 297: The Combine method still mutates `f.Document / f.LineInfoDocument` in place (`f.Document["id"] = f.ID; f.Document["file"] = f.FilePath`). With pointer sharing this is now a shared mutation. If Combine is ever called twice on the same file set (e.g. in tests or a future retry path), the second call is a no-op, Not a blocker here, but just mentioning it

whitemerch (dismissed):
LGTM

whitemerch (approved):

PR #89, opened 2026-04-15
Catch reusable workflows unpinned actions

ChouraquiBen:
## Motivation

Github actions have a reusable workflow feature which allows to use an entire job from the same repo or another one. This works exactly the same as using a GitHub action.

The unpinned reused workflows used under a `job` key are not flagged by the scanner. The goal of this PR is to correct this behaviour

## Changes

- Allow the scanner to scan GitHub actions that do not have a `step` key
- Allow anonymous jobs and steps to be flagged
- Refactored the rule to also flag the `uses` under job
- Added tests around this new detection
- Updated the documentation

## Author Checklist

## QA Instruction

- Check the logic of the rule
- Check the test cases: do they match what needs to be scanned?

## Blast Radius

iac-scanning - unpinned actions rule findings

## Additional Notes

### DIFF
--- a/assets/queries/cicd/github/unpinned_actions_full_length_commit_sha/query.rego
+++ b/assets/queries/cicd/github/unpinned_actions_full_length_commit_sha/query.rego
@@ -18,7 +18,26 @@ CxPolicy[result] {
 		"keyActualValue": "Action is not pinned to a full length commit SHA.",
 		"searchLine": common_lib.build_search_line(["jobs", j, "steps", k, "uses"],[]),
 		"resourceType": "github_action",
-		"resourceName": input.document[i].jobs[j].steps[k].name
+		"resourceName": get_object_name(input.document[i].jobs[j].steps[k], "step", k)
+	}
+}
+
+CxPolicy[result] {
+
+	uses := input.document[i].jobs[j].uses
+	not isAllowed(uses)
+	not isPinned(uses)
+	not isRelative(uses)
+	
+	result := {
+		"documentId": input.document[i].id,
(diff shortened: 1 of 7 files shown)


domalessi (dismissed):

ganeshkumarsv (approved):
LGTM!

whitemerch (dismissed):

assets/queries/cicd/github/unpinned_actions_full_length_commit_sha/query.rego
```diff
@@ -36,3 +55,8 @@ isRelative(use){
     startswith(use,allowed[i])
 }
 
+get_object_name(object, object_title, index) := object_name {
+	object_name := object.name
+} else := object_name {
+	object_name := sprintf("%s-%d", [object_title, index])
```
MikaYuoadas: **⛔ Must fix:** I works for steps where index is an int but I think it’ll break for jobs as they are indexed by job name (a string).

```suggestion
	object_name := sprintf("%s-%v", [object_title, index])
```

assets/queries/cicd/github/unpinned_actions_full_length_commit_sha/test/negative5.yaml
```diff
@@ -0,0 +1,9 @@
+name: test-negative4
```
MikaYuoadas: **🤏 Nitpick:**

```suggestion
name: test-negative5
```

MikaYuoadas (changes requested):

MikaYuoadas (approved):

PR #94, opened 2026-04-16
Update served rule converter

jacobotb:
## Motivation

Follow the current state of server-side rules.

## Changes

Fixed bugs.

## Author Checklist

## QA Instruction

## Blast Radius

## Additional Notes


pkg/engine/source/datadog.go
```diff
@@ -273,20 +272,17 @@ func convertRule(rule *Rule) model.QueryMetadata {
...
-		out.Metadata["descriptionUrl"] = *rule.DocumentationUrl
-	}
-	if rule.DescriptionId != nil {
-		out.Metadata["descriptionId"] = *rule.DescriptionId
-	}
-	if rule.Provider != nil {
-		out.Metadata["cloudProvider"] = *rule.Provider
-	}
-	if rule.Cwe != nil {
+	setStringPtr(out.Metadata, "descriptionUrl", rule.DocumentationUrl)
```
ChouraquiBen, line 279: suggestion: add comments about the `Metadata` building process:
- One saying that the attributes above are necessary
- A second for the optional attributes 
- A third one for the `cwe` that must be there

ChouraquiBen: nit: Another idea is: instead of having three cases, would it work to refactor the `setStringPtr` to have another case:
```go
func setStringPtr[T ~string](m map[string]any, key string, v *T) {
	if v != nil {
		m[key] = string(*v)
	}
    else {
        m[key] = ""
    }
}
```

ChouraquiBen: Nah, forget it that would break the override

pkg/engine/source/datadog.go
```diff
@@ -295,47 +291,52 @@ func convertRule(rule *Rule) model.QueryMetadata {
...
 }
 
+func setStringPtr[T ~string](m map[string]any, key string, v *T) {
+	if v != nil {
+		m[key] = string(*v)
+	}
+}
+
+func frameworksToMeta(fs []Framework) []any {
+	result := make([]any, len(fs))
```
ChouraquiBen, line 328: nit: correct me if I am mistaken, but from a memory PoV and since result is not supposed to change, it would be better to specify a capacity:


```suggestion
	result := make([]any, 0 len(fs))
```

and then below
```go
for _, f := range fs {
	result = append(result, map[string]any{
		"framework":         f.Framework,
		"framework_version": f.Version,
		"requirement":       f.Requirement,
		"control":           f.Control,
	})
}
```

ChouraquiBen (approved):
I left a few comments but they are only nits, LGTM otherwise

pkg/engine/source/datadog.go
jacobotb: I don't see why it would be better. The current code preallocates the slice with the final size and then sets the elements directly, which is the same thing but more explicit.

pkg/engine/source/datadog.go
jacobotb: It's hard to find out which properties are really mandatory (for example, I just found out that "cwe" was.) I'll do some research and update with comments and code changes in a followup PR.

PR #96, opened 2026-04-17
Export the query reading and conversion functions

jacobotb:
## Motivation

So we can use them in other repositories.

## Changes

Export a few functions, expose the embedded query FS so we can call the reading functions on a different FS.

## Author Checklist

## QA Instruction

## Blast Radius

## Additional Notes

### DIFF
--- a/pkg/engine/source/datadog.go
+++ b/pkg/engine/source/datadog.go
@@ -199,7 +199,7 @@ func (s *DatadogSource) filterRules(ruleset *Ruleset, selection *QueryInspectorP
 		if !checkIncluded(rule, selection) || checkExcluded(rule, selection) {
 			continue
 		}
-		out = append(out, convertRule(rule))
+		out = append(out, ConvertRule(rule))
 	}
 	return out, nil
 }
@@ -258,8 +258,8 @@ func isInCaseInsensitiveNotEmptyList(id string, list []string) bool {
 }
 
 // nolint:gocyclo
-// convertRule converts a Datadog api [Rule] to a [model.QueryMetadata]
-func convertRule(rule *Rule) model.QueryMetadata {
+// ConvertRule converts a Datadog api [Rule] to a [model.QueryMetadata]
+func ConvertRule(rule *Rule) model.QueryMetadata {
 	out := model.QueryMetadata{
 		InputData: "{}",
 		Query:     rule.Name,
--- a/pkg/engine/source/datadog.go
+++ b/pkg/engine/source/datadog.go
@@ -284,7 +284,6 @@ func convertRule(rule *Rule) model.QueryMetadata {
 		out.Metadata["cwe"] = ""
 	} else {
 		out.Metadata["cwe"] = *rule.Cwe
-		out.CWE = *rule.Cwe
 	}
 	if rule.Aggregation != nil {
 		out.Metadata["aggregation"] = *rule.Aggregation
--- a/pkg/engine/source/filesystem.go
+++ b/pkg/engine/source/filesystem.go
@@ -9,7 +9,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"io/fs"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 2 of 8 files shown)


pkg/engine/source/filesystem.go
whitemerch, line 1: Before, embedded queries were read via `assets.GetEmbeddedQueryFile`, which internally normalizes paths using `filepath.ToSlash(path)` before calling `embeddedQueries.ReadFile`. This ensured compatibility with `embed.FS`, which expects /-separated paths.

With this PR, ReadQuery now calls `fs.ReadFile(filepath.Join(queryDir, QueryFileName))` directly, without any normalization. AFAIK with Windows, `filepath.Join` produces \-separated paths, which are not supported by `embed.FS.ReadFile`.

Although `ReadEmbeddedQuery` applies `filepath.ToSlash(queryDir)` beforehand, the subsequent `filepath.Join` in ReadQuery will reintroduce \ when appending `QueryFileName`, undoing the normalization.  ReadMetadata has the same problem where `filepath.Join(queryDir, MetadataFileName)` is akso used

AFAIK this will break embedded file reads on Windows, unless I missed something there

pkg/engine/source/filesystem.go
```diff
@@ -353,16 +354,21 @@ func validateMetadata(metadata map[string]interface{}) (exist bool, field string
...
 
+// ReadQuery reads embedded query files for a given path and returns a QueryMetadata struct with its content
+func ReadEmbeddedQuery(ctx context.Context, queryDir string) (model.QueryMetadata, error) {
+	return ReadQuery(ctx, assets.GetEmbeddedQueriesFs(), filepath.ToSlash(queryDir))
+}
+
 // ReadQuery reads query's files for a given path and returns a QueryMetadata struct with it's
 // content
-func ReadQuery(ctx context.Context, queryDir string) (model.QueryMetadata, error) {
+func ReadQuery(ctx context.Context, fs fs.ReadFileFS, queryDir string) (model.QueryMetadata, error) {
```
whitemerch: `fs.ErrNotExist` etc. become unreachable inside the function no? For ReadMetadata too?

pkg/engine/source/datadog.go
```diff
@@ -284,7 +284,6 @@ func convertRule(rule *Rule) model.QueryMetadata {
 		out.Metadata["cwe"] = ""
 	} else {
 		out.Metadata["cwe"] = *rule.Cwe
-		out.CWE = *rule.Cwe
```
whitemerch, line 287: What's the rationale there?

pkg/engine/source/filesystem.go
```diff
@@ -353,16 +354,21 @@ func validateMetadata(metadata map[string]interface{}) (exist bool, field string
 	return
 }
 
+// ReadQuery reads embedded query files for a given path and returns a QueryMetadata struct with its content
```
whitemerch: `ReadEmbeddedQuery` instead of `ReadQuery` in the comment

pkg/engine/source/filesystem.go
```diff
@@ -393,17 +399,22 @@ func ReadQuery(ctx context.Context, queryDir string) (model.QueryMetadata, error
 	}, nil
 }
 
+// ReadEmbeddedMetadata reads the query metadata file inside the embedded query directory
+func ReadEmbeddedMetadata(ctx context.Context, queryDir string) (map[string]any, error) {
```
whitemerch: nit: Should we change the output of `ReadMetadata` to be the same `map[string]any`?

pkg/engine/source/filesystem_test.go
whitemerch, line 1: Could we add a test that calls `ReadQuery / ReadMetadata` directly with a non-embedded FS? Because from what I saw, every test still goes through the embedded wrappers

jacobotb: Done.

pkg/engine/source/filesystem.go
jacobotb: I've changed `interface{}` to `any` in that file.

pkg/engine/source/datadog.go
jacobotb: That was something I found while I was doing my tests. The filesystem code doesn't populate CWE so my other codepath shouldn't do it either. I split that onto a separate commit (it's two lines.)

pkg/engine/source/filesystem.go
jacobotb: I'm not shadowing `fs` anymore.

pkg/engine/source/filesystem.go
jacobotb: Augh.

whitemerch (approved):

PR #97, opened 2026-04-21
Fix backend rule and configuration file parsing

jacobotb:
## Motivation

Fix bugs in the server-side rule and new configuration file parsing.

## Changes

For server-side rules, account for fields that are optional in the backend but mandatory in the client.

For the configuration file, account for files with an empty `iac` section or with an old schema version by allowing the client to fall back to the legacy configuration file.

## Author Checklist

## QA Instruction

## Blast Radius

## Additional Notes


MikaYuoadas (approved):
LGTM 👍

PR #111, opened 2026-04-24
Kustomize foundations: model, bounded filesystem, Helm renderer

whitemerch:
> [!TIP]

## Summary
- Refactor Helm chart rendering into a reusable preprocessor renderer with virtual file outputs and option handling.
- Add Kustomize foundation model/provenance types plus shared YAML line detection support for future resolver work.
- Add bounded filesystem safeguards for resolver file access and wire virtual files through scan preparation

Full design context and architecture details: https://github.com/DataDog/datadog-iac-scanner/issues/113

## Test plan
- `go test ./cmd/scanner ./pkg/featureflags ./pkg/kics ./pkg/resolver/helm ./pkg/resolver/sandbox ./pkg/scanner`
- `git diff --check origin/main`

## Non-goals
- No Kustomize preprocessor is registered or enabled in this PR.
- Existing Helm scan behavior is expected to remain compatible for the default scanner path.

### DIFF
--- a/pkg/featureflags/evaluator.go
+++ b/pkg/featureflags/evaluator.go
@@ -6,7 +6,8 @@ const (
 	IacDisableKicsRule               = "k9-iac-disable-kics-rule"
 	IacEnableKicsPlatform            = "k9-iac-enable-kics-platform"
 	IacEnableKicsHelmResolver        = "k9-iac-enable-kics-helm-resolver"
-	IaCEnableKicsParallelFileParsing = "k9-iac-enable-kics-parallel-file-parsing"
+	IacEnableKustomizeResolver       = "k9-iac-enable-kustomize-resolver"
+	IacEnableKicsParallelFileParsing = "k9-iac-enable-kics-parallel-file-parsing"
 )
 
 type FlagEvaluator interface {
--- a/pkg/resolver/helm/helm.go
+++ b/pkg/resolver/helm/helm.go
@@ -23,12 +23,8 @@ import (
 
 // credit: https://github.com/helm/helm
 
-var (
-	settings = cli.New()
-)
-
 func runInstall(ctx context.Context, args []string, client *action.Install,
-	valueOpts *values.Options) (*release.Release, []string, error) {
+	valueOpts *values.Options, opts *RenderOptions) (*release.Release, []string, error) {
 	contextLogger := logger.FromContext(ctx)
 	log.SetOutput(io.Discard)
 	defer log.SetOutput(os.Stderr)
@@ -46,16 +42,31 @@ func runInstall(ctx context.Context, args []string, client *action.Install,
 		return nil, []string{}, err
 	}
 	contextLogger.Debug().Msgf("Parsed chart name: '%s', chart path: '%s'", name, charts)
-	client.ReleaseName = name
+	if strings.TrimSpace(client.ReleaseName) == "" || client.GenerateName {
+		client.ReleaseName = name
+	}
 
-	cp, err := client.LocateChart(charts, settings)
+	helmSettings := cli.New()
+	repositoryConfig := strings.TrimSpace(opts.RepositoryConfig)
+	if repositoryConfig != "" {
+		helmSettings.RepositoryConfig = repositoryConfig
+	}
+	repositoryCache := strings.TrimSpace(opts.RepositoryCache)
+	if repositoryCache != "" {
+		helmSettings.RepositoryCache = repositoryCache
+	}
+	registryConfig := strings.TrimSpace(opts.RegistryConfig)
+	if registryConfig != "" {
+		helmSettings.RegistryConfig = registryConfig
+	}
+	cp, err := client.LocateChart(charts, helmSettings)
 	if err != nil {
 		contextLogger.Error().Msgf("failed to locate chart '%s': %s", charts, err)
 		return nil, []string{}, err
 	}
 	contextLogger.Debug().Msgf("Located chart at path: '%s'", cp)
 
-	p := getter.All(settings)
+	p := getter.All(helmSettings)
 	vals, err := valueOpts.MergeValues(p)
 	if err != nil {
 		contextLogger.Error().Msgf("failed to merge helm values: %s", err)
@@ -81,7 +92,6 @@ func runInstall(ctx context.Context, args []string, client *action.Install,
 	}
 	contextLogger.Debug().Msg("Chart installability check passed")
 
-	client.Namespace = "kics-namespace"
 	contextLogger.Debug().Msgf("Running helm chart with namespace: '%s', release name: '%s'", client.Namespace, client.ReleaseName)
 	helmRelease, err := client.Run(chartRequested, vals)
 	if err != nil {
@@ -105,8 +115,8 @@ func checkIfInstallable(ch *chart.Chart) error {
 	return errors.Errorf("%s charts are not installable (only 'application' type charts are supported)", ch.Metadata.Type)
 }
 
-// newClient will create a new instance on helm client used to render the chart
-func newClient(ctx context.Context) *action.Install {
+// newClient creates a Helm install client used to render the chart (dry-run).
+func newClient(ctx context.Context, opts *RenderOptions) *action.Install {
 	contextLogger := logger.FromContext(ctx)
 	contextLogger.Debug().Msg("Creating new helm client for chart rendering")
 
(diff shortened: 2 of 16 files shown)


pkg/resolver/sandbox/sandbox.go
```diff
@@ -0,0 +1,31 @@
+// Package sandbox: per-scan temp dirs for preprocessors.
+package sandbox
+
+import (
+	"os"
+)
+
+// Sandbox is a per-scan writable temp directory (0700 from MkdirTemp on Unix), removed on Close.
+type Sandbox struct {
+	RepoRoot   string
```
MikaYuoadas: **❓ Question:** The RepoRoot is never set or used anywhere in this PR. I get that this may be preparation for future usage, but shouldn’t it at least be set by the `New()` call to something?

pkg/resolver/sandbox/bounded_fs.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+// IsDir implements filesys.FileSystem.
+func (b *BoundedFS) IsDir(path string) bool {
+	d, f, err := b.cleanedPath(path, true)
+	if err != nil {
+		return false
+	}
+	_ = d
+	_ = f
```
MikaYuoadas: **⛔ Must fix:** Why are you doing these no-op assignment? If it’s to avoid go compiler nagging about "declared but unused" var, just discard them right when calling the cleanedPath.
```suggestion
	_, _, err := b.cleanedPath(path, true)
	if err != nil {
		return false
	}
```

pkg/resolver/sandbox/bounded_fs.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+// Exists implements filesys.FileSystem.
+func (b *BoundedFS) Exists(path string) bool {
+	d, f, err := b.cleanedPath(path, true)
+	if err != nil {
+		return false
+	}
+	_ = d
+	_ = f
```
MikaYuoadas: **⛔ Must fix:** Why are you doing these no-op assignment? If it’s to avoid go compiler nagging about "declared but unused" var, just discard them right when calling the cleanedPath.
```suggestion
	_, _, err := b.cleanedPath(path, true)
	if err != nil {
		return false
	}
```

pkg/resolver/sandbox/bounded_fs.go
```diff
@@ -0,0 +1,241 @@
...
+		return nil, err
+	}
+	var out []string
+	for _, m := range matches {
+		d, f, errC := b.cleanedPath(m, true)
+		if errC != nil {
+			continue
+		}
+		_ = d
+		_ = f
```
MikaYuoadas: **⛔ Must fix:** Why are you doing these no-op assignment? If it’s to avoid go compiler nagging about "declared but unused" var, just discard them right when calling the cleanedPath.
```suggestion
		_, _, errC := b.cleanedPath(m, true)
		if errC != nil {
			continue
		}
```

pkg/resolver/helm/render.go
```diff
@@ -0,0 +1,369 @@
...
+	}
+	if err := f.Close(); err != nil {
+		_ = os.Remove(f.Name())
+		return func() {}, errors.Wrap(err, "close merged inline helm values")
+	}
+	contextLogger.Debug().Msgf("wrote inline helm values to %s", f.Name())
+	// Single temp file replaces ValueFiles so inline merge is not applied twice.
+	valueOpts.ValueFiles = []string{f.Name()}
+	return func() {
+		_ = os.Remove(f.Name())
```
MikaYuoadas: **🤏 Nitpick:** You don’t need to explicitly discard return value in go.
```suggestion
		f.Close()
		os.Remove(f.Name())
		return func() {}, errors.Wrap(err, "write merged inline helm values")
	}
	if err := f.Close(); err != nil {
		os.Remove(f.Name())
		return func() {}, errors.Wrap(err, "close merged inline helm values")
	}
	contextLogger.Debug().Msgf("wrote inline helm values to %s", f.Name())
	// Single temp file replaces ValueFiles so inline merge is not applied twice.
	valueOpts.ValueFiles = []string{f.Name()}
	return func() {
		= os.Remove(f.Name())
```

pkg/featureflags/evaluator.go
```diff
@@ -6,6 +6,7 @@ const (
 	IacDisableKicsRule               = "k9-iac-disable-kics-rule"
 	IacEnableKicsPlatform            = "k9-iac-enable-kics-platform"
 	IacEnableKicsHelmResolver        = "k9-iac-enable-kics-helm-resolver"
+	IacEnableKustomizeResolver       = "k9-iac-enable-kustomize-resolver"
```
MikaYuoadas, line 9: **💬 Remark:** Not from your PR, but it’s an eyesore that we have both flags starting with `IaC` and `Iac` 😫

pkg/kics/resolver_diagnostics.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+	var digits [20]byte
+	i := len(digits)
+	for v > 0 {
+		i--
+		digits[i] = byte('0' + v%10)
+		v /= 10
+	}
+	return sign + string(digits[i:])
+}
```
MikaYuoadas: **⛔ Must fix:** Why are you reimplementing the standard library? Just use [strconv.Itoa()](https://pkg.go.dev/strconv#Itoa) and [strings.Join()](https://pkg.go.dev/strings#Join)

```suggestion
	return strings.Join([]string{scanID, d.QueryID, d.FilePath, d.Message, strconv.Itoa(d.Line)}, "\x00")
}
```

pkg/resolver/helm/render.go
```diff
@@ -0,0 +1,369 @@
...
+	}
+	return resolved, nil
+}
+
+func isHelmTestTemplate(path string) bool {
+	path = filepath.ToSlash(path)
+	return strings.Contains(path, "/tests/") || strings.HasPrefix(path, "tests/")
+}
+
+func splitManifestYAML(template *release.Release) (*[]splitManifest, error) {
```
MikaYuoadas: **💡 Suggestion:** slice in golang are already reference type (i.e. pointers to the actual structure). There’s hardly ever a point to have a pointer to a slice.
Here juste returning a []splitManifest avoid having the calling code dereferencing the pointer pointlessly.
The `for _, sp := range *splits {` above simply becomes `for _, sp := range splits {`

pkg/resolver/helm/helm.go
```diff
@@ -105,22 +112,42 @@ func checkIfInstallable(ch *chart.Chart) error {
...
 	contextLogger.Debug().Msg("Creating new helm client for chart rendering")
 
 	cfg := new(action.Configuration)
 	client := action.NewInstall(cfg)
 	client.DryRun = true
 	client.ReleaseName = "kics-helm"
+	if strings.TrimSpace(opts.ReleaseName) != "" {
+		client.ReleaseName = opts.ReleaseName
+	}
+	client.GenerateName = strings.TrimSpace(opts.ReleaseName) == ""
```
MikaYuoadas: **🤏 Nitpick:** You’re trimming twice in a row, you can just assign it once and use it twice.
Also we’re assigning the non-trimmed release name, shouldn’t we use the trimmed version instead?

```suggestion
	releaseName := strings.TrimSpace(opts.ReleaseName)
	if releaseName != "" {
		client.ReleaseName = releaseName
	}
	client.GenerateName = releaseName == ""
```

pkg/resolver/helm/helm.go
```diff
@@ -105,22 +112,42 @@ func checkIfInstallable(ch *chart.Chart) error {
...
-	client.IncludeCRDs = false
+	client.APIVersions = chartutil.VersionSet(opts.APIVersions)
+	client.IncludeCRDs = opts.IncludeCRDs
+	client.DisableHooks = opts.SkipHooks
+	client.NameTemplate = strings.TrimSpace(opts.NameTemplate)
+	client.RepoURL = strings.TrimSpace(opts.ChartRepo)
+	client.Version = strings.TrimSpace(opts.Version)
+	client.Devel = opts.Devel
+	if strings.TrimSpace(opts.KubeVersion) != "" {
+		if kv, err := chartutil.ParseKubeVersion(strings.TrimSpace(opts.KubeVersion)); err == nil {
```
MikaYuoadas: **🤏 Nitpick:** same
```suggestion
	kubeVersion := strings.TrimSpace(opts.KubeVersion)
	if kubeVersion != "" {
		if kv, err := chartutil.ParseKubeVersion(kubeVersion); err == nil {
```

pkg/resolver/helm/render.go
```diff
@@ -0,0 +1,369 @@
...
+	sources := make([]*chart.File, 0)
```
MikaYuoadas: **🤏 Nitpick:** You can simply do `var sources []*chart.File` to create a nil slice. `updateName` below will grow it as needed, you don’t need a non-nil empty slice for that.

MikaYuoadas (changes requested):
This PR is very big, its sheer size overflow the context of my limited brainpower (there’s only so much token available before my brains gives up 😆 ).
But I’ve done my best to read it all and left a few comments on minor issues I found.

Otherwise the main thing I’m concerned about is not in the code itself but on the decision to surface resolver failure as vulnerability that IIUC *will* be added to the final SARIF output and will create customer facing findings. Is that intended design? I find it quite strange, and if we want to do so it may warrant more product discussion. We are a security product, not a linter for syntax error. IMHO findings about syntax error is not something customers expect to see in IaC explorer. (but correct me if I understood it wrong).

whitemerch:
> This PR is very big, its sheer size overflow the context of my limited brainpower (there’s only so much token available before my brains gives up 😆 ). But I’ve done my best to read it all and left a few comments on minor issues I found.
> 
> Otherwise the main thing I’m concerned about is not in the code itself but on the decision to surface resolver failure as vulnerability that IIUC _will_ be added to the final SARIF output and will create customer facing findings. Is that intended design? I find it quite strange, and if we want to do so it may warrant more product discussion. We are a security product, not a linter for syntax error. IMHO findings about syntax error is not something customers expect to see in IaC explorer. (but correct me if I understood it wrong).

@MikaYuoadas Nop, you are totally right. I misunderstood this part of the code, I needed a way to report non-fatal resolver problems while still continuing the scan, but I will mimick the Helm behavior for now

PR #114, opened 2026-04-24
Read and apply server-side configurations

jacobotb:
## Motivation

Datadog customers can create and store tool configuration files on the Datadog backend. With this, the IaC scanner can get that server-side configuration and use it.

Legacy configurations also participate in this.

## Changes

Split out the Datadog client into its own package, with "get default rules" and "get configuration" methods.

Change configuration parsing so it will return the config file contents. This lets us send the configuration file to the backend.

Convert legacy configurations to YAML so they can be sent to the backend, and then apply any legacy-only configurations to the end result.

Add a "show-config" command that displays the effective configuration, with or without server-side changes.

## Author Checklist

## QA Instruction

## Blast Radius

## Additional Notes

### DIFF
--- a/cmd/scanner/scan.go
+++ b/cmd/scanner/scan.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/DataDog/datadog-iac-scanner/internal/console"
 	"github.com/DataDog/datadog-iac-scanner/pkg/config"
+	"github.com/DataDog/datadog-iac-scanner/pkg/datadog"
 	"github.com/DataDog/datadog-iac-scanner/pkg/featureflags"
 	"github.com/DataDog/datadog-iac-scanner/pkg/model"
 	"github.com/DataDog/datadog-iac-scanner/pkg/scan"
@@ -93,6 +94,7 @@ const (
 	dirPerms  = 0755
 )
 
+// nolint:gocyclo
 func runScan(ctx context.Context, c *cli.Command) error {
 	if c.Args().Len() > 0 {
 		return fmt.Errorf("unexpected arguments: %v", c.Args().Slice())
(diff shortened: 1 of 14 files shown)


cmd/scanner/scan.go
```diff
@@ -126,7 +128,12 @@ func runScan(ctx context.Context, c *cli.Command) error {
 		return fmt.Errorf("error retrieving repository commit information: %w", err)
 	}
 
-	cfg, err := config.ReadConfiguration(ctx, repoDir)
+	datadogClient, err := datadog.NewDatadogClient()
+	if err != nil {
+		return fmt.Errorf("error creating Datadog client")
```
MikaYuoadas: **💡 Suggestions:** We probably would like to have the original error message surfaced too here.
```suggestion
		return fmt.Errorf("error creating Datadog client: %v", err)
```

pkg/datadog/client.go
```diff
@@ -0,0 +1,272 @@
...
+
+type Client interface {
+	// GetDefaultRuleset returns the content of the default ruleset.
+	GetDefaultRuleset(ctx context.Context) (*Ruleset, error)
+	// GetRemoteConfig applies server-side changes to the local configuration.
+	GetRemoteConfig(ctx context.Context, repoUrl string, localConfig []byte) ([]byte, error)
+}
+
+// NewDatadogClient creates a DatadogSource with the given options.
+func NewDatadogClient(options ...DatadogClientOption) (Client, error) {
```
MikaYuoadas: **❓ Question:** Why add `error` to the return while we never actually return non-nil value in the body?
This force all calling code to handle an error that will never come. 🤔

pkg/config/reader.go
```diff
@@ -2,32 +2,126 @@ package config
...
+			return nil, nil, err
+		}
+		cfgBytes = newCfg
+	}
+
+	cfg, err := ParseConfig(cfgBytes)
+	if err != nil {
+		return nil, nil, fmt.Errorf("could not parse configuration file: %w", err)
+	}
+	return cfg, cfgBytes, err
```
MikaYuoadas: **🤏 Nitpick:** `err` is always nil here with the guarding `if` just before.
```suggestion
	return cfg, cfgBytes, nil
```

MikaYuoadas (dismissed):
LGTM, left a couple suggestions and nitpick

pkg/datadog/client.go
jacobotb: That was a leftover.

MikaYuoadas (approved):
LGTM 👍

PR #121, opened 2026-04-28
Flag only GitHub files cicd

ChouraquiBen:
## Motivation

The github resources scanning is too broad and also flags resources from other providers, such as Azure DevOps. This is because both file types share so much in common. This creates lots of False positives which have to be tackled:

**How to distinguish those files**
❎ Regex only at the analyzer cannot distinguish files that are Azure DevOps or GitHub Actions.
✅ The file path is can distinguish between all providers: github, azure, gitlab, circleci

**Why proceed directly within regos?**
- The only other way to proceed is by filtering directly in the analyzer to only get GitHub actions (shorter)
- However, this means that adding other providers (gitlab, azure...) will require new platforms, while CICD remains for GitHub. This is confusing.
- Thus changing the regos while giving them the path of the files scanned

## Changes

- Added the file path to the informations given to rego via the yaml parser
- Created a cicd function to check the file path and distinguish between github and the other providers
- Used this function throughout the cicd rego checks.

## Author Checklist

## QA Instruction

Check that no rego check has been forgotten
- Checked twice manually and once using Claude code

## Blast Radius

Overall, yaml findings although only cicd findings must be affected

## Additional Notes


assets/queries/cicd/github/anonymous_definition/test/positive_expected_result.json
```diff
@@ -2,11 +2,13 @@
   {
     "queryName": "Anonymous definition",
     "severity": "LOW",
-    "line": 1
+    "line": 1,
+    "fileName": "positive.github.yaml"
```
ChouraquiBen: Files are renamed to be flaggable

assets/libraries/cicd.rego
```diff
@@ -53,3 +53,15 @@ is_bare_inputs_dereference(node) {
 	object.type == "identifier"
 	lower(object.value) == "inputs"
 }
+
+check_provider(doc) := "github" {
+	contains(doc._path, ".github")
```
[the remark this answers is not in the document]
ChouraquiBen: You must store the workflows in the .github.workflow directory, cf [documentation](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#about-yaml-syntax-for-workflows)

assets/queries/cicd/github/unsecured_commands/query.rego
```diff
@@ -2,64 +2,73 @@ package Cx
 
 import data.generic.cicd as cicd_lib
 import data.generic.common as common_lib
+import data.generic.cicd as cicd_lib
```
whitemerch: All of the regos already import `import data.generic.cicd as cicd_lib`, so this is duplicated

assets/libraries/cicd.rego
```diff
@@ -53,3 +53,15 @@ is_bare_inputs_dereference(node) {
...
+} else := "gitlab" {
+	contains(doc._path, ".gitlab-ci")
+} else := "circleci" {
+	contains(doc._path, ".circleci")
+} else := "azure" {
+	contains(doc._path, "azure-pipelines")
+} else := "other" {
+	true
+}
```
whitemerch, line 67: This is too loose. A few problems with the substring approach:

- contains(_, ".github") matches files like my.github.config.yaml, azure-pipelines.github.yaml, ci.github-mirror.yml, etc. A file path component check (e.g. contains(doc._path, "/.github/") or splitting on /) would be much safer

- The first match wins, but .gitlab-ci files placed in a repo that also has .github/ somewhere in the cwd-relative path (which can happen when scanning a parent directory) would be classified as github. Anchoring on a path segment rather than substring would fix this

- contains(_, "azure-pipelines") only catches files literally named azure-pipelines*.yml. Azure DevOps doesn't mandate that name, so this branch will silently mis-classify lots of Azure pipelines as other

whitemerch, line 67: An action.yml at a repo root (extremely common for action-only repos like actions/checkout, actions/setup-node, etc.) is classified as cicd by the analyzer and then dropped by the rego gate, which defeats the purpose of my previous PR

assets/libraries/cicd.rego
whitemerch, line 1: We should have negative tests for the regos to test against this util

assets/libraries/cicd.rego
whitemerch, line 67: Should we just simply drop everything that is not github, and once we add future support, we choose the correct naming related to what we want to name the folders for

whitemerch (changes requested):

assets/libraries/cicd.rego
```diff
@@ -53,3 +53,15 @@ is_bare_inputs_dereference(node) {
...
+	contains(doc._path, "/.github/")
+} else := "github" {
+	regex.match("/action.ya?ml", doc._path)
+} else := "github" {
+	contains(doc._path, "\\.github\\")
+} else := "github" {
+	regex.match("\\action.ya?ml", doc._path)
+} else := "other" {
+	true
+}
```
whitemerch, line 67: Two problems with regex.match("\\action.ya?ml", doc._path):

- In Rego, `\\a` is the 2‑char string `\a`.
- In Go's RE2 (which our OPA uses), \a is the alert/bell character (U+0007), not a literal backslash + a

So this branch matches paths containing `<bell>ction<any>y(a?)ml`, i.e. essentially nothing. To match a literal `\action.ya?ml` on a Windows path, the Rego string needs to be `\\\\action\\.ya?ml` (regex `\\action\.ya?ml`)

whitemerch, line 67: `regex.match("/action.ya?ml", _path) `matches `/actionXyml` and `/actionXyaml` too. regex.match("/action\\.ya?ml", _path) is the correct intent and costs nothing

assets/queries/cicd/github/anonymous_definition/test/.github/negative.yaml
whitemerch, line 1: Could we still put them inside /test/.github/?

whitemerch (changes requested):

whitemerch:
@codex review

whitemerch (approved):

PR #138, opened 2026-05-11
[ID Migration #9] Add testing

ChouraquiBen:
## Motivation

The ids of the rules are being moved: to be easier to use on the configuration files, the IDs are being changed from base64 UUIDs to human-readable IDs. This work is divided in ten PRs for the logic change, the migrations for the different platforms, the tests and finally, the documentation.

This PR adds the missing testing for the queries: checking that no `id`/no `legacyID` are duplicated, checking the format of the new `id`

## Changes

- Check `id` and `legacyId` uniqueness
- Test that the `id` matches the new format
- Check that the `cloudProvider` field is lower case

## Author Checklist

## QA Instruction

- Check that the logic of the tests is correct
- Are there any missing tests

## Blast Radius

This is testing and should not affect the logic of the code

## Additional Notes


test/queries_content_test.go
```diff
@@ -42,44 +43,45 @@ var (
...
 	availablePlatforms = initPlatforms()
 	platformKeys       = MapToStringSlice(availablePlatforms)
 
 	CategoriesKeys = MapToStringSlice(constants.AvailableCategories)
 
-	searchValueProperty = "searchValue"
-
-	requiredQueryMetadataProperties = map[string]func(tb testing.TB, value interface{}, metadataPath string){
-		"id": func(tb testing.TB, value interface{}, metadataPath string) {
+	requiredQueryMetadataProperties = map[string]func(tb testing.TB, value any, metadata map[string]any,
```
whitemerch, line 62: PR description says it adds format checks, but current metadata test validates id == ToID(platform, cloudProvider, folderSlug) only; it does not enforce a strict format policy (e.g. no dots). This allows non-slug provider tokens like 2.0/3.0 to pass, which is based off my comment on the previous PR

pkg/utils/query.go
whitemerch, line 1: Small edge case here: `ChooseQueryID` should probably treat an empty `legacyQueryID` the same way it treats "Undefined".

Right now, if `legacyQueryID == ""`, the function returns `""` instead of falling back to `queryID`. That can leak an empty rule ID into fingerprint/similarity logic and make unrelated findings look identical

whitemerch (dismissed):
I think overall we should also enforce legacyId, since it is used for the hash, wdyt? or maybe adapt the generation of the hash to fallback on id if legacyId is not present

pkg/utils/query.go
ChouraquiBen: `legacyQueryID` is not allowed to be duplicated between rules, so there should be no issue when it comes to the hash.
However, I agree that allowing it to be empty is a bit weird, so I would rather enforce in the tests that it is either Nil or not empty, what do you think?

whitemerch (dismissed):

whitemerch (dismissed):

whitemerch (approved):

PR #141, opened 2026-05-11
Fix: Apply remote config even when no local config file is present

MikaYuoadas:
## What

Refactors `pkg/config` to always apply server-side options (e.g. `WithDatadog`) regardless of whether a local config file exists, and aligns the config file reading logic with the RFC requirements for schema version handling and unknown root properties.

## Why

`WithDatadog`, which fetches the merged org/repo-level config from the Datadog settings API, was only invoked when a local `code-security.datadog.yaml` or `dd-iac-scan.config` was found. Repositories with no local config file never had server-side customizations applied, making org-level and repo-level IaC config settings in the Datadog UI completely ineffective.

A review of the RFC also identified two additional gaps in the local config reading logic:

- A config file with a future minor-version product section (no `iac:` key) was incorrectly passed to the API instead of falling through to the legacy config, causing a server-side error.
- Config files with unknown root properties were silently accepted instead of being rejected as the RFC requires.

## Changes

Four atomic commits:

1. `ReadConfiguration` simplified to: **find local bytes → apply options → parse**. Options are now called even when no local file exists, so the API is always consulted for server-side config.

2. `findLocalConfig` reworked to use a new `hasIacSectionForSupportedVersion` helper that uses lenient YAML parsing to check for the `iac:` key without strict field validation. Schema version boundary constants (`minIacVersion`, `minUnsupportedVersion`) extracted and shared with `ParseConfig`.

3. `hasIacSectionForSupportedVersion` updated to use `KnownFields(true)` on a struct listing all valid root keys (`sast`, `iac`, `sca`, `secrets`, `iast`), so unknown root properties produce an error per the RFC.

4. `show-config`: fix `--path` flag default causing the provided path to be appended to `"."` instead of replacing it (urfave/cli v3 `StringSliceFlag` behavior).

## Behavior matrix

| Scenario | Behavior | RFC requirement | Test |
|---|---|---|---|
| No local file, no remote | empty config | default empty | `TestNoConfig` |
| No local file, remote success | config from remote | use server-side | `TestNoConfigWithDatadog` |
| Unsupported version (>= v2.0) | error, legacy ignored | reject major >= 2 | `TestUnsupportedSchemaVersion` |
| Unknown root key | error, legacy ignored | unknown root props fail | `TestUnknownRootKey` |
| No `iac:` section (v1.2+) | fall through to legacy or nil | use legacy or server-side | `TestIgnoredConfigFile`, `TestConfigFilePrecedenceWithIgnoredConfigFile` |
| Old schema (v1.1) | fall through | accept all v1.x minor versions | `TestIgnoredConfigFile/old_schema_version` |
| Valid config, no remote | config from new file | new file with `iac:` | `TestReadConfig` |
| Valid config + remote success | merged config from remote | merge local with server-side | `TestConfigFileFromDatadog/regular_config_with_iac_section` |
| No `iac:` section + remote success | nil sent, config from remote | use server-side only | `TestConfigFileFromDatadog/regular_config_without_iac_section` |
| Legacy only, no remote | config from legacy | legacy: convert, use result | `TestReadLegacyOnly` |
| Legacy + remote success | merged config from remote | legacy: convert, merge | `TestConfigFileFromDatadog/legacy_config` |

## End-to-end test results

Tested locally using the PR binary with live Datadog API credentials across 15 scenarios covering local config, remote org/repo config, and combinations.

| Scenario | Config source | Expected | Result |
|---|---|---|---|
| S01 baseline | none | 16 findings | 16 ✅ |
| S02 ignore-rules | local | 12 (Rule A excluded) | 12 ✅ |
| S03 use-rules | local | 4 (only Rule A) | 4 ✅ |
| S04 ignore-paths | local | 12 (no `configs/ignored/`) | 12 ✅ |
| S05 only-paths | local | 4 (only `configs/included/`) | 4 ✅ |
| S06a ignore LOW severity | local | 12 (Rule A excluded) | 12 ✅ |
| S06b only HIGH+CRIT | local | 8 (Rules C+D only) | 8 ✅ |
| S07a ignore Best Practices | local | 12 (Rule A excluded) | 12 ✅ |
| S07b only Encryption | local | 4 (Rule C only) | 4 ✅ |
| S08 repo-level UI config | remote (repo) | 12 (Rule A excluded) | 12 ✅ |
| S09 org-level UI config | remote (org) | 12 (no `configs/ignored/`) | 12 ✅ |
| S10 org+repo UI config merged | remote (org+repo) | 9 (Rule A excluded + no `configs/ignored/`) | 9 ✅ |
| S11 local overrides org+repo | local + remote | 3 (only Rule B, no `configs/ignored/`) | 3 ✅ |
| S13 new config wins over legacy | local (both files) | 4 (only Rule B) | 4 ✅ |
| S14 inline suppression | local (inline comments) | 14 (Rule A suppressed in 2 files) | 14 ✅ |

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -52,11 +52,9 @@ func ParseConfig(cfgBytes []byte) (*IacConfig, error) {
 	if err != nil {
 		return nil, err
 	}
-	// Versions older than v1.2 can't have an IaC configuration
 	if version.compare(minIacVersion) < 0 {
 		return nil, nil
 	}
-	// Versions v2.0 and higher are not supported
 	if version.compare(minUnsupportedVersion) >= 0 {
 		return nil, fmt.Errorf("configuration schema version %s is not supported", version)
 	}
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -53,11 +53,11 @@ func ParseConfig(cfgBytes []byte) (*IacConfig, error) {
 		return nil, err
 	}
 	// Versions older than v1.2 can't have an IaC configuration
-	if version.compare(schemaVersion{1, 2}) < 0 {
+	if version.compare(minIacVersion) < 0 {
 		return nil, nil
 	}
 	// Versions v2.0 and higher are not supported
-	if version.compare(schemaVersion{2, 0}) >= 0 {
+	if version.compare(minUnsupportedVersion) >= 0 {
 		return nil, fmt.Errorf("configuration schema version %s is not supported", version)
 	}
 
@@ -142,6 +142,13 @@ func parseSchemaVersion(schema string) (version schemaVersion, err error) {
 type schemaVersion struct {
 	major, minor int
 }
+
+var (
+	// minIacVersion is the minimum schema version supporting the iac: configuration section.
+	minIacVersion = schemaVersion{1, 2}
+	// minUnsupportedVersion is the minimum schema version not supported by this scanner.
+	minUnsupportedVersion = schemaVersion{2, 0}
+)
 
 // compare returns a positive number if this schema version is higher than the other schema version,
 // a negative number if it's smaller, or zero if they are equal
--- a/pkg/config/reader.go
+++ b/pkg/config/reader.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"os"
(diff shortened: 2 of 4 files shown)


pkg/config/reader.go
```diff
@@ -15,54 +15,72 @@ const (
...
+func findLocalConfig(ctx context.Context, rootPath string) (cfgBytes []byte, legacyExcludeResults []string, err error) {
+	if path, found := fileExists(rootPath, ConfigFileNameBase, ".yaml", ".yml"); found {
+		b, err := os.ReadFile(path) // nolint:gosec
+		if err != nil {
+			return nil, nil, newInvalidLocalConfigError(fmt.Errorf("could not read configuration file %s: %w", path, err))
+		}
+		cfg, err := ParseConfig(b)
+		if err != nil {
+			return nil, nil, newInvalidLocalConfigError(fmt.Errorf("could not parse configuration file: %w", err))
+		}
```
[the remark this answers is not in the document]
whitemerch: I agree with this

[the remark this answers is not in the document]
MikaYuoadas: Fixed, I redid a lot of the config’s logic to adhere more closely to the RFC’s

pkg/config/reader.go
```diff
@@ -1,123 +1,151 @@
...
 const (
 	ConfigFileNameBase   = "code-security.datadog"
 	LegacyConfigFileName = "dd-iac-scan.config"
 )
 
-// ReadConfiguration reads and parses the configuration file in the local repository.
-// It returns the parsed configuration and the YAML file (or a YAML conversion, in the case of legacy configurations).
+// ReadConfiguration reads the local config file (if any), applies all options
+// (e.g. WithDatadog), and returns the parsed result.
+//
```
whitemerch: redundant //

pkg/config/config.go
```diff
@@ -53,11 +53,11 @@ func ParseConfig(cfgBytes []byte) (*IacConfig, error) {
 		return nil, err
 	}
 	// Versions older than v1.2 can't have an IaC configuration
-	if version.compare(schemaVersion{1, 2}) < 0 {
+	if version.compare(minIacVersion) < 0 {
 		return nil, nil
 	}
 	// Versions v2.0 and higher are not supported
```
whitemerch: should we remove these comments then? yours are better and it seems a bit easy to fall off async

whitemerch (dismissed):
neat

pkg/config/reader.go
```diff
@@ -1,123 +1,151 @@
...
 	}
+	return &IacConfig{}, cfgBytes, nil
+}
 
-	// Try to convert the legacy configuration to the new format so we can apply the options
-	cfgBytes, err := UnparseConfig(converted)
+// minimalCfgFile is used by hasIacSectionForSupportedVersion to detect whether the config
+// file has an iac: section. All root-level product keys are listed so that KnownFields
+// rejects unrecognized root properties per the RFC requirement.
+type minimalCfgFile struct {
```
ChouraquiBen: I think it is maximal, because you can't have more/other keys than those

ChouraquiBen (dismissed):
LGTM :shipit:

pkg/config/reader.go
MikaYuoadas: Yes, I’ll remove that

pkg/config/config.go
MikaYuoadas: Oh yes, I removed some but seems I missed a few 😬

pkg/config/reader.go
MikaYuoadas: Changed it to `knownRootsCfgFile`

ChouraquiBen (approved):
LGTM

whitemerch (approved):

PR #144, opened 2026-05-13
Apply Datadog remote config when scanning via pkg/scan

whitemerch:
## Motivation

PR [#141](https://github.com/DataDog/datadog-iac-scanner/pull/141) fixes `config.ReadConfiguration` so server-side options (for example `WithDatadog`) run even when there is no local `code-security.datadog.yaml`. That is necessary but not sufficient for every caller: the **library** path used by downstream integrations still called `ReadConfiguration` **without** `WithDatadog`, so the Datadog settings API was never consulted and org or repository settings in the product had no effect.

This change wires the same remote-configuration behavior into the shared entrypoint (`initializeConfig` → `GetDefaultParameters`) by resolving the repository URL from the enclosing git repository’s `origin` remote and passing `config.WithDatadog` when that URL is available. Remote fetch failures are logged and the scan continues with the local configuration only, so library callers are not broken by transient API issues; the CLI path is unchanged and still surfaces hard errors when the API fails while credentials are set (existing behavior).

## Changes

- Resolve `origin` URL from the git repo containing the scan root (`pkg/scan/repo_url.go`).
- Pass `WithDatadog` from `initializeConfig` when a remote URL exists; wrap the option so API errors fall back to local config for library scans (`pkg/scan/pre_scan.go`).
- Add unit tests for the wrapper, the no-git case, and the happy path for repo URL discovery and `configurationOptions` (`pkg/scan/pre_scan_test.go`).

## Author Checklist

## QA Instruction

From the repository root:

```bash
go test ./pkg/scan/... ./pkg/config/... -count=1
```

Optional: in a temp git repo with `origin` set and `DD_API_KEY` / `DD_APP_KEY` / `DD_SITE` configured, call `scan.GetDefaultParameters` (or run a scan through the library) and confirm the merged configuration reflects Datadog UI settings when no local IaC config file exists.

## Blast Radius

Affects **library** consumers of `github.com/DataDog/datadog-iac-scanner/pkg/scan` (for example embedded scanners): they will now attempt the same remote config merge as the `scanner` CLI when a git `origin` is present. The `cmd/scanner` binary behavior is unchanged. No change to SARIF schema or query assets.

## Additional Notes

### DIFF
--- a/pkg/scan/pre_scan.go
+++ b/pkg/scan/pre_scan.go
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	"github.com/DataDog/datadog-iac-scanner/pkg/config"
+	"github.com/DataDog/datadog-iac-scanner/pkg/datadog"
 	"github.com/rs/zerolog/log"
 )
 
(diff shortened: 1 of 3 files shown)


whitemerch:
Closing in favor of #145: the head branch was renamed to `chakib.hamie/apply_remote_config_in_library_entrypoint` (GitHub does not allow changing a PR's head branch after creation).

PR #145, opened 2026-05-13
Add DD_JWT_TOKEN and DD_HOSTNAME support to the Datadog API client

whitemerch:
## Motivation

[#141](https://github.com/DataDog/datadog-iac-scanner/pull/141) fixed the local-file gate so `WithDatadog` always runs. The scanner can still miss remote configuration in two scenarios that the current client cannot express:

1. **JWT-based authentication.** `pkg/datadog/client.go` only honored `DD_API_KEY` / `DD_APP_KEY`; with neither set, `GetRemoteConfig` silently echoed the local configuration with no HTTP request, no log, and no error. Callers that authenticate via a `dd-auth-jwt` header had no way to opt in.
2. **Endpoints that are not `api.<DD_SITE>`.** The host was always built from `DD_SITE`, so there was no way to point the client at a different hostname (for example a custom mock or a non-public host) without also changing the rest of the API contract.

[datadog-static-analyzer](https://github.com/DataDog/datadog-static-analyzer) already exposes both knobs (`DD_JWT_TOKEN` for the `dd-auth-jwt` header, and `DD_HOSTNAME` taking precedence over `DD_SITE`). This PR brings the IaC scanner client to the same contract.

## Changes

- Add `WithJwtToken` / `WithJwtTokenFromEnv` reading `DD_JWT_TOKEN` (or `DATADOG_JWT_TOKEN`); when set, send `dd-auth-jwt: <token>` on outgoing requests.
- Extend the silent "no credentials" bailout in `GetRemoteConfig` to also check the JWT, so the request is only skipped when no API key, no application key, **and** no JWT are configured.
- Add `WithHostnameFromEnv` reading `DD_HOSTNAME` (or `DATADOG_HOSTNAME`), taking precedence over `DD_SITE`. The value may include an `http://` or `https://` scheme; when omitted, `https://` is used.
- Treat an explicitly empty `DD_*` env var as absent so the `DATADOG_*` form is honored when the `DD_` alias is exported empty.
- Tests in `pkg/datadog/client_test.go` cover each path: JWT auth, no-credentials no-op, empty-`DD_*` fallback, `DD_HOSTNAME`-over-`DD_SITE` precedence, and the `http://` scheme passthrough.

## Author Checklist

## QA Instruction

From the repository root:

```bash
go test ./pkg/datadog/... -count=1 -v
```

The new tests assert the exact request shape (path, method, headers) on every new path; a regression on credential precedence or hostname handling will fail one of them.

## Blast Radius

Strictly additive. Existing usage with `DD_API_KEY` + `DD_APP_KEY` and `DD_SITE` is unchanged: when `DD_HOSTNAME` is unset and credentials are present, the client still issues the same request to `api.<DD_SITE>` it did before. Callers with no credentials at all still hit the same silent no-op. No change to SARIF output, query assets, rule evaluation, or scan exit codes.

## Additional Notes

### DIFF
--- a/pkg/datadog/client.go
+++ b/pkg/datadog/client.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"net/http"
 	"os"
+	"strings"
 
 	"github.com/DataDog/jsonapi"
 )
@@ -25,6 +26,9 @@ func NewDatadogClient(options ...DatadogClientOption) Client {
 	}
 	for _, option := range options {
 		option(out)
+	}
+	if out.hostname == "" {
+		WithHostnameFromEnv()(out)
 	}
 	if out.hostname == "" {
 		WithSiteFromEnv()(out)
@@ -109,11 +113,19 @@ func WithHttpClient(client *http.Client) DatadogClientOption {
 }
 
 // WithHostname lets you specify the hostname to use for Datadog API requests.
-// Used in the implementation and unit tests.
+// The value may include an http:// or https:// scheme; when omitted, https:// is used.
 func WithHostname(hostname string) DatadogClientOption {
 	return func(ds *datadogClient) {
 		ds.hostname = hostname
 	}
+}
+
+// WithHostnameFromEnv uses the hostname specified in the DD_HOSTNAME or DATADOG_HOSTNAME environment variable.
+// When set, it takes precedence over DD_SITE so that internal Datadog runners can point the scanner at a
+// private static-analysis-api address (e.g. an internal fabric.dog host). Matches the behavior of
+// datadog-static-analyzer's get_datadog_basename.
+func WithHostnameFromEnv() DatadogClientOption {
+	return WithHostname(getDdEnvvar("HOSTNAME"))
 }
 
 type DatadogClientOption func(source *datadogClient)
@@ -194,7 +206,7 @@ func (s *datadogClient) GetRemoteConfig(ctx context.Context, repoUrl string, loc
 
 // sendRequest sends a Datadog API request
 func (s *datadogClient) sendRequest(ctx context.Context, method, path string, requestBody io.Reader) (*http.Response, error) {
-	url := fmt.Sprintf("https://%s/api/v2/static-analysis/%s", s.hostname, path)
+	url := fmt.Sprintf("%s/api/v2/static-analysis/%s", s.baseURL(), path)
 	req, err := http.NewRequestWithContext(ctx, method, url, requestBody)
 	if err != nil {
 		return nil, fmt.Errorf("error building %s %s request: %w", method, url, err)
@@ -210,6 +222,15 @@ func (s *datadogClient) sendRequest(ctx context.Context, method, path string, re
 		req.Header.Add("dd-auth-jwt", s.jwtToken)
 	}
 	return s.httpClient.Do(req)
+}
+
+// baseURL returns the API base URL, preserving an explicit http:// or https:// scheme
+// when present in the hostname and defaulting to https:// otherwise.
+func (s *datadogClient) baseURL() string {
+	if strings.HasPrefix(s.hostname, "http://") || strings.HasPrefix(s.hostname, "https://") {
+		return s.hostname
+	}
+	return "https://" + s.hostname
 }
 
 // getDdEnvvar returns the value of the given Datadog environment variable.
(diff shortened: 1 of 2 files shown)


pkg/scan/pre_scan.go
```diff
@@ -14,7 +15,36 @@ func initializeConfig(ctx context.Context, rootPath string) (*config.IacConfig,
...
+	inner := config.WithDatadog(client, repoURL)
+	return func(ctx context.Context, cfgBytes []byte) ([]byte, error) {
+		out, err := inner(ctx, cfgBytes)
+		if err != nil {
+			log.Ctx(ctx).Warn().Err(err).Msg("could not fetch remote configuration from Datadog; using local configuration only")
+			return cfgBytes, nil
+		}
+		return out, nil
+	}
+}
```
MikaYuoadas: ❓ **Question:** I don’t see the value in wrapping `WithDatadog()`? We could just have modified the original function to fallback in case of API issues.

pkg/scan/pre_scan.go
```diff
@@ -14,7 +15,36 @@ func initializeConfig(ctx context.Context, rootPath string) (*config.IacConfig,
...
+// performed via the library entrypoint. Remote configuration is fetched when a git
+// `origin` remote is resolvable for rootPath; otherwise only the local
+// configuration is used.
+func configurationOptions(rootPath string) []config.ReadConfigurationOption {
+	repoURL := lookupRepositoryURL(rootPath)
+	if repoURL == "" {
+		return nil
+	}
+	return []config.ReadConfigurationOption{
+		bestEffortDatadog(datadog.NewDatadogClient(), repoURL),
```
MikaYuoadas: ⛔ **Must fix:** Way overkill to try to lookup the repository and do path traversal to go upward to the parent dir until a git repo is found. We know here that rootPath is the repo root.
```suggestion
		bestEffortDatadog(datadog.NewDatadogClient(), rootpath),
```

pkg/scan/repo_url.go
```diff
@@ -0,0 +1,57 @@
...
+				return nil, openErr
+			}
+		}
+		parent := filepath.Dir(dir)
+		if parent == dir || parent == "" || parent == "." {
+			return nil, git.ErrRepositoryNotExists
+		}
+		dir = parent
+	}
+}
```
MikaYuoadas: ⛔ **Must fix:** As stated in the other comment, all this can be removed.

pkg/scan/pre_scan.go
```diff
@@ -14,7 +15,36 @@ func initializeConfig(ctx context.Context, rootPath string) (*config.IacConfig,
 
 	baseLogger.Debug().Msg("console.initializeConfig()")
 
-	configParams, _, err := config.ReadConfiguration(ctx, rootPath)
+	configParams, _, err := config.ReadConfiguration(ctx, rootPath, configurationOptions(rootPath)...)
```
MikaYuoadas: ⛔ **Must fix:** InitializeConfig is only used by test code,  that’s why no configuration option is passed here.

MikaYuoadas (dismissed):
I don’t think this PR fix the issue, as far as I can tell it’s only touching code used for the tests. (`initializeConfig()` is only used by `GetDefaultParameters()` which itself is only called by tests code.

whitemerch:
> I don’t think this PR fix the issue, as far as I can tell it’s only touching code used for the tests. (`initializeConfig()` is only used by `GetDefaultParameters()` which itself is only called by tests code.

@MikaYuoadas Indeed, I didn't pull it out of draft as I didn't review it yet, I will check what is actually produced

pkg/datadog/client.go
```diff
@@ -82,6 +85,21 @@ func WithAppKeyFromEnv() DatadogClientOption {
...
+	return func(ds *datadogClient) {
+		ds.jwtToken = jwtToken
+	}
+}
+
+// WithJwtTokenFromEnv uses the JWT specified in the DD_JWT_TOKEN or DATADOG_JWT_TOKEN environment variable.
+// If neither variable exists, an empty JWT will be used. The JWT is sent as the `dd-auth-jwt`
+// header and is the auth mechanism used by Datadog code-workload-runner environments.
+func WithJwtTokenFromEnv() DatadogClientOption {
+	return WithJwtToken(getDdEnvvar("JWT_TOKEN"))
```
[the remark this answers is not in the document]
whitemerch: pre-existing bug, but yes

ChouraquiBen (approved):
The PR looks good to me:
✅ I have tested locally and it seems to be working with and without remote config
✅ Checked the tests made on the repository view to check CWP behavior

Found out that the `ignore-paths` option does not work but unrelated to this PR

MikaYuoadas (approved):

PR #148, opened 2026-05-19
Delete four redundant rule-corpus tests

whitemerch:
## Motivation

A handful of tests in `pkg/engine/source` and `pkg/scan` only assert that one specific rule's ~200-line metadata struct matches verbatim, or deep-equal a hardcoded list of paths under `assets/queries/`. They each have a stronger counterpart already in the same package (structural assertions or pure unit tests on the filter functions), and the cost of keeping them is that any non-trivial rule edit forces a verbatim re-typing of metadata in the test, while teaching the reader nothing new.

This PR removes them so the test suite stops paying that tax. It is also a small prerequisite for an upcoming move of rule definitions out of this repository: the scanner-side test surface should reflect scanner behavior, not the contents of a specific rule.

## Changes

Removed:
- `TestFilesystemSource_GetQueries`, `TestFilesystemSource_GetQueriesWithExclude`, `TestFilesystemSource_GetQueriesWithInclude` (`pkg/engine/source/filesystem_test.go`) — each deep-equals a hand-rolled ~200-line metadata struct for `terraform/aws/alb_deletion_protection_disabled`. Filter behavior is already covered by `TestCheckQueryInclude/ExcludeWithLegacyId` (pure unit) and `TestGetQueriesWithLegacyIdFiltering` (integration), and the structural variant `TestFilesystemSource_GetQueriesWithIncludeFilter` remains as the structural counterpart.
- `Test_AnalyzePaths` (`pkg/scan/utils_test.go`): deep-equals four hardcoded `assets/queries/terraform/alicloud/...` paths; `analyzePaths` is exercised on every full scan run.
- The now-unused imports in both files.

Net: -343 lines, +0 lines.

## Author Checklist

- [x] I have added or updated relevant unit tests where necessary. If no tests are added, I've explained why. (Pure deletion; each removed test has a stronger counterpart in the same package — see Motivation.)
- [x] I have tested my changes on staging (if applicable). (N/A — test-only change.)

## QA Instruction

```
go test ./pkg/engine/source ./pkg/scan
```

Both packages should be green. The remaining tests exercise the same code paths the deleted tests did (filter functions and `analyzePaths`); if any of them regress, the deleted tests would not have caught anything the survivors don't already catch.

## Blast Radius

Test-only change in `datadog-iac-scanner`. No production code path is touched, no public API, no rule semantics, no CLI behavior.

## Additional Notes

First of a small series cleaning up tests that are coupled to specific rule slugs/metadata in preparation for moving rule definitions to a dedicated repository. Each follow-up will be a similarly small, standalone PR.


pkg/scan/scan_test.go
```diff
@@ -1,87 +1,13 @@
 package scan
```
ChouraquiBen, line 1: > Test_ExecuteScan (pkg/scan/scan_test.go): one subtest, test exec scan, strictly weaker than Test_E2EExclusions/no_exclusions.

True but an E2E test should not replace a unit test: the issue that makes the E2E test fail may be located either in the `ExecuteScan` function or somewhere else. Keeping the Unit test makes bug detection faster and easier.

pkg/engine/inspector_test.go
```diff
@@ -128,140 +125,6 @@ func TestInspector_GetCoverageReport(t *testing.T) {
 }
 
 // TestNewInspector tests the functions [NewInspector()] and all the methods called by them
-func TestNewInspector(t *testing.T) { //nolint
```
ChouraquiBen, line 131: Creating a new inspector is at the basis of the logic. Even if the testing was not great, it ensured that the new inspector was indeed created and that it was within expected behaviour. I would not remove this test without any replacement

ChouraquiBen (commented):
Some of the tests are described as duplicated and they might be removed without harm. The other ones should be either kept or replaced by better alternatives

ChouraquiBen (approved):
LGTM! Thanks!

PR #149, opened 2026-05-19
Convert four inspector tests, executeScan smoke, cyclonedx, and jsonencode helper to local testdata

whitemerch:
## Summary

Phase 2 of the scanner test triage plan (follow-up to #148). Several tests only needed `assets/queries/` as **setup**, not as the thing under test. That made them brittle (hardcoded SHAs, rule-specific paths, corpus size) and blocked removing the embedded rules later.

This PR decouples those tests from the rule corpus where we can do it in test-only code. No production changes.

### `pkg/engine`

- **`newTestInspector(t, opts)`** runs real `NewInspector` against a stub `QueriesSource` (small `[]model.QueryMetadata`, optional `querySource` override). **Replaces** walking `assets/queries/` via `NewFilesystemSource` just to build an inspector. **Unblocks** unit tests for wiring and inspector methods without loading ~1400 rules.
- **`TestNewInspector`**, **`TestEngine_LenQueriesByPlat`**, **`TestEngine_GetFailedQueries`**, **`TestInspector_DecodeQueryResults`** use that helper with synthetic data. **Replaces** corpus/fixture setup the assertions never used (including the `alb_*` “≥ 2” subtest and `len(queries) > 0` smoke). **Unblocks** these tests staying green when the embed is removed. Removed **`newInspectorInstance`** / **`mockSource`**.

### `pkg/scan`

- **`Test_ExecuteScan`** now scans `pkg/scan/testdata/sample.tf` and only asserts: no error, non-nil result, every severity is valid. No pinned finding count (that would change whenever the rule corpus changes). **Replaces** a test that asserted an exact result count tied to the live rule set. **Unblocks** a stable smoke signal for `NewClient` → `executeScan` without re-bumping counts on every rule add/remove.
- **Why it still touches the embedded corpus:** `executeScan` → `Client.createQuerySource()` → `FilesystemSource`, and `FilesystemSource.GetQueries` reads from `assets.GetEmbeddedQueryDirs()` (the embed), not from `ScanParams.QueriesPath`. Unlike the engine tests above, we cannot pass a stub source without a small production hook (e.g. injectable `createQuerySource` on `Client`). That hook is intentionally out of scope here; this PR documents the limit in the test comment and keeps a useful smoke signal for the scan path until a follow-up can inject a one-rule source.

### `pkg/report/model`

- **`TestBuildCycloneDxReport`** uses local **`testdata/`** fixtures and runtime SHA-256 for expected `BomRef` / `Purl` / `Hashes`. **Replaces** hardcoded Unix/Windows hash tables and paths under `assets/queries/.../guardduty_detector_disabled/`. **Unblocks** CycloneDX tests when those rule files leave this repo.
- **`retargetSummaryFiles`** deep-copies shared summary mocks and rewrites paths to local testdata; fails if a mock `FileName` has no mapping. **Replaces** in-place mock mutation and silent drift when `test/helpers.go` paths change.

### `pkg/detector/terraform`

- **`TestResolveListIndexStatementInJsonencodeThirdElement`** reads **`testdata/jsonencode_third_element.tf`**. **Replaces** `assets/queries/.../s3_bucket_access_to_any_principal/test/positive3.tf`. **Unblocks** the `resolveListIndex` / `jsonencode` regression after that rule moves out of the embed.

### Not in this PR

- **`TestInspectorSimilarityID`** — still on `test/helpers.go` (~1.2k lines). **Deferred** until we split rule-agnostic mocks; then move to `pkg/engine/similarity/` with synthetic rules.

## How to test

```bash
go test -count=1 ./pkg/engine ./pkg/scan ./pkg/report/model ./pkg/detector/terraform
```

To confirm corpus-independence for the engine/report/detector tests (and that `Test_ExecuteScan` still runs as a smoke test): temporarily move `assets/queries/` aside and blank the queries embed in `assets/assets.go`, then re-run the command above with `-run` for the converted test names. That's what I am doing to verify what would break without the rules, once we move them to the new repo.

## Checklist

- [x] Reviewed my own PR
- [x] Added or updated unit tests
- [x] Tests pass locally
- [ ] Documentation (N/A)

### DIFF
--- a/pkg/detector/terraform/test/jsonencode_third_element.tf
+++ b/pkg/detector/terraform/test/jsonencode_third_element.tf
@@ -1,0 +1,61 @@
+resource "aws_s3_bucket" "this" {
+  bucket = "my_tf_test_bucket"
+  tags = {
+    Name = "My bucket"
+  }
+}
+
+resource "aws_s3_bucket_policy" "this" {
+  bucket = aws_s3_bucket.this.id
+  policy = jsonencode({
+    Version = "2012-10-17"
+    Statement = [{
+      # When used directly as a Cloudfront origin.
+      Effect = "Allow"
+      Action = "s3:GetObject"
+      Principal = {
+        Service = "cloudfront.amazonaws.com"
+      }
+      Resource = [
+        "${aws_s3_bucket.this.arn}/*",
+      ]
+      Condition = {
+        StringEquals = {
+          "AWS:SourceArn" = aws_cloudfront_distribution.this.arn
+        }
+      }
+      },
+      {
+        # Admin access for policy updates, etc.
+        Effect = "Allow"
+        Action = "s3:*"
+        Principal = {
+          AWS = [
+            data.aws_caller_identity.current.id,
+          ]
+        }
+        Resource = [
+          "${aws_s3_bucket.this.arn}/*",
+        ]
+      },
+      {
+        # Delegate access to the access point.
+        Effect = "Allow"
+        Action = "*"
+        Principal = {
+          AWS = [
+            "*"
+          ]
+        }
+        Resource = [
+          aws_s3_bucket.this.arn,
+          "${aws_s3_bucket.this.arn}/*",
+        ]
+        Condition = {
+          "StringEquals" = {
+            "s3:DataAccessPointAccount" = data.aws_caller_identity.current.account_id
+          }
+        }
+    }]
+  })
+}
--- a/pkg/detector/terraform/terraform_helper_test.go
+++ b/pkg/detector/terraform/terraform_helper_test.go
@@ -380,7 +380,7 @@ func TestGetArrayElementLine(t *testing.T) {
 }
 
 func TestResolveListIndexStatementInJsonencodeThirdElement(t *testing.T) {
-	data, err := os.ReadFile(filepath.Join("..", "..", "..", "assets", "queries", "terraform", "aws", "s3_bucket_access_to_any_principal", "test", "positive3.tf"))
+	data, err := os.ReadFile(filepath.Join("test", "jsonencode_third_element.tf"))
 	require.NoError(t, err)
 	linesPtr := utils.SplitLines(string(data))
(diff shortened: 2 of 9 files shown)


pkg/detector/terraform/test/jsonencode_third_element.tf
```diff
@@ -0,0 +1,61 @@
+resource "aws_s3_bucket" "this" {
```
ChouraquiBen, line 1: nit: wouldn't it be better if all the "testdata' were grouped under the test/fixture path?

pkg/engine/testing_test.go
```diff
@@ -0,0 +1,108 @@
+/*
```
ChouraquiBen: nit: The naming is kinda cryptic. I think that those tests could go inside inspector_test.go, what do you think?

pkg/engine/testing_test.go
```diff
@@ -0,0 +1,108 @@
...
+	needsLog            bool
+	numWorkers          int
+	kicsComputeNewSimID bool
+	vb                  VulnerabilityBuilder
+	tracker             Tracker
+}
+
+// newTestInspector runs the real [NewInspector] against a configurable
+// [source.QueriesSource], so callers get the production constructor wiring
+// without depending on `assets/queries/`.
```
ChouraquiBen: The comment will be very shortly deprecated since assets/queries will not be a path of this repo anymore

ChouraquiBen (dismissed):
Minor suggestions, otherwise lgtm!

pkg/engine/testing_test.go
whitemerch: It's expected, I have it part of my plan

pkg/detector/terraform/test/jsonencode_third_element.tf
whitemerch: i will rename the folder, but keep it next to the file that is using it

ChouraquiBen (approved):

PR #150, opened 2026-05-20
De-brittle Test_E2EExclusions, TestScanner_StartScan, TestGetQueriesWithLegacyIdFiltering

whitemerch:
Follow up of #149. Refactors four integration tests so their assertions are structural rather than slug- or count-coupled. No production changes.

### `test/e2e`

- **`Test_E2EExclusions`** now scans each baseline/inline-disable pair, computes the per-rule violation diff, and asserts the disable directive silences exactly one rule and adds none. **Replaces** the hardcoded `Rules: 1123 / 142 / 26` totals (4 of 7 historical commits to this file were just count bumps) and the per-slug `ViolationBreakdowns` equality. **Unblocks** the test from rule corpus changes (count bumps, new rules, slug renames) and keeps a real signal for the inline-disable behavior.

### `pkg/engine/source`

- **`TestGetQueriesWithLegacyIdFiltering`** now uses synthetic metadata and the combined `checkQueryInclude && !checkQueryExclude` predicate iterateQueryDirs applies on every rule. **Replaces** loading `alb_deletion_protection_disabled` from `assets/queries/`. **Unblocks** the test from the embed; include-only and exclude-only paths stay unit-covered by `TestCheckQueryIncludeWithLegacyId` / `TestCheckQueryExcludeWithLegacyId`. The remaining value is pinning that exclude wins when both filters target the same rule under different id forms.
- **`TestFilesystemSource_GetQueriesWithIncludeFilter`** gets a docstring earmarking it as the porting template for the eventual `TestDatadogSource_*` filter test when `FilesystemSource` is retired.

### `pkg/scanner`

- **`TestScanner_StartScan`** trimmed to the cancellation case with a pre-cancelled context (1ns timeout + a 10ms sleep before the call). **Replaces** the `time.Second*1` deadline that was flaky on slow runners and the redundant no-timeout smoke subtest. **Removes** the commented-out `TestScanner_ConcurrentScans` / `_ConcurrentScansWithTimeout` / `_StressTestConcurrency` blocks (K9VULN-8520) — they had been disabled for months and were just file noise.

## How to test

go test -count=1 ./pkg/engine/source ./pkg/scanner ./test/e2e

## Checklist

- [x] Reviewed my own PR
- [x] Added or updated unit tests
- [x] Tests pass locally
- [ ] Documentation (N/A)


pkg/scanner/scanner_test.go
```diff
@@ -145,186 +91,3 @@ func createServices(types, cloudProviders []string) (serviceSlice, *storage.Memo
...
-
-func createContext(ctx context.Context, timeout time.Duration) testContext {
-	ctx, cancel := context.WithTimeout(ctx, timeout)
-	return testContext{
-		ctx,
-		cancel,
-	}
-}
-
-// TODO: fix those tests see K9VULN-8520
```
ChouraquiBen, line 157: Checked: Cancelled project according to note

test/e2e/e2e_test.go
```diff
@@ -12,157 +12,84 @@ import (
...
 	}{
 		{
-			name:     "no exclusions",
-			testFile: filepath.Join("fixtures", "no-exclusions.tf"),
-			expectedOutput: scan.ScanStats{
-				Violations: 5,
-				Files:      1,
-				Rules:      1123,
-				ViolationBreakdowns: map[string]map[string]int{
-					"LOW": {
```
ChouraquiBen, line 29: I must say that the only time I had to fix those tests was when I had added or removed a rule to change the counts... I glad it ends here

pkg/engine/source/filesystem_test.go
```diff
@@ -675,92 +676,42 @@ func TestCheckQueryIncludeWithLegacyId(t *testing.T) {
...
-			description: "Exclude should take precedence - excluded by legacy ID",
-		},
-		{
-			name:        "include_legacy_exclude_new",
-			includeIDs:  []string{"afecd1f1-6378-4f7e-bb3b-60c35801fdd4"},
-			excludeIDs:  []string{"terraform-aws-alb-deletion-protection-disabled"},
-			shouldFind:  false,
-			description: "Exclude should take precedence - excluded by new ID",
-		},
+		{"include_by_legacy_id", []string{ruleLegacyID}, nil, true},
```
ChouraquiBen: remark: I don't know if it needs to be changed but not having the usual

```go
{
	name:        "include_by_legacy_id",
	includeIDs:  []string{ruleLegacyID},
	excludeIDs:  nil,
	keep:  true,
},
```

is weird. Plus it is the only place you do it in the PR

ChouraquiBen (dismissed):
LGTM

pkg/engine/source/filesystem_test.go
whitemerch: Will fix

ChouraquiBen (approved):
LGTM!!!

PR #152, opened 2026-05-20
Delete assets/queries/ and retire embedded rule corpus

whitemerch:
Follow up of #151. The backend rule pipeline is live and the FF `k9-iac-enable-kics-backend-rules` is on for all environments; the embedded rule corpus in `assets/queries/` is no longer served to end users.

## Changes

**`assets/assets.go`**
- Removed `//go:embed queries/...` directive.
- Stubbed `GetEmbeddedQueriesFs` / `GetEmbeddedQueryDirs` / `GetEmbeddedQueryFile` to return empty/nil/error so callers (`FilesystemSource`) gracefully serve zero rules rather than panicking.

**`assets/queries/` (52 MB)**
- Deleted in full. Rules are authoritative in `datadog-iac-scanner-default-rules` and served via `DatadogSource`.

**`pkg/engine/similarity_id_test.go` + `pkg/engine/testdata/simid/`**
- New synthetic Terraform rule (`testdata/simid/query.rego`) and two sample fixtures that exercise `TestInspectorSimilarityID` without the deleted corpus. Three scenarios: id-change → different simID, path-change → different simID, content-change-same-path → same simID.
- Previous `test/similarity_id_test.go` + `test/main_test.go` + `test/fixtures/tc-sim{01,02,03}/` removed.

Overall, I am not even sure we use it. Might be worth it to check it, and remove it if so.

**`pkg/engine/source/filesystem_test.go`**
- Deleted `TestFilesystemSource_GetQueriesWithIncludeFilter`: depended on embedded corpus; filtering logic is covered by `TestCheckQueryIncludeWithLegacyId` + `TestCheckQueryExcludeWithLegacyId`.
- Deleted `TestGetQueriesWithLegacyIdFiltering`: read a specific corpus file.

**`pkg/scanner/scanner_test.go`**
- Trimmed `TestScanner_StartScan` to a single smoke assertion (no timeout subtest: with zero queries the scan completes before any deadline fires).

**`test/`**: includes the PR #151 deletions: `test/queries_test.go`, `test/queries_content_test.go`, `test/main_test.go`, `test/similarity_id_test.go` all gone.

## Merge order

1. Merge #151 (corpus test cleanup) first.
2. Merge this PR second.

This PR is kept as draft until the team confirms the backend rollout is stable.

## How to test

CI should pass


whitemerch:
> ### 💡 Codex Review
> https://github.com/DataDog/datadog-iac-scanner/blob/a7a2479cd5060f44f3eaf0382b9d95c5a63c4a0e/pkg/engine/source/filesystem.go#L299-L302
> 
> 
> **P1 Badge Enable a rule source for default scans**
> When users run `datadog-iac-scanner scan` without the hidden `x-downloadqueriesfromdatadog` flag, `DownloadQueriesFromDatadog` remains false and `QueriesPath` is still `./assets/queries`, but this commit deletes that directory. This new path handling only logs and continues when the local rule path is absent, so `GetQueries` returns an empty rule set and default scans can complete with zero detections. Please either make the Datadog source the default or fail/load a real local ruleset when no query path exists.
> 
> ℹ️ About Codex in GitHub

@chatgpt-codex-connector This is true, but this is mainly because from now, we will always query on the API. This flag will be removed in a future PR

ChouraquiBen (approved):
Looks good to me:
- Except for assets.go, all changes are in assets/queries
- Changes in assets.go remove the embedded queries as intended
- Tried testing locally and rules worked

PR #154, opened 2026-05-21
User-friendly configuration error messages

jacobotb:
## Motivation

The error messages produced by the yaml decoder are a little uninformative. We can do better.

## Changes

* Added specific types for severities and categories so invalid values are rejected.
* Added a post-processing function that rewrites yaml error messages to be more useful.
* Also, remove some spurious newlines and add a stack trace when we recover from a panic.

## Author Checklist

## QA Instruction

Create a `code-security.datadog.yaml` file with an invalid configuration and use `go run ./cmd/scanner show-config`.

```
% go run ./cmd/scanner show-config
Program failed: error reading the configuration: could not parse configuration file: yaml: unmarshal errors:
  line 3: field `foo` is not valid in the `iac` section
exit status 126
```

Other examples of error messages:

* Original messages:
```
line 123: field use-rules not found in type config.iacGlobalConfig
line 123: field ignore-rulesets not found in type config.iacConfig
line 123: mapping key "use-rules" already defined at line 12
line 123: cannot unmarshal !!int `1234` into map
```
* Rewritten messages:
```
line 123: field `use-rules` is not valid in the `iac.global-config` section; maybe you should include it in the `iac` section instead
line 123: field `ignore-rulesets` is not valid in the `iac` section; it looks like a datadog-static-analyzer configuration option, not a datadog-iac-scanner option
line 123: field `use-rules` has already been set in line 12
line 123: value `1234` of type int is not valid here; a map was expected
```
* New messages (for severity and category):
```
line 123: invalid severity: "not that bad"
line 123: invalid category: "giants that look like windmills"
```

## Blast Radius

For now, the changes will be visible to users running the scanner on their console or CI.


pkg/config/config_test.go
```diff
@@ -22,11 +22,7 @@ func TestParseSchemaVersion(t *testing.T) {
 	_, err = ParseConfig([]byte("schema-version: 1.1\niac: {ignore-rules: [\"abc\"]}"))
 	assert.Error(t, err, "Invalid schema-version was expected to be rejected")
 
-	cfg, err := ParseConfig([]byte("schema-version: v1.0\niac: {ignore-rules: [\"abc\"]}"))
-	assert.Nil(t, cfg, "Too low schema-version should result in empty IaC config")
-	assert.NoError(t, err)
-
-	cfg, err = ParseConfig([]byte("schema-version: v1.2\niac: {ignore-rules: [\"abc\"]}"))
```
whitemerch, line 29: Might be worth it to test the behavior changes here wdyt?

whitemerch (dismissed):

pkg/config/config_test.go
jacobotb: Added a test for that.

whitemerch (approved):

PR #155, opened 2026-05-22
Emit SARIF suppressions for in-source ignored IaC findings

whitemerch:
## Motivation

Findings suppressed via `# dd-iac-scan` directives (`disable`, `ignore-line`, `ignore-block`) were dropped before SARIF generation, so downstream consumers could not record them as suppressed-in-code. Static analysis already emits SARIF `suppressions[]` for the same pattern; this change aligns IaC scanner behavior so the existing pipeline can surface the suppressed badge without further scanner-side changes.

## Changes

- Tag suppressed vulnerabilities in the engine instead of dropping them (`IsSuppressed`, `SuppressionKind`, `SuppressionJustification`).
- Propagate suppression metadata through `VulnerableFile` / `CreateSummary`.
- Exclude suppressed files from severity counters (CLI exit codes unchanged).
- Emit SARIF `result.suppressions[]` with `kind: inSource` and `status: accepted` for suppressed findings.
- Add unit tests for engine suppression paths, SARIF emission, and summary counting.

## Author Checklist

## QA Instruction

Run `go test ./pkg/engine/... ./pkg/report/model/... -count=1`, then scan a fixture that uses `# dd-iac-scan ignore-line` and inspect the SARIF output: the matching `result` should include a `suppressions` array with `kind: "inSource"` and `status: "accepted"`, while non-suppressed results must omit `suppressions`.

## Blast Radius

This PR only changes `datadog-iac-scanner` output. SARIF reports gain `suppressions` entries for previously dropped findings Severity counters and CLI failure semantics stay the same for non-suppressed findings.

## Additional Notes

Downstream consumers already handles `suppressions[]` for IaC tool results.

[K9VULN-15194]: https://datadoghq.atlassian.net/browse/K9VULN-15194?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/report/model/sarif.go
+++ b/pkg/report/model/sarif.go
@@ -140,7 +140,30 @@ type sarifResult struct {
 	ResultLevel         string                   `json:"level"`
 	ResultProperties    sarifProperties          `json:"properties,omitempty"`
 	ResultFixes         []model.SarifFix         `json:"fixes,omitempty"`
+	// Suppressions follows the SARIF 2.1.0 result.suppressions shape so
+	// downstream consumers (for example ci-sarif-processor) can mark the
+	// finding as suppressed instead of dropping it. The slice is omitted
+	// from JSON when empty to keep non-suppressed results byte-for-byte
+	// identical to the previous output.
+	Suppressions []sarifSuppression `json:"suppressions,omitempty"`
 }
+
+// sarifSuppression mirrors the SARIF 2.1.0 suppression object. Only the
+// fields that the scanner populates are declared; optional fields like
+// `guid` and `location` are left out so the encoder never emits `null`
+// values for them.
+type sarifSuppression struct {
+	Kind          string `json:"kind"`
+	Status        string `json:"status,omitempty"`
+	Justification string `json:"justification,omitempty"`
+}
+
+const (
+	// sarifSuppressionStatusAccepted is the only suppression status this
+	// scanner emits today. Suppression directives are user-authored and
+	// therefore always accepted by definition.
+	sarifSuppressionStatusAccepted = "accepted"
+)
 
 type SarifPartialFingerprints struct {
 	Sha                string `json:"SHA,omitempty"`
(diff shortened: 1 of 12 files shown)


pkg/report/model/sarif.go
```diff
@@ -602,6 +616,27 @@ func (sr *sarifReport) BuildSarifIssue(ctx context.Context, issue *model.QueryRe
...
+// because every current suppression path either matches that exactly or
+// sets the kind explicitly.
+func buildSarifSuppressions(file *model.VulnerableFile) []sarifSuppression {
+	if file == nil || !file.IsSuppressed {
+		return nil
+	}
+	kind := file.SuppressionKind
+	if kind == "" {
+		kind = model.SuppressionKindInSource
+	}
```
MikaYuoadas, line 636: 💡 *Suggestion:* As far as I see in this PR it can’t ever happen today. But if we add other kind in the future, silently defaulting to this could hide a bug. Shouldn’t we instead have some "Unknown" kind of some sort to explicitely show error cases?

MikaYuoadas (dismissed):
LGTM, just made a small optional suggestion.

pkg/report/model/sarif.go
whitemerch: I see what you mean indeed, but SARIF 2.1.0 spec, which is what we are using, defines suppression.kind as an enum with only two legal values: inSource and external (see [SARIF 2.1.0 3.35.2](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317897)). Emitting "unknown" would produce schema-invalid SARIF and likely make ci-sarif-processor reject, drop the result, or panic (I didn't check the code in details, breaking the very pipeline this PR exists to feed. What I can do instead, is ddrrop the fallback and treat an empty kind as a warning log. Wdyt?

MikaYuoadas: Oh ok wasn’t aware this field only had `inSource` and `external`. Well seems fine then, I would let it as is (and maybe add a warning log line so we’re aware if that happens).

MikaYuoadas (approved):

PR #157, opened 2026-05-28
Drop the --x-downloadqueriesfromdatadog flag and always fetch rules from the backend

whitemerch:
## Summary

The rules backend is fully rolled out and `--x-downloadqueriesfromdatadog` has defaulted to `true` for a while. The flag and the `DownloadQueriesFromDatadog` parameter are now dead config.

This PR:

- Drops the `--x-downloadqueriesfromdatadog` CLI flag.
- Drops the `DownloadQueriesFromDatadog` field from `pkg/scan.Parameters`.
- Collapses the branch in `pkg/scan.createQuerySource` and `cmd/scanner/list_queries.go::getQuerySource` so they always return a `DatadogSource` (with a `FilesystemSource` for libraries).

## Merge order

Draft until [DataDog/dd-source#450905](https://github.com/DataDog/dd-source/pull/450905) (drop the `k9-iac-enable-kics-backend-rules` feature flag in `iac-scanning`) is merged **and** a scanner release with this change has been picked up downstream. Merging this PR before that lands will cause `iac-scanning` to invoke the scanner with an unknown flag.

## Test plan

- [ ] `go test ./...`
- [ ] Verify scan-from-binary still works end-to-end against the rules backend


ChouraquiBen (approved):
Also looks good to me, let's ship!

PR #159, opened 2026-05-28
Decouple scanner tests from the deleted rule corpus

whitemerch:
## Summary

Now that `assets/queries/` is gone, a few corners of the test suite were either silently no-op-ing or pointing at paths that no longer exist. This PR is pure test housekeeping:

- **`pkg/scan/scan_test.go::Test_ExecuteScan`**: now exercises the real scan pipeline against a synthetic in-memory rule via a new `Client.querySourceFactory` hook. Previously the test ran with zero rules (because `assets/queries/` was deleted) and asserted nothing meaningful.
- **`internal/console/helpers/helpers_test.go::TestHelpers_GetDefaultQueryPath`**: swap the corpus path for `assets/libraries` (still on disk) and drop a misleading comment about embedded queries.
- **`test/helpers.go` + `pkg/report/model/cyclonedx_test.go`**: replace the mock `FileName` labels under `assets/queries/...` with corpus-neutral `testdata/...` strings. They are just identifiers shared between the two files; nothing reads them from disk.
- **`pkg/engine/source/filesystem_test.go`**: drop `BenchmarkFilesystemSource_GetQueries` (benchmarked the now-empty corpus) and two stale `// Should be uncommented when UUID are moved to LegacyId` comments.

## Test plan

- [ ] `go test ./...`
- [ ] CI green

### DIFF
--- a/pkg/report/model/cyclonedx_test.go
+++ b/pkg/report/model/cyclonedx_test.go
@@ -194,8 +194,8 @@ func TestBuildCycloneDxReport(t *testing.T) {
 	// paths. Keyed on slash form because that's what filepath.ToSlash returns
 	// on both Windows and Unix.
 	mapping := map[string]string{
-		filepath.ToSlash(filepath.Join("assets", "queries", "terraform", "aws", "guardduty_detector_disabled", "test", "positive.tf")): positivePath,
-		filepath.ToSlash(filepath.Join("assets", "queries", "terraform", "aws", "guardduty_detector_disabled", "test", "negative.tf")): negativePath,
+		filepath.ToSlash(filepath.Join("testdata", "guardduty_detector_disabled", "positive.tf")): positivePath,
+		filepath.ToSlash(filepath.Join("testdata", "guardduty_detector_disabled", "negative.tf")): negativePath,
 		filepath.ToSlash(filepath.Join("test", "fixtures", "test_critical_custom_queries", "amazon_mq_broker_encryption_disabled", "test", "positive1.yaml")): criticalPath,
 	}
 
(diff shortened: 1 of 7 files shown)


pkg/report/model/cyclonedx_test.go
```diff
@@ -194,8 +194,8 @@ func TestBuildCycloneDxReport(t *testing.T) {
 	// paths. Keyed on slash form because that's what filepath.ToSlash returns
 	// on both Windows and Unix.
 	mapping := map[string]string{
-		filepath.ToSlash(filepath.Join("assets", "queries", "terraform", "aws", "guardduty_detector_disabled", "test", "positive.tf")): positivePath,
-		filepath.ToSlash(filepath.Join("assets", "queries", "terraform", "aws", "guardduty_detector_disabled", "test", "negative.tf")): negativePath,
+		filepath.ToSlash(filepath.Join("testdata", "guardduty_detector_disabled", "positive.tf")): positivePath,
+		filepath.ToSlash(filepath.Join("testdata", "guardduty_detector_disabled", "negative.tf")): negativePath,
```
ChouraquiBen, line 198: Is this file supposed to exist yet?

ChouraquiBen: Does not look like existence is tested but may not be interesting to fix any further if as discussed, cyclonedx is about BOMs

test/helpers.go
```diff
@@ -129,7 +129,7 @@ var queryMedium2 = model.QueryResult{
 	Severity:  model.SeverityMedium,
 	Files: []model.VulnerableFile{
 		{
-			FileName:         filepath.Join("assets", "queries", "terraform", "aws", "guardduty_detector_disabled", "test", "positive.tf"),
+			FileName:         filepath.Join("testdata", "guardduty_detector_disabled", "positive.tf"),
```
ChouraquiBen, line 132: Same as above as it also applies for cyclonedx

ChouraquiBen (approved):
Thanks for keeping the code clean!

PR #160, opened 2026-05-28
Fix CloudFormation short-form intrinsic tag parsing and resolve resource names from parameter defaults

whitemerch:
## Summary

- Rewrites CloudFormation short-form intrinsic tags (`!Sub`, `!Ref`, `!GetAtt`, `!Join`, and others) into their canonical long-form map representation in the YAML parser (`pkg/model/model_yaml.go`) so custom tags are preserved instead of being flattened into raw strings.
- Ensures rewritten short-form intrinsic maps also carry `_kics_lines` metadata, matching long-form map behavior for line attribution.
- Extends `get_resource_name` in `assets/libraries/cloudformation.rego` to resolve CloudFormation intrinsic name fields from:
  - scalar `Fn::Sub`
  - sequence-form `Fn::Sub` (`[template, vars]`)
  - `Ref`
  - CloudFormation pseudo-parameters (`AWS::Region`, `AWS::AccountId`, etc.)

## Behavior change

| Template syntax | Before | After |
|---|---|---|
| `BucketName: my-literal-bucket` | `my-literal-bucket` | `my-literal-bucket` |
| `BucketName: !Sub my-app-${Env}` with `Env` default `prod` | `my-app-${Env}` | `my-app-prod` |
| `BucketName: !Sub ["${App}-${Env}", {App: myapp, Env: prod}]` | `S3Bucket` (logical id fallback) | `myapp-prod` |
| `BucketName: !Sub my-app-${AWS::Region}-${AWS::AccountId}` | logical id fallback | `my-app-aws-region-aws-account-id` |
| `BucketName: !Ref BucketParam` with default `my-bucket` | logical id fallback | `my-bucket` |
| `BucketName: !Ref AWS::Region` | logical id fallback | `aws-region` |

## Notes

- `Fn::Sub` name resolution only returns when all placeholders are resolved; unresolved `${...}` still falls back to tag name/logical id.
- For pseudo-parameters we resolve to deterministic placeholders (`aws-region`, `aws-account-id`, etc.) so names stay stable and readable without requiring runtime stack context.

## Test plan

- [x] `go test ./pkg/model/...` (short-form intrinsic parser tests, including `_kics_lines` parity test)
- [x] `opa test assets/libraries/cloudformation.rego assets/libraries/common.rego assets/libraries/test/cloudformation_test.rego --v0-compatible` (69/69 pass, including new sequence-form/pseudo-parameter cases)
- [x] `go test ./e2e/...`
- [x] Local scan of a CloudFormation template with `BucketName: !Sub my-app-${Env}` and `Owner` default `prod` reports `IAC_RESOURCE_NAME:my-app-prod`

### DIFF
--- a/assets/libraries/cloudformation.rego
+++ b/assets/libraries/cloudformation.rego
@@ -262,6 +262,27 @@ get_resource_name(resource, resourceDefinitionName) = name {
 	fieldValue != ""
 	name := fieldValue[_]
 } else = name {
+	field := resourceFieldName[resource.Type]
+	template := resource.Properties[field]["Fn::Sub"]
+	is_string(template)
+	params := input.document[_].Parameters
+	replacements := {old: new |
+		some param_name
+		param := params[param_name]
+		is_string(param.Default)
+		old := sprintf("${%s}", [param_name])
+		new := param.Default
+	}
+	resolved := strings.replace_n(replacements, template)
+	not contains(resolved, "${")
+	resolved != ""
+	name := resolved
+} else = name {
+	field := resourceFieldName[resource.Type]
+	ref_name := resource.Properties[field].Ref
+	name := input.document[_].Parameters[ref_name].Default
+	name != ""
+} else = name {
 	name := common_lib.get_tag_name_if_exists(resource)
 } else = name {
 	name := resourceDefinitionName
--- a/assets/libraries/cloudformation.rego
+++ b/assets/libraries/cloudformation.rego
@@ -249,6 +249,85 @@ resourceFieldName = {
 	"AWS::Serverless::Function": "FunctionName",
 }
 
+pseudo_parameter_default_replacements = {
+	"${AWS::AccountId}":   "aws-account-id",
+	"${AWS::NotificationARNs}": "aws-notification-arns",
+	"${AWS::NoValue}":     "aws-no-value",
+	"${AWS::Partition}":   "aws",
+	"${AWS::Region}":      "aws-region",
+	"${AWS::StackId}":     "aws-stack-id",
+	"${AWS::StackName}":   "aws-stack-name",
+	"${AWS::URLSuffix}":   "amazonaws.com",
+}
+
+parameter_default_replacements = replacements {
+	params := input.document[_].Parameters
+	replacements := {old: new |
+		some param_name
+		param := params[param_name]
+		param.Default != null
+		old := sprintf("${%s}", [param_name])
+		new := sprintf("%v", [param.Default])
+	}
+} else = replacements {
+	replacements := {}
+}
+
+sub_template(sub_expr) = template {
+	is_string(sub_expr)
+	template := sub_expr
+} else = template {
+	is_array(sub_expr)
+	count(sub_expr) > 0
+	is_string(sub_expr[0])
+	template := sub_expr[0]
+}
+
+intrinsic_value_to_string(v) = stringified {
+	is_string(v)
+	stringified := v
+} else = stringified {
+	is_number(v)
+	stringified := sprintf("%v", [v])
+} else = stringified {
+	is_boolean(v)
+	stringified := sprintf("%v", [v])
+} else = stringified {
+	common_lib.valid_key(v, "Ref")
+	ref_name := v.Ref
+	stringified := input.document[_].Parameters[ref_name].Default
+	stringified != null
+} else = stringified {
+	common_lib.valid_key(v, "Ref")
+	ref_name := v.Ref
+	old := sprintf("${%s}", [ref_name])
+	stringified := pseudo_parameter_default_replacements[old]
+}
+
+sequence_sub_replacements(sub_expr) = replacements {
+	is_array(sub_expr)
+	count(sub_expr) > 1
+	vars := sub_expr[1]
+	is_object(vars)
+	replacements := {old: new |
+		some var_name
+		raw := vars[var_name]
+		new := intrinsic_value_to_string(raw)
+		old := sprintf("${%s}", [var_name])
+	}
+} else = replacements {
+	replacements := {}
+}
+
+resolve_sub_name(sub_expr) = resolved {
+	template := sub_template(sub_expr)
+	step1 := strings.replace_n(parameter_default_replacements, template)
+	step2 := strings.replace_n(pseudo_parameter_default_replacements, step1)
+	resolved := strings.replace_n(sequence_sub_replacements(sub_expr), step2)
+	not contains(resolved, "${")
+	resolved != ""
+}
+
 get_resource_name(resource, resourceDefinitionName) = name {
 	field := resourceFieldName[resource.Type]
 	fieldValue := resource.Properties[field]
@@ -263,25 +342,19 @@ get_resource_name(resource, resourceDefinitionName) = name {
 	name := fieldValue[_]
 } else = name {
 	field := resourceFieldName[resource.Type]
-	template := resource.Properties[field]["Fn::Sub"]
-	is_string(template)
-	params := input.document[_].Parameters
-	replacements := {old: new |
-		some param_name
-		param := params[param_name]
-		is_string(param.Default)
-		old := sprintf("${%s}", [param_name])
-		new := param.Default
-	}
-	resolved := strings.replace_n(replacements, template)
-	not contains(resolved, "${")
-	resolved != ""
-	name := resolved
+	sub_expr := resource.Properties[field]["Fn::Sub"]
+	name := resolve_sub_name(sub_expr)
 } else = name {
 	field := resourceFieldName[resource.Type]
 	ref_name := resource.Properties[field].Ref
 	name := input.document[_].Parameters[ref_name].Default
+	name != null
 	name != ""
+} else = name {
+	field := resourceFieldName[resource.Type]
+	ref_name := resource.Properties[field].Ref
+	old := sprintf("${%s}", [ref_name])
+	name := pseudo_parameter_default_replacements[old]
 } else = name {
 	name := common_lib.get_tag_name_if_exists(resource)
 } else = name {
--- a/assets/libraries/cloudformation.rego
+++ b/assets/libraries/cloudformation.rego
@@ -273,6 +273,14 @@ parameter_default_replacements = replacements {
 	replacements := {}
 }
 
+# sub_template extracts the template string from an Fn::Sub value.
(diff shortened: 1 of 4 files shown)


assets/libraries/cloudformation.rego
```diff
@@ -249,6 +249,85 @@ resourceFieldName = {
 	"AWS::Serverless::Function": "FunctionName",
 }
 
+pseudo_parameter_default_replacements = {
```
ChouraquiBen, line 252: imho, there are already many mappings, this will only make the code more difficult to maintain. Would it seem reasonable to use sort of a `toSlug` function here so that `AWS::NoValue -> aws-no-value` and `AWS::Partition -> aws-partition`?

assets/libraries/cloudformation.rego
```diff
@@ -249,6 +249,85 @@ resourceFieldName = {
...
+	is_string(sub_expr[0])
+	template := sub_expr[0]
+}
+
+intrinsic_value_to_string(v) = stringified {
+	is_string(v)
+	stringified := v
+} else = stringified {
+	is_number(v)
+	stringified := sprintf("%v", [v])
```
ChouraquiBen, line 299: nit:
```suggestion
	stringified := sprintf("%d", [v])
```

assets/libraries/cloudformation.rego
```diff
@@ -249,6 +249,85 @@ resourceFieldName = {
...
+	"${AWS::NoValue}":     "aws-no-value",
+	"${AWS::Partition}":   "aws",
+	"${AWS::Region}":      "aws-region",
+	"${AWS::StackId}":     "aws-stack-id",
+	"${AWS::StackName}":   "aws-stack-name",
+	"${AWS::URLSuffix}":   "amazonaws.com",
+}
+
+parameter_default_replacements = replacements {
+	params := input.document[_].Parameters
```
ChouraquiBen, line 264: TIL: input is a global variable, so you can indeed use it this way

assets/libraries/cloudformation.rego
```diff
@@ -249,6 +249,85 @@ resourceFieldName = {
...
+	stringified := input.document[_].Parameters[ref_name].Default
+	stringified != null
+} else = stringified {
+	common_lib.valid_key(v, "Ref")
+	ref_name := v.Ref
+	old := sprintf("${%s}", [ref_name])
+	stringified := pseudo_parameter_default_replacements[old]
+}
+
+sequence_sub_replacements(sub_expr) = replacements {
```
ChouraquiBen, line 320: Suggestion: This is rather obscure. `sub_expr` should be an array and the second element is used and should be an object... tbh I am rather lost, could you please add a comment as to what `sub_expr` is and should look like?

assets/libraries/cloudformation.rego
```diff
@@ -249,6 +249,85 @@ resourceFieldName = {
...
+		some var_name
+		raw := vars[var_name]
+		new := intrinsic_value_to_string(raw)
+		old := sprintf("${%s}", [var_name])
+	}
+} else = replacements {
+	replacements := {}
+}
+
+resolve_sub_name(sub_expr) = resolved {
```
ChouraquiBen, line 340: I see, so for example, 
```
[
	"${AppName}-${Env}-bucket",
	{"AppName": "myapp", "Env": "prod"},
]
```
is a `sub_expr`. The template is the first part and the variables will be put in the following object.

assets/libraries/cloudformation.rego
```diff
@@ -249,6 +249,85 @@ resourceFieldName = {
...
+	template := sub_template(sub_expr)
+	step1 := strings.replace_n(parameter_default_replacements, template)
+	step2 := strings.replace_n(pseudo_parameter_default_replacements, step1)
+	resolved := strings.replace_n(sequence_sub_replacements(sub_expr), step2)
```
ChouraquiBen: Question: I don't know if it is a valid expression but that means that if the attribute exists both in the `Parameters` and the `sub_expr`, the `Parameters` value will be used. Is it the expected outcome? I don't think that I have seen a test that would have made this clear

assets/libraries/cloudformation.rego
```diff
@@ -261,6 +340,21 @@ get_resource_name(resource, resourceDefinitionName) = name {
 	is_string(fieldValue)
 	fieldValue != ""
 	name := fieldValue[_]
+} else = name {
+	field := resourceFieldName[resource.Type]
+	sub_expr := resource.Properties[field]["Fn::Sub"]
```
ChouraquiBen, line 363: Suggestion: could you add more tests to `get_resource_name`? In particular, I am wondering if it will correctly skip with some input like:
```
resource := {
	"Type": "AWS::S3::Bucket",
	"Properties": {"BucketName": {"Fn::Sub": [
        {"Fn::If": [
            ref_to_condition, value_true, value_false]}, 
            {"AppName": "myapp", "Env": "prod"},
    ]}},
}
```
More complex expressions might be hiding behind the `Fn::Sub`, so, that leaves me wondering 🤔

pkg/model/model_yaml.go
```diff
@@ -9,12 +9,37 @@ import (
...
 	"errors"
+	"strings"
 
 	"github.com/DataDog/datadog-iac-scanner/pkg/logger"
 	"github.com/DataDog/datadog-iac-scanner/pkg/utils"
 	"gopkg.in/yaml.v3"
 )
 
+// cfnShortFormTags maps CloudFormation short-form intrinsic tags to long-form keys.
+var cfnShortFormTags = map[string]string{
```
ChouraquiBen, line 20: Same question as previously: would it be possible to replace with a function? Would a list of the allowed tag do since the transformation is straightforward? I don't know if it really helps maintainability

pkg/model/model_yaml.go
```diff
@@ -71,6 +96,13 @@ func unmarshalWithDepth(ctx context.Context, val *yaml.Node, visited map[*yaml.N
 		// iterate two by two, since first iteration is the key and the second is the value
 		for i := 0; i < len(val.Content); i += 2 {
 			if val.Content[i].Kind == yaml.ScalarNode {
+				if rewritten, ok := rewriteCFNShortFormIntrinsic(ctx, val.Content[i+1], visited, ignore); ok {
```
ChouraquiBen, line 133: I would say must-fix but up for debate: Could you please add a test on the platform so that only CloudFormation files are affected?

pkg/model/model_yaml.go
```diff
@@ -181,3 +216,67 @@ func scalarNodeResolver(ctx context.Context, val *yaml.Node) interface{} {
 	}
 	return resolved
 }
+
+// rewriteCFNShortFormIntrinsic converts a short-form CFN intrinsic to its long-form map.
+func rewriteCFNShortFormIntrinsic(ctx context.Context, val *yaml.Node, visited map[*yaml.Node]bool, ignore *Ignore) (interface{}, bool) {
```
ChouraquiBen, line 256: I think the test on the platform would better go here actually: the function explicitly states that it is CloudFormation only and returns a bool for failures, so it seems like the perfect place to me

ChouraquiBen (commented):
Woah! What a change!
For now I have a few comments, but that seems like a good start to reduce the potential discrepancies

pkg/model/model_yaml.go
whitemerch: The mapping isn't mechanically derivable (e.g. !Condition → "Condition", !GetAtt takes special-case splitting logic). A function would add complexity without benefit here. I guess that's how we have to handle it for these cases

pkg/model/model_yaml.go
whitemerch: Done

pkg/model/model_yaml.go
whitemerch: Done

assets/libraries/cloudformation.rego
whitemerch: I agree, but the values aren't mechanically sluggable: AWS::Partition → "aws" (not aws-partition), AWS::URLSuffix → "amazonaws.com" (not a slug at all)

assets/libraries/cloudformation.rego
whitemerch: hehe

assets/libraries/cloudformation.rego
whitemerch: Rego numbers are IEEE 754 doubles, so a value like 42.0 satisfies `is_number` but `%d` would panic/error on it, whereas `%v` always produces `"42"`. `%v` is consistent with the boolean branch immediately above it too

assets/libraries/cloudformation.rego
whitemerch: Added comments above `sub_template` and `sequence_sub_replacements` explaining the string vs. sequence form, and what the second element of the sequence represents

assets/libraries/cloudformation.rego
whitemerch: fixed

assets/libraries/cloudformation.rego
whitemerch: Safe fallback, `sub_template` requires `is_string(sub_expr[0])`, so when the head is a Fn::If map, the rule produces no result and `get_resource_name` falls through to the logical ID. Added `test_get_resource_name_fn_sub_complex_head_falls_back` to document and lock this in.

assets/libraries/cloudformation.rego
ChouraquiBen: Alright! Let's do that!

assets/libraries/cloudformation.rego
ChouraquiBen: hmmm, I don't see them, did you push? 🤔

pkg/model/model_yaml.go
ChouraquiBen: Thanks!

assets/libraries/cloudformation.rego
whitemerch: yes
https://github.com/DataDog/datadog-iac-scanner/pull/160/changes/80f1271abd0a681a08002610ffc1b9dadd34a562#diff-7f2b4262cfcd2499d4a906bbfafc0fd09c3045b21c985dd9c5e6f1fb65d87943

ChouraquiBen (approved):
🚢

PR #161, opened 2026-05-28
Expand CloudFormation Fn::ForEach resources for truthful resource names

whitemerch:
## Summary

- Expands CloudFormation `Fn::ForEach::*` resources during YAML parsing (`pkg/model/model_yaml.go`) so loop-generated resources are materialized as concrete logical IDs before Rego rules run.
- Resolves loop variable placeholders in resource keys and nested values (for example `${Identifier}` → `0`, `1`) using the ForEach collection.
- Resolves `Ref` to loop variables inside expanded resources (for example `Ref: Identifier` → `"0"`).
- Supports ForEach collections from literal lists and from `Ref` to parameter defaults (including comma-separated defaults like `"0,1"`).

## Behavior change

| Template syntax | Before | After |
|---|---|---|
| `Fn::ForEach::RDSInstances` with `RDSInstance${Identifier}` and `Instances` default `"0"` | `RDSInstance${Identifier}` | `RDSInstance0` |
| Same with `Instances` default `"0,1"` | `RDSInstance${Identifier}` | `RDSInstance0`, `RDSInstance1` |
| `Ref: Identifier` inside expanded resource properties | `{"Ref":"Identifier"}` | `"0"` (per iteration) |
| `Fn::Sub app_${Identifier}` inside expanded resource properties | `Fn::Sub app_${Identifier}` | `Fn::Sub app_0` (per iteration) |
| Non-ForEach resources | unchanged | unchanged |

## Example

Template pattern CloudFormation:

```yaml
Parameters:
  Instances:
    Type: List<Number>
    Default: "0"
Resources:
  Fn::ForEach::RDSInstances:
    - Identifier
    - !Ref Instances
    - RDSInstance${Identifier}:
        Type: AWS::RDS::DBInstance
```

| | Resource name |
|---|---|
| Before | `RDSInstance${Identifier}` |
| After | `RDSInstance0` |

## Notes

- Expansion uses parameter defaults only (static analysis); runtime stack parameters are not available at scan time.
- If the ForEach collection cannot be resolved, the original `Fn::ForEach::*` entry is left unchanged.
- Depends on short-form intrinsic preservation from #160 (`!Ref` in the collection expression must parse as `{"Ref": "..."}`).

## Test plan

- [x] `go test ./pkg/model/ -run CloudFormationForEachExpansion -v`
- [x] `go test ./pkg/model/...`
- [x] Local scan of example reports `IAC_RESOURCE_NAME:RDSInstance0` (not `RDSInstance${Identifier}`) for loop-generated RDS findings

### DIFF
--- a/pkg/model/model_yaml.go
+++ b/pkg/model/model_yaml.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	json "encoding/json"
 	"errors"
+	"fmt"
 	"strings"
 
 	"github.com/DataDog/datadog-iac-scanner/pkg/logger"
@@ -78,6 +79,7 @@ func (m *Document) UnmarshalYAML(ctx context.Context, value *yaml.Node, ignore *
 	ctx = withCloudFormation(ctx, isCloudFormationDocument(value))
 	dpc := unmarshal(ctx, value, ignore)
 	if mapDcp, ok := dpc.(map[string]interface{}); ok {
+		expandCloudFormationForEach(mapDcp)
 		// set line information for root level objects
 		mapDcp["_kics_lines"] = getLines(value, 0)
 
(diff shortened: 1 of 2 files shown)


pkg/model/model_yaml.go
```diff
@@ -89,6 +91,205 @@ func (m *Document) UnmarshalYAML(ctx context.Context, value *yaml.Node, ignore *
...
+	for {
+		forEachKeys := collectForEachKeys(resources)
+		if len(forEachKeys) == 0 {
+			break
+		}
+
+		anyExpanded := false
+		for _, key := range forEachKeys {
+			spec, ok := resources[key].([]interface{})
+			if !ok || len(spec) != 3 {
```
ChouraquiBen: Suggestion: a short comment with what the spec should look like would be helpful -- this is counterbalanced by the well-named variables below

pkg/model/model_yaml.go
```diff
@@ -78,6 +79,7 @@ func (m *Document) UnmarshalYAML(ctx context.Context, value *yaml.Node, ignore *
 	ctx = withCloudFormation(ctx, isCloudFormationDocument(value))
 	dpc := unmarshal(ctx, value, ignore)
 	if mapDcp, ok := dpc.(map[string]interface{}); ok {
+		expandCloudFormationForEach(mapDcp)
```
ChouraquiBen: Cloud you please gate this behind a test so that it only triggers if the document is CloudFormation?

ChouraquiBen (dismissed):
Honestly, my biggest concern right now is that the default yaml parser is only getting bigger. This will decrease performance, slowing down the scans - and also increase much the responsibilities of the parser itself. Otherwise, the code looks good to me

whitemerch:
> Honestly, my biggest concern right now is that the default yaml parser is only getting bigger. This will decrease performance, slowing down the scans - and also increase much the responsibilities of the parser itself. Otherwise, the code looks good to me

@ChouraquiBen Agree we shouldn’t keep growing the default YAML parser, this is gated to CFN docs + AWS::LanguageExtensions + actual Fn::ForEach entries so other YAML shouldn’t hit it. Parse-time expansion is so rules see real logical IDs before Rego runs, but this is better than reporting wrong values/findings. Overall, there are some fixes/architectural rethinking we should do, but not sure we can do anything about it right now

ChouraquiBen (approved):
LGTM

PR #162, opened 2026-05-29
IaC config v1.3 support

whitemerch:
## Motivation

This PR introduces `schema-version: v1.3` for `code-security.datadog.yaml`:

* **Per-rule path scoping**: `rule-configs.<rule-id>.ignore-paths` and `only-paths` drop findings for a specific rule outside the configured path set.
* **Global platform filters**: `iac.global-config.only-platforms` and `ignore-platforms` restrict which platforms are scanned, intersecting with the `--type` CLI flag.
* **Per-rule severity override**: `rule-configs.<rule-id>.severity` overrides the default severity of a rule's findings.

Schema v1.2 files remain fully supported; the scanner auto-emits `v1.3` from `UnparseConfig` only when new fields are in use.

## Changes

* New `rule-configs` map in the IaC config allows per-rule path scoping and severity overrides.
* New `only-platforms` / `ignore-platforms` fields in `global-config` filter which platforms are scanned.
* `SupportedPlatforms` is now a shared constant, keeping platform validation and CLI defaults in sync.

## Tests

Unit tests cover config parsing, platform validation, inspector path matching (`matchesPath`, `rulePathExcluded`), and CLI platform filter logic.

A new e2e test suite (`test/e2e/v13_test.go`) exercises all three features end-to-end using local rule fixtures and real scan execution, across 42 test cases:

**Per-rule `ignore-paths`**
- Drops findings from the specified directory (hard-drop, not SARIF suppression).
- Keeps findings from all other directories.
- Dropping all dirs for a rule produces zero findings for that rule only.
- Other rules are completely unaffected.

**Per-rule `only-paths`**
- Restricts findings to the specified directory.
- Drops findings outside that directory.
- If `only-paths` matches nothing, zero findings are produced for that rule.
- Multiple directories in `only-paths` all produce findings.
- `ignore-paths` takes precedence over `only-paths` when both are set on the same rule.

**Per-rule severity override**
- Changes the severity in scan stats and SARIF output.
- All five levels work: `critical`, `high`, `medium`, `low`, `info`.
- Finding count is unchanged by an override.
- Other rules' severities are unaffected.
- SARIF `level` field is correct: `critical→error`, `high→warning`, `medium→note`, `low/info→none`.

**Combined rule-config (severity + path)**
- `ignore-paths` and `severity` on the same rule cooperate correctly.

**Global `only-platforms`**
- Single-platform and multi-platform restrictions work; all-platforms equals no filter.

**Global `ignore-platforms`**
- Excluding one or multiple platforms leaves the rest unaffected.

**Combined platform + rule-config**
- Platform filter, `ignore-paths`, and severity override all work together correctly.

**Edge cases**
- Legacy UUID as `rule-configs` key matches the rule correctly.
- Unknown rule ID in `rule-configs` has no effect on other rules.
- Production vs. staging subdirectory path filtering tested independently for Kubernetes.

## How to test locally

```sh
# Unit tests
go test ./pkg/config/... ./pkg/engine/... ./cmd/scanner/... ./pkg/scan/... -count=1

# E2e tests (no credentials needed — uses local rule fixtures)
go test ./test/e2e/... -run TestV13_ -v -count=1
```

## Blast Radius

Config parsing, engine inspection, and CLI platform selection. No changes to SARIF output structure or Rego query evaluation. Existing v1.2 configs are unaffected.

## Additional Notes

`SupportedPlatforms` must stay in sync with the `--platforms` default in `datadog-iac-scanner-default-rules` whenever a new platform is released, until I find another solution to centralize.

### DIFF
--- a/cmd/scanner/scan.go
+++ b/cmd/scanner/scan.go
@@ -16,6 +16,7 @@ import (
 	"github.com/DataDog/datadog-iac-scanner/pkg/datadog"
 	"github.com/DataDog/datadog-iac-scanner/pkg/featureflags"
 	"github.com/DataDog/datadog-iac-scanner/pkg/model"
+	iacplatforms "github.com/DataDog/datadog-iac-scanner/pkg/platforms"
 	"github.com/DataDog/datadog-iac-scanner/pkg/scan"
 	git "github.com/go-git/go-git/v5"
 	cli "github.com/urfave/cli/v3"
@@ -74,7 +75,7 @@ var scanAction = &cli.Command{
 			Name:    "type",
 			Aliases: []string{"t"},
 			Usage:   "a list of platform types to scan",
-			Value:   constants.SupportedPlatforms,
+			Value:   iacplatforms.Supported,
 		},
 		// NOTE: --x-parallelparsing flag disabled due to pre-existing race conditions
 		// in concurrent query workers (SetupLogs shared state write, docker detector
@@ -379,7 +380,7 @@ func selectPlatforms(platforms []string) []string {
 		set[strings.ToLower(p)] = struct{}{}
 	}
 	var out []string
-	for _, p := range constants.SupportedPlatforms {
+	for _, p := range iacplatforms.Supported {
 		if _, found := set[strings.ToLower(p)]; found {
 			out = append(out, p)
 		}
--- a/cmd/scanner/scan.go
+++ b/cmd/scanner/scan.go
@@ -74,7 +74,7 @@ var scanAction = &cli.Command{
 			Name:    "type",
 			Aliases: []string{"t"},
 			Usage:   "a list of platform types to scan",
-			Value:   GetSupportedPlatforms(),
+			Value:   constants.SupportedPlatforms,
 		},
 		// NOTE: --x-parallelparsing flag disabled due to pre-existing race conditions
 		// in concurrent query workers (SetupLogs shared state write, docker detector
@@ -148,6 +148,23 @@ func runScan(ctx context.Context, c *cli.Command) error {
 	}
 	cfg.OnlyPaths = onlyPaths
 	cfg.IgnoreRules = append(c.StringSlice("exclude-queries"), cfg.IgnoreRules...)
+
+	activePlatforms := scan.ApplyPlatformFilters(c.StringSlice("type"), cfg.OnlyPlatforms, cfg.IgnorePlatforms)
+
+	for ruleID, rc := range cfg.RuleConfigs {
+		resolvedIgnore, pathErr := getRepoRelativePaths(repoDir, rc.IgnorePaths)
+		if pathErr != nil {
+			return errorWithExitCode(fmt.Errorf("invalid ignore-path in rule-config %q: %w", ruleID, pathErr), constants.InvalidConfigErrorCode)
+		}
+		resolvedOnly, pathErr := getRepoRelativePaths(repoDir, rc.OnlyPaths)
+		if pathErr != nil {
+			return errorWithExitCode(fmt.Errorf("invalid only-path in rule-config %q: %w", ruleID, pathErr), constants.InvalidConfigErrorCode)
+		}
+		rc.IgnorePaths = resolvedIgnore
+		rc.OnlyPaths = resolvedOnly
+		cfg.RuleConfigs[ruleID] = rc
+	}
+
 	params := &scan.Parameters{
 		CloudProvider:    []string{""},
 		OutputPath:       outputPath,
@@ -158,7 +175,7 @@ func runScan(ctx context.Context, c *cli.Command) error {
 		QueriesPath:      []string{"./assets/queries"},
 		LibrariesPath:    "./assets/libraries",
 		ReportFormats:    []string{"sarif"},
-		Platform:         selectPlatforms(c.StringSlice("type")),
+		Platform:         selectPlatforms(activePlatforms),
 		QueryExecTimeout: c.Int("timeout"),
 		DisableSecrets:   true,
 		ScanID:           "console",
@@ -173,7 +190,7 @@ func runScan(ctx context.Context, c *cli.Command) error {
 
 	metadata, err := console.ExecuteScan(ctx, params)
 	if err != nil {
-		return fmt.Errorf("error during IaC scan: %w", err)
+		return errorWithExitCode(fmt.Errorf("error during IaC scan: %w", err), constants.EngineErrorCode)
 	}
 
 	reportResult(repoDir, params.OutputPath, params.OutputName, &metadata)
@@ -362,7 +379,7 @@ func selectPlatforms(platforms []string) []string {
 		set[strings.ToLower(p)] = struct{}{}
 	}
 	var out []string
-	for _, p := range GetSupportedPlatforms() {
+	for _, p := range constants.SupportedPlatforms {
 		if _, found := set[strings.ToLower(p)]; found {
 			out = append(out, p)
 		}
--- a/internal/constants/constants.go
+++ b/internal/constants/constants.go
@@ -61,6 +61,10 @@ var (
 		"ServerlessFW":            "serverlessFW",
 	}
 
+	// SupportedPlatforms - Platforms with published rules, available to customers.
+	// Keep in sync with the default --platforms flag in datadog-iac-scanner-default-rules.
+	SupportedPlatforms = []string{"Ansible", "CICD", "CloudFormation", "Dockerfile", "Kubernetes", "Terraform"}
+
 	// AvailableSeverities - All severities available
 	AvailableSeverities = []string{
 		"critical",
--- a/internal/constants/constants.go
+++ b/internal/constants/constants.go
@@ -61,10 +61,6 @@ var (
 		"ServerlessFW":            "serverlessFW",
 	}
 
-	// SupportedPlatforms - Platforms with published rules, available to customers.
-	// Keep in sync with the default --platforms flag in datadog-iac-scanner-default-rules.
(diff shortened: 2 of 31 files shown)


internal/constants/constants.go
```diff
@@ -61,6 +61,10 @@ var (
 		"ServerlessFW":            "serverlessFW",
 	}
 
+	// SupportedPlatforms - Platforms with published rules, available to customers.
+	// Keep in sync with the default --platforms flag in datadog-iac-scanner-default-rules.
+	SupportedPlatforms = []string{"Ansible", "CICD", "CloudFormation", "Dockerfile", "Kubernetes", "Terraform"}
```
whitemerch: Not happy with this, but we should find a way to centralize these, in this pr or to follow up

cmd/scanner/scan.go
```diff
@@ -148,6 +148,23 @@ func runScan(ctx context.Context, c *cli.Command) error {
 	}
 	cfg.OnlyPaths = onlyPaths
 	cfg.IgnoreRules = append(c.StringSlice("exclude-queries"), cfg.IgnoreRules...)
+
+	activePlatforms := scan.ApplyPlatformFilters(c.StringSlice("type"), cfg.OnlyPlatforms, cfg.IgnorePlatforms)
```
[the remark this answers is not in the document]
jacobotb: Yeah, do the platform filtering inside the scan, like we do the path filtering inside the scan.

pkg/config/config.go
```diff
@@ -34,6 +35,15 @@ type iacGlobalConfig struct {
 	OnlySeverities   []iacSeverity `yaml:"only-severities,omitempty"`
 	IgnoreCategories []iacCategory `yaml:"ignore-categories,omitempty"`
 	OnlyCategories   []iacCategory `yaml:"only-categories,omitempty"`
+	OnlyPlatforms    []iacPlatform `yaml:"only-platforms,omitempty"`
+	IgnorePlatforms  []iacPlatform `yaml:"ignore-platforms,omitempty"`
```
jacobotb:
```suggestion
	IgnorePlatforms  []iacPlatform `yaml:"ignore-platforms,omitempty"`
	OnlyPlatforms    []iacPlatform `yaml:"only-platforms,omitempty"`
```

pkg/config/config.go
```diff
@@ -34,6 +35,15 @@ type iacGlobalConfig struct {
...
+}
+
+// iacRuleConfigYaml is the YAML representation of a per-rule override block.
+type iacRuleConfigYaml struct {
+	IgnorePaths []string    `yaml:"ignore-paths,omitempty"`
+	OnlyPaths   []string    `yaml:"only-paths,omitempty"`
+	Severity    iacSeverity `yaml:"severity,omitempty"`
```
jacobotb: Make it a pointer because an empty string is not a valid `iacSeverity`
```suggestion
	Severity    *iacSeverity `yaml:"severity,omitempty"`
```

pkg/config/types.go
```diff
@@ -11,4 +11,16 @@ type IacConfig struct {
...
+	IgnorePlatforms []string
+}
+
+// IacRuleConfig holds per-rule overrides: path scoping and severity override.
+// nil/empty slices mean "no restriction". An empty Severity means "no override".
+type IacRuleConfig struct {
+	IgnorePaths []string
+	OnlyPaths   []string
+	Severity    string // empty means no override
 }
```
jacobotb, line 26: In Go we use pointers for optionals.
```suggestion
// IacRuleConfig holds per-rule overrides: path scoping and severity override.
type IacRuleConfig struct {
	IgnorePaths []string
	OnlyPaths   []string
	Severity    *string
}
```

pkg/config/config.go
```diff
@@ -150,6 +200,8 @@ type schemaVersion struct {
 var (
 	// minIacVersion is the minimum schema version supporting the iac: configuration section.
 	minIacVersion = schemaVersion{1, 2}
+	// minV13Version is the schema version that introduced per-rule configs and platform filters.
+	minV13Version = schemaVersion{1, 3}
```
jacobotb:
```suggestion
	// currentVersion is the current schema version.
	currentVersion = schemaVersion{1, 3}
```

pkg/config/config.go
```diff
@@ -100,15 +143,22 @@ func UnparseConfig(cfg *IacConfig) ([]byte, error) {
...
 		return []byte{}, nil
 	}
 
+	schemaVer := minIacVersion.String()
+	if len(cfg.RuleConfigs) > 0 || len(cfg.OnlyPlatforms) > 0 || len(cfg.IgnorePlatforms) > 0 {
+		schemaVer = minV13Version.String()
+	}
 	outCfg := cfgFileYaml{
-		SchemaVersion: minIacVersion.String(),
+		SchemaVersion: schemaVer,
```
jacobotb:
```suggestion
		SchemaVersion: currentVersion,
```

pkg/config/config.go
```diff
@@ -100,15 +143,22 @@ func UnparseConfig(cfg *IacConfig) ([]byte, error) {
...
 	}
 
 	if reflect.DeepEqual(iac, iacConfig{}) {
 		return []byte{}, nil
 	}
 
+	schemaVer := minIacVersion.String()
+	if len(cfg.RuleConfigs) > 0 || len(cfg.OnlyPlatforms) > 0 || len(cfg.IgnorePlatforms) > 0 {
+		schemaVer = minV13Version.String()
+	}
```
jacobotb: Just use the current version.
```suggestion
```

pkg/engine/inspector.go
```diff
@@ -613,6 +631,96 @@ func getVulnerabilitiesFromQuery(ctx context.Context, qCtx *QueryContext, c *Ins
...
+	}
+	if len(onlyPaths) > 0 {
+		for _, pattern := range onlyPaths {
+			if matchesPath(pattern, filePath) {
+				return false
+			}
+		}
+		return true
+	}
+	return false
```
jacobotb: Purely stylistic note: here I would avoid one indentation level and three 'return' statements in a row by doing this:


```suggestion
	if len(onlyPaths) == 0 {
	    return false
	}
	for _, pattern := range onlyPaths {
		if matchesPath(pattern, filePath) {
			return false
		}
	}
	return true
```

pkg/engine/inspector.go
```diff
@@ -613,6 +631,96 @@ func getVulnerabilitiesFromQuery(ctx context.Context, qCtx *QueryContext, c *Ins
...
+}
+
+// matchesPath reports whether filePath is covered by pattern.
+// pattern may be a literal path, a directory prefix, a single-star glob (*, ?, [...]),
+// or a double-star glob (**) where ** matches zero or more path segments.
+// Both arguments are normalized to forward slashes for OS-independent behavior.
+func matchesPath(pattern, filePath string) bool {
```
jacobotb: Move `matchesPath` and `matchDoublestar` to a common or utility package and export `MatchesPath`.

pkg/report/model/sarif.go
```diff
@@ -21,7 +21,8 @@ import (
 	"github.com/google/uuid"
 )
 
-var severityLevelEquivalence = map[model.Severity]string{
+// SeverityLevelEquivalence maps a finding severity to its SARIF result level.
```
jacobotb, line 24: Maybe you can revert this file entirely; I don't see `SeverityLevelEquivalence` used elsewhere.

pkg/config/types.go
whitemerch: That's the side effect of doing too much typescript haha

pkg/report/model/sarif.go
whitemerch: Yes oversight from a previous attempt

whitemerch: Ah no, I forgot, it is definitely using it in test/e2e/v13_test.go

pkg/engine/inspector_test.go
```diff
@@ -1020,3 +1022,93 @@ func TestInspector_checkComment(t *testing.T) {
 		})
 	}
 }
+
+func TestMatchesPath(t *testing.T) {
```
jacobotb: Move this test to pathutil

pkg/config/config.go
```diff
@@ -101,10 +101,15 @@ func parseRuleConfigs(in map[string]iacRuleConfigYaml) map[string]IacRuleConfig
...
+		var sev *string
+		if rc.Severity != nil {
+			s := string(*rc.Severity)
+			sev = &s
+		}
 		out[ruleID] = IacRuleConfig{
 			IgnorePaths: rc.IgnorePaths,
 			OnlyPaths:   rc.OnlyPaths,
-			Severity:    string(rc.Severity),
+			Severity:    sev,
```
jacobotb: I think you can use `(*string)(rc.Severity)` to cast the `*iacSeverity` pointer to `*string` directly and avoid the block above.

pkg/config/config.go
```diff
@@ -116,10 +121,15 @@ func unparseRuleConfigs(in map[string]IacRuleConfig) map[string]iacRuleConfigYam
...
+		var sev *iacSeverity
+		if rc.Severity != nil {
+			s := iacSeverity(*rc.Severity)
+			sev = &s
+		}
 		out[ruleID] = iacRuleConfigYaml{
 			IgnorePaths: rc.IgnorePaths,
 			OnlyPaths:   rc.OnlyPaths,
-			Severity:    iacSeverity(rc.Severity),
+			Severity:    sev,
```
jacobotb: Same but `(*iacSeverity)(rc.Severity)`

jacobotb (approved):

PR #167, opened 2026-06-01
[K9CODESEC-2075] Include Helm chart CRD manifests in scan coverage

whitemerch:
## Motivation

Helm charts can ship CustomResourceDefinitions under `crds/`, but the scanner did not analyze them on a successful chart render. The dry-run client omitted CRDs from rendered output, and the filesystem walk skipped the chart subtree after render.

Related ticket: [K9CODESEC-2075](https://datadoghq.atlassian.net/browse/K9CODESEC-2075)

## Changes

**Helm resolver**

Enable `IncludeCRDs` so CRD manifests are part of the rendered output. Stamp YAML CRD files with `KICS_HELM_ID` markers for line mapping, include CRD files in source mapping (including subcharts), and map rendered CRD splits back to their originals. Multi-document CRD files are attributed via inline Helm markers and by carrying forward the last known `# Source:` when Helm omits it on later documents. Unknown source headers are dropped, nested `crds/.../crds/...` paths stay distinct, and chart source paths are normalized for cross-platform lookup.

**Runner and detection**

Helm CRD line info is reconstructed lazily from the stamped original instead of retaining rendered copies. Helm YAML is routed only to the matching parser service to avoid duplicate parsing across Kubernetes/CICD YAML handlers. Helm documents are prepared in place where safe, and compact `HelmIDLineRange` metadata replaces large per-line `IDInfo` maps. Marker IDs use original source line indices so multi-document templates and CRDs report correct line numbers.

**JSON CRDs**

JSON CRD files are included in scan output but are not stamped (YAML comments would corrupt JSON). They are parsed through the YAML parser while preserving their CRD source path, with degraded line attribution when markers are absent.

## QA Instruction

1. `go test ./pkg/resolver/helm/... ./pkg/runner/... ./pkg/detector/helm/...`
2. `make build && ./bin/datadog-iac-scanner scan -p test/fixtures/test_helm_with_crds -t Kubernetes -o /tmp/helm-crd-out` — CRD resources from `crds/` should appear in SARIF with correct paths; no duplicate raw-template findings from `templates/`.
3. Scan `test/fixtures/test_helm_subchart` and confirm the subchart CRD under `charts/subchart/crds/` is resolved.
4. Confirm nested, multi-document, and JSON CRD fixtures under `test/fixtures/test_helm_with_crds/` are all present in scan output.

## Impact

Affects CLI scans of repositories that include Helm charts with `crds/`. YAML CRDs gain full scan coverage and line mapping; JSON CRDs are scanned with degraded line precision. CRDs reach the engine with correct `_path` values even though the bundled public rule set does not currently emit CRD-specific findings by default. Peak memory on the Helm regression corpus is at or below `main`. No changes to rule definitions or SARIF schema.

[K9CODESEC-2075]: https://datadoghq.atlassian.net/browse/K9CODESEC-2075?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/detector/helm/helm_detect.go
+++ b/pkg/detector/helm/helm_detect.go
@@ -101,14 +101,15 @@ func (d DetectKindLine) DetectLine(ctx context.Context, file *model.FileMetadata
 		}
 		// Update found line
 		curLineRes.lineRes = removeLines(curLineRes.lineRes, lineRemove)
+		adjustedLine := curLineRes.lineRes + 1
 		return model.VulnerabilityLines{
-			Line:                  curLineRes.lineRes + 1,
+			Line:                  adjustedLine,
 			VulnLines:             detector.GetAdjacentVulnLines(curLineRes.lineRes, outputLines, lines),
 			LineWithVulnerability: strings.Split(lines[curLineRes.lineRes], ": ")[0],
 			ResolvedFile:          file.FilePath,
 			VulnerablilityLocation: model.ResourceLocation{
-				Start: start,
-				End:   end,
+				Start: model.ResourceLine{Line: adjustedLine, Col: start.Col},
+				End:   model.ResourceLine{Line: adjustedLine, Col: end.Col},
 			},
 		}
 	}
@@ -213,10 +214,24 @@ func detectLastSingle(line int, dis map[int]int, idInfo map[int]interface{}, id 
 	if idInfo == nil {
 		return true
 	}
+	var containsLine func(int) bool
+	switch originalLines := idInfo[id].(type) {
+	case model.HelmIDLineRange:
+		containsLine = func(line int) bool {
+			return line >= originalLines.Start && line <= originalLines.End
+		}
+	case map[int]int:
+		containsLine = func(line int) bool {
+			_, ok := originalLines[line]
+			return ok
+		}
+	default:
+		return true
+	}
 	for key, value := range dis {
 		if value == dis[line] && key != line {
 			// check if we are only looking at original data equivalent to the vulnerability
-			if ok := idInfo[id].(map[int]int)[key]; ok != 0 {
+			if containsLine(key) {
 				return false
 			}
 		}
--- a/pkg/detector/helm/helm_detect.go
+++ b/pkg/detector/helm/helm_detect.go
@@ -45,7 +45,9 @@ const (
 func (d DetectKindLine) DetectLine(ctx context.Context, file *model.FileMetadata, searchKey string,
 	outputLines int) model.VulnerabilityLines {
 	contextLogger := logger.FromContext(ctx)
-	searchKey = fmt.Sprintf("%s.%s", strings.TrimRight(strings.TrimLeft(file.HelmID, "# "), ":"), searchKey)
+	if file.HelmID != "" {
+		searchKey = fmt.Sprintf("%s.%s", strings.TrimRight(strings.TrimLeft(file.HelmID, "# "), ":"), searchKey)
+	}
 
 	lines := make([]string, len(*file.LinesOriginalData))
 	copy(lines, *file.LinesOriginalData)
@@ -142,6 +144,14 @@ func removeLines(current int, lineRemove map[int]int) int {
 	return current
 }
 
+func containsHelmKey(line, key string) bool {
+	if strings.Contains(line, key) {
+		return true
+	}
+	key = strings.TrimSuffix(key, ":")
+	return strings.Contains(line, `"`+key+`":`) || strings.Contains(line, `'`+key+`':`)
+}
+
 // nolint:gocritic
 func (d detectCurlLine) detectCurrentLine(lines []string, str1,
 	str2 string, byKey bool, idInfo map[int]interface{}, id int) (detectCurlLine, model.ResourceLine, model.ResourceLine) {
@@ -155,7 +165,7 @@ func (d detectCurlLine) detectCurrentLine(lines []string, str1,
 				ends[i] = model.ResourceLine{Line: i + 1, Col: len(lines[i])}
 			}
 		} else if str1 != "" {
-			if strings.Contains(lines[i], str1) {
+			if containsHelmKey(lines[i], str1) {
 				distances[i] = levenshtein.ComputeDistance(
 					detector.ExtractLineFragment(strings.TrimSpace(lines[i]), str1, byKey), str1)
 				starts[i] = model.ResourceLine{Line: i + 1, Col: 0}
--- a/pkg/resolver/helm/helm.go
+++ b/pkg/resolver/helm/helm.go
@@ -1,6 +1,7 @@
 package helm
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"io"
@@ -15,6 +16,7 @@ import (
 
 	"github.com/DataDog/datadog-iac-scanner/pkg/logger"
 	"github.com/pkg/errors"
+	"gopkg.in/yaml.v3"
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/chart"
 	"helm.sh/helm/v3/pkg/chart/loader"
@@ -35,6 +37,12 @@ const (
 	maxCandidateMinor = 40
 	minCandidateMinor = 0
 	maxCandidatePatch = 30
+
+	crdDirPrefix = "crds/"
+	extYAML      = ".yaml"
+	extYML       = ".yml"
+	extJSON      = ".json"
+	crdDirName   = "crds"
 )
 
 var (
@@ -91,7 +99,7 @@ func chartKubeVersionConstraint(ch *chart.Chart) string {
 }
 
 func runInstall(ctx context.Context, args []string, client *action.Install,
-	valueOpts *values.Options) (*release.Release, []string, error) {
+	valueOpts *values.Options) (*release.Release, *chart.Chart, []string, error) {
 	contextLogger := logger.FromContext(ctx)
 	log.SetOutput(io.Discard)
 	defer log.SetOutput(os.Stderr)
@@ -105,21 +113,21 @@ func runInstall(ctx context.Context, args []string, client *action.Install,
 
 	name, charts, err := client.NameAndChart(args)
 	if err != nil {
-		return nil, []string{}, err
+		return nil, nil, []string{}, err
 	}
 	contextLogger.Debug().Msgf("Parsed chart name: '%s', chart path: '%s'", name, charts)
 	client.ReleaseName = name
 
 	cp, err := client.LocateChart(charts, settings)
 	if err != nil {
-		return nil, []string{}, err
+		return nil, nil, []string{}, err
 	}
 	contextLogger.Debug().Msgf("Located chart at path: '%s'", cp)
 
 	p := getter.All(settings)
 	vals, err := valueOpts.MergeValues(p)
 	if err != nil {
-		return nil, []string{}, err
+		return nil, nil, []string{}, err
 	}
 	contextLogger.Debug().Msgf("Merged helm values successfully, values count: %d", len(vals))
 
@@ -127,7 +135,7 @@ func runInstall(ctx context.Context, args []string, client *action.Install,
 	contextLogger.Debug().Msgf("Loading chart from path: '%s'", cp)
 	chartRequested, err := loader.Load(cp)
 	if err != nil {
-		return nil, []string{}, err
+		return nil, nil, []string{}, err
 	}
 
 	// Set KubeVersion; clear the constraint only when unsatisfiable.
@@ -143,7 +151,7 @@ func runInstall(ctx context.Context, args []string, client *action.Install,
 	chartRequested = setID(chartRequested)
 
 	if instErr := checkIfInstallable(chartRequested); instErr != nil {
-		return nil, []string{}, instErr
+		return nil, nil, []string{}, instErr
 	}
 	contextLogger.Debug().Msg("Chart installability check passed")
 
@@ -151,12 +159,12 @@ func runInstall(ctx context.Context, args []string, client *action.Install,
 	contextLogger.Debug().Msgf("Running helm chart with namespace: '%s', release name: '%s'", client.Namespace, client.ReleaseName)
 	helmRelease, err := client.Run(chartRequested, vals)
 	if err != nil {
-		return nil, []string{}, err
+		return nil, nil, []string{}, err
 	}
 
 	contextLogger.Debug().Msgf("Successfully rendered helm chart '%s', manifest length: %d bytes",
 		chartRequested.Metadata.Name, len(helmRelease.Manifest))
-	return helmRelease, excluded, nil
+	return helmRelease, chartRequested, excluded, nil
 }
 
 // checkIfInstallable validates if a chart can be installed
@@ -182,7 +190,7 @@ func newClient(ctx context.Context) *action.Install {
 	client.Replace = true // Skip the name check
 	client.ClientOnly = true
 	client.APIVersions = chartutil.VersionSet([]string{})
-	client.IncludeCRDs = false
+	client.IncludeCRDs = true
 
 	contextLogger.Debug().Msgf("Configured helm client - DryRun: %t, ClientOnly: %t, IncludeCRDs: %t, ReleaseName: '%s'",
 		client.DryRun, client.ClientOnly, client.IncludeCRDs, client.ReleaseName)
@@ -404,6 +412,12 @@ func setID(chartReq *chart.Chart) *chart.Chart {
 			continue
 		}
 	}
+	// Stamp YAML CRDs for line mapping; JSON CRDs are skipped (YAML comments corrupt JSON).
+	for _, f := range localCRDFiles(chartReq) {
+		if isYAMLCRD(f.Name) {
+			addID(f)
+		}
+	}
 	for _, dep := range chartReq.Dependencies() {
 		dep = setID(dep)
 		if dep != nil {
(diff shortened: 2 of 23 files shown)


pkg/resolver/helm/resolver.go
```diff
@@ -44,22 +43,23 @@ func (r *Resolver) Resolve(ctx context.Context, filePath string) (model.Resolved
...
-		origpath := filepath.Join(filepath.Dir(filePath), subFolder, splited)
+		sourceKey := chartSourceKey(split.path)
+		seenCRDs[sourceKey] = struct{}{}
+		chartRelative, ok := chartRelativeFromSource(sourceKey)
+		if !ok {
+			continue
+		}
+		origpath := resolvedChartFilePath(filePath, chartRelative)
 		rfiles.File = append(rfiles.File, model.ResolvedHelm{
 			FileName:     origpath,
```
robertohuertasm-datadog: Thanks for adding CRD coverage. I think there is a regression here that is worth fixing before merge.

Helm emits CRDs first, so on `test/fixtures/test_helm_with_crds` this `FileName` is still `crds/gadget.json`. Downstream, `storeResolvedFiles` in `pkg/runner/resolver_sink.go` hands that list to the YAML parser used for Kubernetes scans.

That parser returns `kind="break"` for the unsupported `.json` extension. In `storeResolvedFiles` (around the `if documents.Kind == "break"` check), the loop then `return`s, so none of the remaining files are stored. The YAML CRDs and `templates/service.yaml` never get scanned. A successful render also excludes the chart from the filesystem walk, so they are not picked up as raw files either.

On main, templates in a chart that contains a JSON CRD still scan. On this branch they disappear. The QA path `scan -p test/fixtures/test_helm_with_crds -t Kubernetes` would not show the CRDs or the Service.

A couple of related gaps:

- Kubernetes scans do not attach the JSON parser (`ShouldScanTfPlans` only), so JSON CRDs would not produce Kubernetes findings as written.
- Helm content is `# Source: ...` plus JSON. That is valid YAML, but not valid JSON.
- Resolver tests never go through `storeResolvedFiles`, so they stay green.

Could we:

1. In `pkg/runner/resolver_sink.go` (`storeResolvedFiles`), `continue` on `documents.Kind == "break"` instead of `return`, so one unsupported extension does not drop sibling files.
2. Parse JSON CRD content with the YAML parser rather than relying on `.json` extension dispatch, if we want those CRDs as Kubernetes findings.
3. Add a test that goes through `storeResolvedFiles` (or a real scan) on `test_helm_with_crds` and asserts the Service and YAML CRDs are stored even when `gadget.json` is first.

Until (1) lands, I would omit `.json` files from this resolved list. Shipping them as-is is a coverage regression for templates that used to scan.

pkg/resolver/helm/helm.go
```diff
@@ -429,6 +441,90 @@ func addID(file *chart.File) *chart.File {
...
+	name = normalizeChartPath(name)
+	ext := strings.ToLower(filepath.Ext(name))
+	if ext != extYAML && ext != extYML && ext != extJSON {
+		return false
+	}
+	parts := strings.Split(name, "/")
+	if len(parts) < 2 {
+		return false
+	}
+	return parts[len(parts)-2] == crdDirName
```
robertohuertasm-datadog: `isCRDManifest` only treats a file as a CRD when its immediate parent directory is `crds/`. Helm’s own discovery (`Chart.CRDs()` / `CRDObjects()`) is `strings.HasPrefix(f.Name, "crds/")` plus a yaml/yml/json extension, and Helm’s tests include nested paths like `crds/foo/bar/baz.yaml`.

Those nested files still get into the rendered manifest via `IncludeCRDs`, then `splitManifestYAML` drops them because they are missing from `origData`. `appendDirectCRDFiles` uses the same predicate, so it cannot recover them either.

Please key off `crdChartRelativePath` (already used for Windows/subchart names) and treat a file as a CRD when that relative path has prefix `crds/` and a yaml/yml/json extension, rather than requiring `parts[len(parts)-2] == "crds"`.

pkg/resolver/helm/resolver.go
```diff
@@ -78,83 +83,275 @@ func (r *Resolver) SupportedTypes() []model.FileKind {
...
-		path := strings.Split(strings.TrimPrefix(splited, "\n# Source: "), "\n") // get source of split yaml
-		// ignore auxiliary files used to render chart
-		if path[0] == "" {
-			continue
+
+		sourcePath, hasSource := parseManifestSource(splited)
+		if hasSource {
+			sourcePath = chartSourceKey(sourcePath)
+			if origData[sourcePath] == nil {
+				hasSource = false
```
robertohuertasm-datadog: Nice work on the source-header handling here! I did want to raise one thing before we merge, because I think this line has a wider effect than it looks.

When a `# Source:` header *is* present but isn't found in `origData`, we downgrade it to "sourceless" and fall through to the `lastSource` branch below. Before this PR that split was dropped, so I'm wondering whether we should keep that behaviour? As written, the split gets attributed to whatever file happened to come before it, which means `content` and `original` end up coming from two different files — and since `OriginalData` drives the Helm detector's line mapping, the finding is reported on the wrong file at the wrong line.

There's also a knock-on effect I hit while testing. Once a split is attributed to a foreign source, `crdSplitCount` can outrun `markers` in the block just above (line 157), leaving `SplitID` empty while `original` *is* stamped. That combination means `getIDMap` never emits a `-1` key, `DetectLine` resolves `helmID` to `-1`, and `detectLastSingle` panics on `idInfo[id].(map[int]int)`.

I managed to reproduce it end to end with a chart shaped like this:

- `crds/aaa.yaml` — one Service with `metadata.name` and `spec.selector.name`
- `crds/zsub/crds/aaa.yaml` — two Services of the same shape, separated by `---`
- `templates/svc.yaml` — any Service

`scan -t Kubernetes` on that chart logs `Recovered from panic during query 'using_unrecommended_namespace'` and `'service_does_not_target_pod'` and reports 1 violation. Because the engine's `doRun` recover is per-query across the whole scan, one malformed chart silences those rules for *every* file in the repo. After changing this line to `continue` (plus the `crdChartRelativePath` tweak I commented on separately), the same scan reports 7 violations with no panics.

Should we just bail out instead? The genuinely header-less multi-document CRD path below is untouched by this, since those splits have no header at all:

```suggestion
				continue
```

Heads-up that with this change `TestSplitManifestYAML_userSourceCommentInLaterDocument` fails, since it currently asserts the misattribution as intended — I think that one wants inverting to assert the split is dropped. Everything else in `pkg/resolver/helm`, `pkg/runner` and `pkg/detector/helm` stayed green for me. What do you think?

pkg/resolver/helm/helm.go
```diff
@@ -414,21 +428,128 @@ func setID(chartReq *chart.Chart) *chart.Chart {
...
+	if ext != extYAML && ext != extYML {
+		return false
+	}
+	return strings.HasPrefix(name, crdDirPrefix)
+}
+
+// crdChartRelativePath returns the chart-relative crds/ path for a CRD file name.
+func crdChartRelativePath(name string) string {
+	name = normalizeChartPath(name)
+	if idx := strings.Index(name, "/crds/"); idx >= 0 {
```
robertohuertasm-datadog: Small ordering question on this helper. Since `isCRDManifest` already guarantees the name starts with `crds/`, every name reaching here is chart-relative and the `HasPrefix` branch below is the one we actually want. With `/crds/` checked first, a nested CRD directory gets rewritten: `crds/zsub/crds/aaa.yaml` becomes `crds/aaa.yaml`.

That collision has two effects I noticed. `localCRDFiles` dedupes on this value, so two genuinely distinct files collapse into one and whichever loses is never stamped by `setID`. And the `origData` key stops matching the `# Source:` value Helm emits (`<chart>/crds/zsub/crds/aaa.yaml`), which drops the split into the `lastSource` fallback I commented on in `resolver.go`.

Should we swap the two checks so the prefix case wins first? Or, since `f.Name` from `ch.Files` is already chart-relative, would it be simpler to drop the rewrite here altogether and use `f.Name` directly — that would keep our keys identical to Helm's by construction rather than by coincidence. Happy either way, just flagging that the current order is the bit that lets the keys diverge.

pkg/detector/helm/helm_detect.go
```diff
@@ -45,7 +45,9 @@ const (
 func (d DetectKindLine) DetectLine(ctx context.Context, file *model.FileMetadata, searchKey string,
 	outputLines int) model.VulnerabilityLines {
 	contextLogger := logger.FromContext(ctx)
-	searchKey = fmt.Sprintf("%s.%s", strings.TrimRight(strings.TrimLeft(file.HelmID, "# "), ":"), searchKey)
+	if file.HelmID != "" {
```
robertohuertasm-datadog, line 48: This guard is a good catch, and it made me look at what happens downstream once `HelmID` can legitimately be empty.

When it is, `helmID` resolves to `-1` a few lines below, and `detectLastSingle` then does an unchecked `idInfo[id].(map[int]int)`. That's safe for JSON CRDs only because `getIDMap` on unstamped content happens to emit a `-1` key — for any *stamped* original, `IDInfo` has keys like `{0, 16}` and no `-1`, so the assertion panics on a nil interface. I hit this for real via the resolver path (details in my `resolver.go` comment), and because the engine recovers per query, the effect is that the whole rule silently returns zero results for the entire scan.

I know `detectLastSingle` isn't in this diff, but since this line is what makes the empty case reachable, would it be worth hardening it in the same PR? Something like:

```go
m, ok := idInfo[id].(map[int]int)
if !ok {
    continue
}
if m[key] != 0 {
    return false
}
```

That way the invariant "empty `HelmID` implies `IDInfo` has a `-1` key" doesn't have to hold implicitly. Would also make a nice belt-and-braces alongside the resolver fix. What do you reckon?

pkg/resolver/helm/resolver.go
```diff
@@ -78,83 +83,275 @@ func (r *Resolver) SupportedTypes() []model.FileKind {
...
+		part = strings.TrimSpace(part)
+		if part == "" {
+			continue
+		}
+		docs = append(docs, []byte(part))
+	}
+	return docs
+}
+
+func appendDirectCRDFiles(
```
robertohuertasm-datadog: I instrumented both new fixtures out of curiosity, and this function contributes **0 files** in each. I think it may be unreachable rather than just unused in practice.

The reason looks structural: Helm's `renderResources` emits every `crds/` manifest with a `# Source: <ChartFullPath>/<f.Name>` header, and `updateName` registers those same files under keys built the same way, so `seen` already contains all of them by the time we get here.

Related, and this is the part that surprised me most: the `markersBySource` / `crdSplitCount` "Nth marker for the Nth document" logic at line 157 also never drives multi-document attribution. `setID` stamps the very `*chart.File` pointers that `CRDObjects()` renders from, so each YAML CRD document arrives with its own inline marker already. The rendered `crds/multi.yaml` is literally:

```yaml
# Source: test_helm_with_crds/crds/multi.yaml
# KICS_HELM_ID_0:
apiVersion: apiextensions.k8s.io/v1
...
---
# KICS_HELM_ID_16:
apiVersion: apiextensions.k8s.io/v1
```

So `TestHelm_Resolve_MultiDocCRD` passes through the inline-marker path, which means it's green regardless of whether this fallback is correct. The `lastSource` attribution genuinely is needed (document 2 really has no header), but the marker-counting isn't — and as I mentioned in my line-133 comment, the counter is keyed on marker-less splits rather than document index, so the one case where it does fire hands out the wrong marker.

Should we drop `appendDirectCRDFiles`, `crdDocuments`, and the marker-count fallback? It'd remove roughly 90 lines and one source of wrong-line bugs. Would also be worth trimming the two matching paragraphs from the PR description, since they describe this as the multi-document mechanism. Totally understand if it was intended as a safety net, just wanted to share what I measured.

pkg/resolver/helm/resolver_test.go
```diff
@@ -195,15 +539,35 @@ spec:
 	ctx := context.Background()
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			got, err := res.Resolve(ctx, tt.args.filePath)
+			filePath := tt.args.filePath
+			if tt.name == "test_with_dependencies" {
```
robertohuertasm-datadog: Minor readability thought on the table here. Branching on `tt.name` inside the loop means this case ignores its own `args.filePath` and runs a completely different set of assertions from every other row, so the table no longer really describes what the test does; and there's a second `tt.name` check further down at line 550 doing the same.

Would it be worth pushing the difference into the case struct instead, something like a `chartFixture string` plus a `checkSubchartCRD bool`? That way the table stays the single source of truth and the loop body stays uniform. Very much a nice-to-have, not a blocker.

pkg/resolver/helm/helm.go
```diff
@@ -182,7 +190,7 @@ func newClient(ctx context.Context) *action.Install {
 	client.Replace = true // Skip the name check
 	client.ClientOnly = true
 	client.APIVersions = chartutil.VersionSet([]string{})
-	client.IncludeCRDs = false
+	client.IncludeCRDs = true
```
robertohuertasm-datadog, line 193: Not a problem with the change, just something I'd love to understand better so I can review the value side properly.

I fetched the live bundle (1397 rules, 142 Kubernetes) and couldn't find any rule that references `CustomResourceDefinition`. `using_unrecommended_namespace` has an explicit kind allowlist that excludes it, which is why scanning the new fixture produced no CRD findings even after I added `namespace: default` to both the YAML and JSON CRDs. And the 44 kind-agnostic Kubernetes rules all turned out to be control-plane/kubelet flag checks.

The plumbing itself definitely works: I confirmed all five CRD documents reach the engine with correct `_path` values. So I'm guessing the payoff is custom rulesets or rules that are still coming? Would it be worth a line in the PR description saying so, just so reviewers don't expect new findings from this change on its own?

pkg/resolver/helm/resolver.go
whitemerch: Yes they shouldn't. Im dropping nknown source headers now and clear the previous source, so following documents can’t be misattributed

pkg/resolver/helm/helm.go
whitemerch: True, im making root-relative crds/ paths now take precedence

pkg/resolver/helm/resolver.go
```diff
@@ -78,83 +83,275 @@ func (r *Resolver) SupportedTypes() []model.FileKind {
...
-	sources = updateName(sources, template.Chart, template.Chart.Name())
+	sources = updateName(sources, sourceChart, sourceChart.Name())
 	var splitedManifest []splitManifest
-	splitedSource := strings.Split(template.Manifest, "---") // split manifest by '---'
+	splitedSource := splitHelmManifest(template.Manifest)
 	origData := toMap(sources)
+	// crdSplitCount and markersBySource support multi-document CRD files: Helm renders
+	// each document as a separate split without repeating the Source header, and omits
+	// KICS_HELM_ID markers. We attribute sourceless splits to lastSource and pick the
+	// Nth cached marker for the Nth occurrence of a given source path.
```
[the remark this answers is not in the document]
whitemerch: Removed

pkg/resolver/helm/helm.go
whitemerch: Yes, the immediate value is complete manifest coverage for custom and future rules; current built-in rules don’t produce CRD findings. I am mainly doing this as I noticed this gap, the fix is easy and have a complete scanner

pkg/resolver/helm/resolver_test.go
whitemerch: Indeed, they now live in the test case instead

pkg/resolver/helm/resolver.go
whitemerch: Yes it's just a remnant of the multi commits i did, so i removed it

pkg/detector/helm/helm_detect.go
whitemerch: yes, fixed!

pkg/resolver/helm/helm.go
```diff
@@ -414,21 +428,208 @@ func setID(chartReq *chart.Chart) *chart.Chart {
...
-			split[i] = fmt.Sprintf("# KICS_HELM_ID_%d:", i)
-			i++
+	apiVersionLines := topLevelAPIVersionLines(split, isYAMLCRD(file.Name))
+
+	stamped := make([]byte, 0, len(file.Data)+len(apiVersionLines)*24)
+	nextAPIVersion := 0
+	for i, line := range split {
+		if nextAPIVersion < len(apiVersionLines) && apiVersionLines[nextAPIVersion] == i {
+			stamped = append(stamped, "# KICS_HELM_ID_"...)
+			stamped = strconv.AppendInt(stamped, int64(i+nextAPIVersion), helmIDNumberBase)
```
robertohuertasm-datadog: **non-blocking**

The ID here is the marker's own index in the stamped output, which is what the previous implementation produced too, so this is not something you introduced. It does mean the line reported for the second and later documents of a multi-document file drifts by the number of markers already inserted above them, and CRD support makes that path a lot easier to reach.

I hit it on a 15 line chart file holding two Services split by `---`:

```
 5  spec:                 <- doc 1
 7      name: nested-one
 8  ---
 9  apiVersion: v1
13  spec:                 <- doc 2
15      name: nested-two
```

`scan -t Kubernetes` puts document 1 on lines 5 and 7, which is right, and document 2 on 14 and 16 instead of 13 and 15. Line 16 is past the end of the file. The drift is exactly the one marker sitting above document 2.

I checked whether this PR caused it before flagging it: building the merge base `44079c4e` and scanning a multi-document *template* with no CRDs anywhere gives the same 14 and 16. So the mapping has always ignored the marker offset, and this change only widens the surface to `crds/`.

Since it is pre-existing I would not hold the merge for it. Would you rather fix it here while the line mapping is fresh in your head, or split it out and keep this PR scoped to CRD coverage? Either is fine by me, I just did not want it to get lost now that CRDs make it easier to trip over.

whitemerch: I can fix it it's alright

robertohuertasm-datadog (approved):
LGTM! 🚀

PR #169, opened 2026-06-02
Fix false suppression of Helm scan findings

whitemerch:
## Motivation

When scanning Helm charts whose templates contain `{{ … }}` expressions (without a leading dash), the scanner's attempt to parse the original template for suppression directives fails. As a fallback, suppression line numbers are derived from the rendered YAML instead. The rendered YAML always begins with two scanner-injected header lines, `# Source: …` added by Helm, and `# KICS_HELM_ID_N:` added by the scanner, which the YAML parser treats as regular head comments and adds to the ignore-line list. The Helm line detector can resolve `vulnerability.Line` to those same positions (via a duplicate-attribute fallback), causing any finding whose line coincides with them to be silently marked as suppressed with justification `dd-iac-scan ignore`, even though no suppression directive was ever written by the user.

## Changes

**Scan pipeline: Helm resolver sink**

When `getOriginalIgnoreLines` fails, the ignore-line list from the rendered YAML parse is now filtered to remove entries whose corresponding line is a scanner-injected header (`# Source:` or `# KICS_HELM_ID_N:`). User-authored `dd-iac-scan ignore-line` and `ignore-block` directives do not match either prefix and are unaffected.

## Author Checklist

## QA Instruction

1. CI should pass
2. Scan a Helm chart that uses `{{ .Values.something }}` expressions and confirm that findings previously reported as suppressed (with justification `dd-iac-scan ignore`) are now reported as active when no suppression directive exists in the source.

## Blast Radius

Only affects the Helm scan path, and only for templates where `getOriginalIgnoreLines` fails (templates with non-dash `{{ }}` expressions). The change removes false suppression; it cannot introduce new false positives. Templates that already parse correctly via `getOriginalIgnoreLines` are untouched.

## Additional Notes

### DIFF
--- a/pkg/kics/resolver_sink_test.go
+++ b/pkg/kics/resolver_sink_test.go
@@ -1,0 +1,94 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache-2.0 License.
+ *
+ * This product includes software developed at Datadog (https://www.datadoghq.com)  Copyright 2024 Datadog, Inc.
+ */
+package kics
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestFilterHelmGeneratedLines(t *testing.T) {
+	// Rendered Helm split content that resembles real scanner output.
+	// Line 1: empty
+	// Line 2: # Source: … (Helm-generated)
+	// Line 3: # KICS_HELM_ID_2: (scanner-injected)
+	// Line 4: apiVersion: apps/v1
+	// Line 5: kind: Deployment
+	// Line 6: # some user comment
+	// Line 7: metadata:
+	renderedContent := []byte("\n# Source: dsm-demo/templates/deployment.yaml\n# KICS_HELM_ID_2:\napiVersion: apps/v1\nkind: Deployment\n# some user comment\nmetadata:\n")
+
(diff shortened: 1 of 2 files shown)


pkg/kics/resolver_sink_test.go
```diff
@@ -0,0 +1,94 @@
...
+		{
+			name:        "removes Source and KICS_HELM_ID lines",
+			content:     renderedContent,
+			ignoreLines: []int{2, 3, 6},
+			want:        []int{6},
+		},
+		{
+			name:        "keeps user comment lines untouched",
+			content:     renderedContent,
+			ignoreLines: []int{6, 7},
```
ChouraquiBen, line 40: You wanted me to react on this line xDDD

ChouraquiBen (approved):
LGTM

I don't get how the line ends up being resolved as the comment line in the end, though. I saw the duplicate attribute fallback but couldn't find an example where it would link to the comments lines.

whitemerch:
> LGTM
> 
> I don't get how the line ends up being resolved as the comment line in the end, though. I saw the duplicate attribute fallback but couldn't find an example where it would link to the comments lines.

@ChouraquiBen very obscure bug, that I can explain if you want afterwards

PR #172, opened 2026-06-02
Skip Helm library charts in resolver

whitemerch:
## Motivation

Helm library charts (`type: library`) cannot be installed; the scanner was still running the Helm dry-run path and logging errors such as `chart is not installable: library charts are not installable (only 'application' type charts are supported)`, which added noise to `service:iac-scanning` without improving coverage (library charts have no deployable manifests).

## Changes

**Resolver**

`GetType` reads `Chart.yaml` and returns `KindCOMMON` for `type: library`, so the Helm resolver is not invoked for those directories (same skip behavior as non-chart paths).

**Tests**

Fixture `test/fixtures/test_helm_library` and `TestGetType` cases for application vs library charts.

## Author Checklist

## QA Instruction

CI should pass.

## Blast Radius

Only the Helm classification path in `pkg/resolver`; library chart directories are no longer passed to the Helm renderer.

## Additional Notes

### DIFF
--- a/pkg/resolver/resolver.go
+++ b/pkg/resolver/resolver.go
@@ -7,11 +7,13 @@ package resolver
 
 import (
 	"context"
+	"io/fs"
 	"os"
 	"path/filepath"
 
 	"github.com/DataDog/datadog-iac-scanner/pkg/logger"
 	"github.com/DataDog/datadog-iac-scanner/pkg/model"
+	"gopkg.in/yaml.v3"
 )
 
 // kindResolver is a type of resolver interface (ex: helm resolver)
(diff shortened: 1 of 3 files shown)


pkg/resolver/resolver.go
```diff
@@ -79,9 +81,24 @@ func (r *Resolver) Resolve(ctx context.Context, filePath string, kind model.File
 
 // GetType will analyze the filepath to determine which resolver to use
 func (r *Resolver) GetType(filePath string) model.FileKind {
-	_, err := os.Stat(filepath.Join(filePath, "Chart.yaml"))
-	if err == nil {
-		return model.KindHELM
+	chartFS := os.DirFS(filepath.Clean(filePath))
+	data, err := fs.ReadFile(chartFS, "Chart.yaml")
```
ChouraquiBen, line 85: Question: I have rarely seen `fs.ReadFile` used, usually, it is `os.ReadFile`, is there a reason why using `fs.ReadFile` here?

ChouraquiBen (approved):
I don't think it is blocking but one question

pkg/resolver/resolver.go
whitemerch: I used os.ReadFile and it made go-ci fail from what I recall

PR #173, opened 2026-06-02
Bypass Chart kubeVersion for Helm scan renders

whitemerch:
## Motivation

Helm client-only dry-run uses a synthetic cluster, not the user’s real Kubernetes version. When `Chart.yaml` declares `kubeVersion`, Helm can refuse to render before any manifest exists. A too-old default failed charts with higher lower bounds, and picking a single “new” version from build metadata could still fail charts with upper bounds or other constraints. Those failures only reduced scan coverage without modeling a real install.

## Changes

**Helm resolver**

After loading a chart for scan-time rendering, clear `metadata.kubeVersion` so Helm does not gate rendering on that field. `newClient` sets `KubeVersion` to a fixed value (`v1.30.0`) so `.Capabilities.KubeVersion` in templates stays stable across builds and dependency bumps. Tests and fixtures cover a lower-bound chart and a deliberately unsatisfiable `kubeVersion` so we keep rendering for scanning.

## QA Instruction

1. `go test ./pkg/resolver/helm/...`
2. Optional: run a scan against a chart that declares `kubeVersion` (including upper-bound only) and confirm Helm still produces rendered manifests without a version incompatibility error.

## Impact

Helm dry-run rendering during scans only (no cluster API). Scan-time rendering no longer honors `Chart.yaml` `kubeVersion` for pass/fail; templates that branch on `.Capabilities.KubeVersion` still see the fixed dry-run version.

## Additional Notes

This contribution is licensed under Apache-2.0.

### DIFF
--- a/pkg/resolver/helm/helm.go
+++ b/pkg/resolver/helm/helm.go
@@ -49,22 +49,17 @@ func dryRunKubeVersion() *chartutil.KubeVersion {
 	return cachedKubeVersion
 }
 
-// resolveChartKubeVersion returns a version satisfying constraint; true means
-// nothing in range satisfies it (caller should drop the constraint).
-func resolveChartKubeVersion(constraint string) (*chartutil.KubeVersion, bool) {
+// resolveChartKubeVersion returns a version satisfying constraint; unsatisfiable
+// means nothing in range satisfies it (caller should drop the constraint).
+func resolveChartKubeVersion(constraint string) (kv *chartutil.KubeVersion, unsatisfiable bool) {
 	def := dryRunKubeVersion()
 	if constraint == "" || chartutil.IsCompatibleRange(constraint, def.Version) {
 		return def, false
 	}
 	defMinor, _ := strconv.Atoi(def.Minor)
 	// Walk minors outward from the default, probing all patch levels per minor.
-	for d := 0; d <= maxCandidateMinor; d++ {
-		var minors []int
-		if d == 0 {
-			minors = []int{defMinor}
-		} else {
-			minors = []int{defMinor - d, defMinor + d}
-		}
+	minors := []int{defMinor}
+	for d := 1; d <= maxCandidateMinor+1; d++ {
 		for _, minor := range minors {
 			if minor < minCandidateMinor || minor > maxCandidateMinor {
 				continue
@@ -81,6 +76,7 @@ func resolveChartKubeVersion(constraint string) (*chartutil.KubeVersion, bool) {
 				}
 			}
 		}
+		minors = []int{defMinor - d, defMinor + d}
 	}
 	return def, true
 }
--- a/pkg/resolver/helm/helm.go
+++ b/pkg/resolver/helm/helm.go
@@ -7,7 +7,9 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
+	"sync"
 
 	"github.com/DataDog/datadog-iac-scanner/pkg/logger"
 	"github.com/pkg/errors"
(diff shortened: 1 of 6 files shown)


pkg/resolver/helm/helm.go
```diff
@@ -23,10 +25,73 @@ import (
...
+const (
+	maxCandidateMinor = 40
+	minCandidateMinor = 0
+	maxCandidatePatch = 30
+)
+
 var (
 	settings = cli.New()
+
+	kubeVersionOnce   sync.Once
```
ChouraquiBen, line 41: Question: Does it mean that it will have to run with the same version for all the the files?

Imagine a repo with lots of helm charts maintained by different teams working on different versions of helm. Will the use of `Once` not prevent checking the different versions that can be found in the repo?

pkg/resolver/helm/helm.go
```diff
@@ -23,10 +25,73 @@ import (
...
+func dryRunKubeVersion() *chartutil.KubeVersion {
+	kubeVersionOnce.Do(func() {
+		cachedKubeVersion, _ = chartutil.ParseKubeVersion(defaultDryRunKubeVersion)
+	})
+	return cachedKubeVersion
+}
+
+// resolveChartKubeVersion returns a version satisfying constraint; true means
+// nothing in range satisfies it (caller should drop the constraint).
+func resolveChartKubeVersion(constraint string) (*chartutil.KubeVersion, bool) {
```
ChouraquiBen: good practice: with more than one return var, it is best to name them

ChouraquiBen: imho better than a comment 'cause it cannot be forgotten

pkg/resolver/helm/helm.go
ChouraquiBen: Okay, no: the version to test against will be computed once and then there will be comparisons against this

pkg/resolver/helm/helm.go
```diff
@@ -23,10 +25,73 @@ import (
...
+		return def, false
+	}
+	defMinor, _ := strconv.Atoi(def.Minor)
+	// Walk minors outward from the default, probing all patch levels per minor.
+	for d := 0; d <= maxCandidateMinor; d++ {
+		var minors []int
+		if d == 0 {
+			minors = []int{defMinor}
+		} else {
+			minors = []int{defMinor - d, defMinor + d}
```
ChouraquiBen: nit: I guess that the goal is to get the closest version and that is why it is looped this way. This could be a comment

pkg/resolver/helm/helm.go
```diff
@@ -23,10 +25,73 @@ import (
...
+					continue // already checked at the top
+				}
+				if chartutil.IsCompatibleRange(constraint, candidate) {
+					if kv, err := chartutil.ParseKubeVersion(candidate); err == nil {
+						return kv, false
+					}
+				}
+			}
+		}
+	}
```
ChouraquiBen, line 80: Remark: This feels very inefficient. There are many loops and condition checks for nothing. This at least gets rid of the check of the first iteration. Maybe we could do better without removing much readability. Maybe this is going too far over nothing

```suggestion
	// Walk minors outward from the default, probing all patch levels per minor.
	minors := []int{defMinor}
	for d := 1; d <= maxCandidateMinor+1; d++ {
		for _, minor := range minors {
			if minor < minCandidateMinor || minor > maxCandidateMinor {
				continue
			}
			for patch := 0; patch <= maxCandidatePatch; patch++ {
				candidate := fmt.Sprintf("v1.%d.%d", minor, patch)
				if candidate == def.Version {
					continue // already checked at the top
				}
				if chartutil.IsCompatibleRange(constraint, candidate) {
					if kv, err := chartutil.ParseKubeVersion(candidate); err == nil {
						return kv, false
					}
				}
			}
		}
		minors = []int{defMinor - d, defMinor + d}
	}
```

ChouraquiBen (dismissed):
A few comments but nothing blocking

ChouraquiBen (approved):
LGTM

PR #174, opened 2026-06-03
Rename Rego entry-point from package Cx/CxPolicy to package datadog/DatadogPolicy

whitemerch:
## Motivation

The scanner engine and its default rules were originally forked from KICS (Checkmarx), which used `package Cx` and `CxPolicy` as the Rego namespace and policy entry-point, `Cx` standing for Checkmarx. Before exposing Rego rules to users for customization, this naming should reflect Datadog branding rather than a third-party product.

## Changes

**Engine (pkg/utils, pkg/engine)**

Introduced a backward-compatible shim module (`dd_iac_compat`) that is injected alongside every rule at evaluation time. The shim unions results from both `data.Cx.CxPolicy` (legacy) and `data.datadog.DatadogPolicy` (new) using two incremental Rego rule definitions, so each evaluates to an empty set when the corresponding package is absent. `RegoQuery` now targets `data.dd_iac_compat.policy`. This means the engine and the rules repo can be deployed independently, there is no flag day.

## QA Instruction

1. `go test ./pkg/utils/... ./pkg/engine/... ./pkg/scan/...`
2. Run a full scan against any local IaC fixture and confirm findings are still produced.
3. To verify backward compatibility, temporarily revert `test/fixtures/all_auth_users_get_read_access/query.rego` to `package Cx` / `CxPolicy`, run `go test ./pkg/engine/...`, and confirm the test still passes.

## Blast Radius

Affects the OPA query evaluated for every rule. No change in scan output, findings from both naming conventions are collected and merged. The only observable effect is that the new `package datadog` / `DatadogPolicy` naming is now the canonical form for rule authors.

## Additional Notes

A companion PR renames all 1,770 production `query.rego` files. That PR should be merged after this one is live. Once migration is confirmed complete, the compat shim can be removed in a follow-up and `RegoQuery` simplified back to `result = data.datadog.DatadogPolicy`.

### DIFF
--- a/pkg/engine/inspector.go
+++ b/pkg/engine/inspector.go
@@ -775,6 +775,7 @@ func (q QueryLoader) LoadQuery(ctx context.Context, query *model.QueryMetadata,
 			rego.Module("Common", q.commonLibrary.LibraryCode),
 			rego.Module("Generic", platformGeneralQuery.LibraryCode),
 			rego.Module(query.Query, query.Content),
+			rego.Module("dd_iac_compat.rego", utils.RegoCompatShim),
 			rego.Store(store),
 			rego.UnsafeBuiltins(unsafeRegoFunctions),
 		).PrepareForEval(ctx)
--- a/pkg/engine/testdata/simid/query.rego
+++ b/pkg/engine/testdata/simid/query.rego
@@ -1,8 +1,8 @@
-package Cx
+package datadog
 
 # Minimal rule for TestInspectorSimilarityID: fires on every aws_instance resource.
 # Deliberately avoids library imports so it works with the stubQueriesSource library stub.
-CxPolicy[result] {
+DatadogPolicy[result] {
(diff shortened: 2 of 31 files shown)


pkg/engine/inspector.go
```diff
@@ -775,6 +775,7 @@ func (q QueryLoader) LoadQuery(ctx context.Context, query *model.QueryMetadata,
 			rego.Module("Common", q.commonLibrary.LibraryCode),
 			rego.Module("Generic", platformGeneralQuery.LibraryCode),
 			rego.Module(query.Query, query.Content),
+			rego.Module("dd_iac_compat.rego", utils.RegoCompatShim),
```
ChouraquiBen, line 778: TIL: we build rego queries in a very interesting way

ChouraquiBen (approved):
LGTM

PR #180, opened 2026-06-11
Fix Rego correctness bugs in shared libraries

whitemerch:
## Motivation

Static analysis of the shared Rego libraries revealed a set of correctness bugs: variable names shadowing OPA built-ins (`object`, `array`, `min`, `max`, `lower`, `time`), duplicate rule definitions, a dead `else` branch with a constant `false` condition, and inconsistently named function arguments across overloaded heads. These issues carry runtime risk ranging from subtle wrong behavior to the engine picking the wrong rule head.

A Regal config is also added to suppress the `leaked-internal-reference` rule, which false-positives on engine-injected underscore-prefixed input fields (`_path`, `_op`, `_selector`, `_value`). Those keys are input data, not internal Rego rules, so the convention is deliberately different from what the lint rule is designed to catch.

## Changes

**Variable shadowing (`cicd.rego`, `common.rego`, `dockerfile.rego`, `k8s.rego`, `openapi.rego`)**

Local variables named `object`, `array`, `min`/`max`, `lower`, and `time` were renamed to non-shadowing alternatives (`obj`, `arr`, `minVal`/`maxVal`, `lowerLimit`, `timeStr`, etc.). The rename is purely local to each rule body.

**Duplicate rule definitions (`terraform.rego`)**

Two exact duplicates of `matches/2` and one exact duplicate of `check_member/2` were removed. OPA silently evaluates all heads, so duplicates waste work but don't change the result, removing them is safe.

**Dead condition (`terraform.rego`)**

`empty_array` had an `else if { false }` branch that could never match. Removed.

**Inconsistent function arguments (`terraform.rego`)**

`portOpenToInternet` is overloaded with both a singular-rule and a rules-array form. The two array heads used `rules` as the first parameter name while the singular head used `rule`. Unified to `rule` across all five heads (the array heads iterate with `r := rule[_]`).

**Regal config (`.regal/config.yaml`)**

Ignores `leaked-internal-reference` with an explanatory comment. All other rules remain at their default levels.

## Author Checklist

## QA Instruction

Run the library test suite:

```
cd assets/libraries && opa test .
```

Expected: `PASS: 312/312`.

Optionally run Regal to confirm `bugs/*` is clean:

```
regal lint assets/libraries
```

## Blast Radius

Affects only the shared Rego library files consumed at scan time. The renamed variables are purely local; no public function signatures changed. The `portOpenToInternet` fix touches argument names only, callers pass positional arguments and are unaffected.

## Additional Notes

A companion PR in `datadog-iac-scanner-default-rules` fixes the same class of bugs in the detection rules, including 66 `sprintf` argument mismatches that were corrupting finding message text.

### DIFF
--- a/assets/libraries/common.rego
+++ b/assets/libraries/common.rego
@@ -85,9 +85,9 @@ calc_IP_value(ip) := result if {
 }
 
 # Checks if a value is within a range
-between(value, min, max) if {
-	value >= min
-	value <= max
+between(value, minVal, maxVal) if {
+	value >= minVal
+	value <= maxVal
 }
 
 # Checks if a list contains an item
@@ -333,8 +333,8 @@ unsecured_cors_rule(methods, headers, origins) if {
 }
 
 get_module_equivalent_key(provider, moduleName, resource, key) := keyInResource if {
-	providers := data.common_lib.modules[provider]
-	module := providers[moduleName]
+	providerModules := data.common_lib.modules[provider]
+	module := providerModules[moduleName]
 	inArray(module.resources, resource)
 	keyInResource := module.inputs[key]
 }
@@ -555,11 +555,11 @@ has_wildcard(statement, typeAction) if {
 # array_vals := ["elem1", "elem2", "elem4"]
 #
 # return_value := {"valid": false, "searchKey": "elem1.elem2"}
-get_nested_values_info(object, array_vals) := return_value if {
+get_nested_values_info(obj, array_vals) := return_value if {
 	arr := [x |
 		some i, _ in array_vals
 		path := array.slice(array_vals, 0, i + 1)
-		walk(object, [path, _]) # evaluates to false if path is not in object
+		walk(obj, [path, _]) # evaluates to false if path is not in object
 		x := path[i]
 	]
 
--- a/assets/libraries/terraform.rego
+++ b/assets/libraries/terraform.rego
@@ -33,20 +33,20 @@ portOpenToInternet(rule, port) if {
 	containsPort(rule, port)
 }
 
-portOpenToInternet(rules, port) if {
-	rule := rules[_]
-	rule.action != "deny"
-	check_cidr(rule)
-	rule.protocol == "tcp"
-	containsPort(rule, port)
+portOpenToInternet(rule, port) if {
+	r := rule[_]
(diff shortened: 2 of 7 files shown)


assets/libraries/common.rego
```diff
@@ -333,8 +333,8 @@ unsecured_cors_rule(methods, headers, origins) if {
 }
 
 get_module_equivalent_key(provider, moduleName, resource, key) := keyInResource if {
-	providers := data.common_lib.modules[provider]
```
ChouraquiBen, line 336: Question: is providers a reserved keyword?

assets/libraries/terraform.rego
```diff
@@ -33,20 +33,20 @@ portOpenToInternet(rule, port) if {
 	containsPort(rule, port)
 }
 
-portOpenToInternet(rules, port) if {
-	rule := rules[_]
-	rule.action != "deny"
-	check_cidr(rule)
-	rule.protocol == "tcp"
-	containsPort(rule, port)
+portOpenToInternet(rule, port) if {
```
ChouraquiBen, line 36: nit: would have name that `ruleList` so that it is clear that it is a list of rules

assets/libraries/terraform.rego
```diff
@@ -579,24 +577,6 @@ check_member(attribute, search) if {
 	startswith(attribute.member, search)
 }
 
-check_member(attribute, search) if {
```
ChouraquiBen, line 582: Question: probably written in the description but are they not needed? Unused I guess?

ChouraquiBen (approved):
A few comments but nothing blocking, lgtm

assets/libraries/common.rego
whitemerch: it's an OPA builtin, not a reserved keyword in the language-syntax sense, but a builtin function namespace

assets/libraries/terraform.rego
whitemerch: They were byte-for-byte identical copies of the same rule body, and yes it is in the description

assets/libraries/terraform.rego
whitemerch: yeah it makes sense

PR #181, opened 2026-06-11
Preserve Terraform data source policies with unknown values

whitemerch:
## Motivation

Terraform `aws_iam_policy_document` data sources can contain values that are unknown during static analysis. Previously, a single unknown value caused JSON serialization to fail, which dropped the whole policy document from analysis and emitted noisy error logs.

## Changes

Treat unknown cty values as null before serializing Terraform data source policies, preserving the statically-known portions of the policy instead of dropping the whole document. The remaining serialization failure log is downgraded because routine unknown Terraform values are now handled explicitly.

A regression test covers a policy with an unknown scalar field while preserving known statement fields.

## Author Checklist

## QA Instruction

1. `go test ./pkg/parser/terraform/... -run Test_getDataSourcePolicy -v`
2. Confirm the unknown-value fixture still emits the known IAM policy statement instead of an empty policy.

## Blast Radius

This PR affects Terraform data source policy parsing in the IaC scanner. It only changes handling of unknown values while decoding `aws_iam_policy_document` data sources.

## Additional Notes

### DIFF
--- a/pkg/parser/terraform/data_source.go
+++ b/pkg/parser/terraform/data_source.go
@@ -257,7 +257,7 @@ func parseDataSourceBody(ctx context.Context, body *hclsyntax.Body, inputVariabl
 			contextLogger.Debug().Msgf("Error trying to eval data source block: %s", decErr.Summary)
 			return ""
 		}
-		contextLogger.Debug().Msgf("Dismissed unresolvable reference when decoding policy: %s", decErr.Summary)
+		contextLogger.Debug().Msgf("Dismissed unresolvable reference when decoding policy: %s: %s", decErr.Summary, decErr.Detail)
 	}
 
 	dataSourceJSON := decodeDataSourcePolicy(ctx, target)
--- a/pkg/parser/terraform/data_source.go
+++ b/pkg/parser/terraform/data_source.go
@@ -122,9 +122,9 @@ func getDataSourcePolicy(ctx context.Context, currentPath string, inputVariables
 
 func decodeDataSourcePolicy(ctx context.Context, value cty.Value) dataSourcePolicy {
 	contextLogger := logger.FromContext(ctx)
-	jsonified, err := ctyjson.Marshal(value, cty.DynamicPseudoType)
+	jsonified, err := ctyjson.Marshal(cty.UnknownAsNull(value), cty.DynamicPseudoType)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


ChouraquiBen (commented):
The code seems to be doing what it means to do to me, but I would like a little bit more context if possible 🙏 
Do you know which data resources could not be resolved? Within the test, I am not sure which part is supposed to make the scan fail if the feature is inactive

whitemerch:
> The code seems to be doing what it means to do to me, but I would like a little bit more context if possible 🙏 Do you know which data resources could not be resolved? Within the test, I am not sure which part is supposed to make the scan fail if the feature is inactive

@ChouraquiBen The affected data source is `aws_iam_policy_document`. When any scalar field in a statement (effect, sid, id, version, condition.test, condition.variable, principals.type) references a Terraform variable that has no default value, the parser was failing to serialise the entire policy block and silently dropping it from analysis, so IAM policy rules would never evaluate it.

If you want to try it out:  run the new test added in this PR, but without what is added in the scanner in this PR. It will fail because `getDataSourcePolicy` returns an empty result for `partial_unknowns`, whereas the expected output is a valid policy statement with effect, actions, and resources all present

pkg/parser/terraform/data_source.go
```diff
@@ -244,13 +245,19 @@ func parseDataSourceBody(ctx context.Context, body *hclsyntax.Body, inputVariabl
...
-		if decErr.Summary != "Unknown variable" {
+		isUnknownVar := decErr.Summary == "Unknown variable"
+		isUnsupportedAttrOnVar := decErr.Summary == "Unsupported attribute" &&
+			strings.Contains(decErr.Detail, "does not have an attribute named")
+		if !isUnknownVar && !isUnsupportedAttrOnVar {
 			contextLogger.Debug().Msgf("Error trying to eval data source block: %s", decErr.Summary)
 			return ""
 		}
-		contextLogger.Debug().Msg("Dismissed Error when decoding policy: Found unknown variable")
+		contextLogger.Debug().Msgf("Dismissed unresolvable reference when decoding policy: %s", decErr.Summary)
```
ChouraquiBen: nit: Could be great to log what the unresolvable reference was

ChouraquiBen (approved):
LGTM

PR #182, opened 2026-06-11
Helm render error logs

whitemerch:
## Motivation

Many Helm charts in large monorepos are deployed via build systems that inject additional value files at deploy time (release names, datacenter config, OCI image references, etc.) that are not present in the chart's own `values.yaml`. The scanner cannot replicate that injection at scan time, so rendering those charts fails with Go template missing-value errors. This is a known, intentional limitation: full value resolution from build metadata would require parsing BAZEL files, which is out of scope given the complexity and narrow population of affected charts.

The consequence of these render failures was a severe log-level problem, not a loss of findings. When rendering fails the scanner falls back to scanning raw template files, which still detects and emits genuine security findings,  the only degradation is that `resource_name` may contain an unresolved template expression rather than the concrete deployed name. Each render failure was producing 3–4 duplicate ERROR lines across independent pipeline layers, compounding further because each raw template file containing Go template syntax then produced an additional ERROR from the YAML parser. None of these logs carried information not already available at the outermost layer, and their volume made the ERROR logs noisy relative to failures that actually warrant attention.

## Changes

**Helm resolver**

All `Error()` calls in `runInstall` and `renderHelm` are removed. Every failure path in those functions now returns the error silently. The `resolverSink` is the canonical signal for a chart render failure and now carries the full specific error message via `errors.Wrap` rather than the generic message that previously discarded the underlying cause.

**Render-failure classification**

`resolverSink` now distinguishes expected render failures (charts that reference deploy-time values absent at scan time, identified by well-known Go template missing-value error signatures) from unexpected ones. Expected failures log at Debug and register the chart directory as a known fallback; any other render failure stays at Error. The fallback announcement in `resolveChartDir` is similarly demoted to Debug since the render error is already logged by the layer below.

**Raw template parse errors**

`Service` tracks which chart directories failed to render. When the YAML parser fails on a file inside one of those directories, the expected outcome for raw Go templates, the error is logged at Debug. Parse failures on all other files remain at Error, preserving visibility into genuine parser bugs.

## Author Checklist

## QA Instruction

1. `go test ./pkg/runner/... ./pkg/resolver/helm/... ./pkg/engine/provider/...`
2. Scan a Helm chart that requires deploy-time values and confirm: zero ERROR lines are emitted for that chart, the raw template files inside it produce no ERROR, and a single Debug line with the specific template error is present when debug logging is enabled.
3. Scan a non-Helm YAML file with intentionally invalid syntax and confirm the parse failure still surfaces at Error.

## Impact

Only the Helm rendering and raw-template fallback path are affected. No scan results, rule logic, or output formats change, only log levels for known, handled failure conditions are reduced.

## Additional Notes

### DIFF
--- a/pkg/engine/provider/filesystem.go
+++ b/pkg/engine/provider/filesystem.go
@@ -462,11 +462,9 @@ func (s *FileSystemSourceProvider) resolveChartDir(ctx context.Context, path str
 	contextLogger := logger.FromContext(ctx)
 	excluded, errRes := resolverSink(ctx, strings.ReplaceAll(path, "\\", "/"))
 	if errRes != nil {
-		// resolverSink returns an error only for actual chart directories (with Chart.yaml).
-		// For non-chart directories it returns nil/[] and we should not log.
-		if _, statErr := os.Stat(filepath.Join(path, "Chart.yaml")); statErr == nil {
-			contextLogger.Warn().Err(errRes).Msgf("Failed to render Helm chart '%s'; scanning its raw files as a fallback", path)
-		}
+		// The render failure is already logged by the resolver sink; this is
+		// just the fallback announcement, so keep it at Debug.
+		contextLogger.Debug().Msgf("Scanning raw files of Helm chart '%s' as a fallback after render failure", path)
 		return nil
 	}
 	s.mu.Lock()
--- a/pkg/engine/provider/filesystem.go
+++ b/pkg/engine/provider/filesystem.go
@@ -465,7 +465,7 @@ func (s *FileSystemSourceProvider) resolveChartDir(ctx context.Context, path str
 		// resolverSink returns an error only for actual chart directories (with Chart.yaml).
 		// For non-chart directories it returns nil/[] and we should not log.
 		if _, statErr := os.Stat(filepath.Join(path, "Chart.yaml")); statErr == nil {
-			contextLogger.Warn().Msgf("Failed to render Helm chart '%s'; scanning its raw files as a fallback", path)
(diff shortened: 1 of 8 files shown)


pkg/engine/provider/filesystem.go
```diff
@@ -462,11 +462,9 @@ func (s *FileSystemSourceProvider) resolveChartDir(ctx context.Context, path str
...
 	excluded, errRes := resolverSink(ctx, strings.ReplaceAll(path, "\\", "/"))
 	if errRes != nil {
-		// resolverSink returns an error only for actual chart directories (with Chart.yaml).
-		// For non-chart directories it returns nil/[] and we should not log.
-		if _, statErr := os.Stat(filepath.Join(path, "Chart.yaml")); statErr == nil {
-			contextLogger.Warn().Msgf("Failed to render Helm chart '%s'; scanning its raw files as a fallback", path)
-		}
+		// The render failure is already logged by the resolver sink; this is
+		// just the fallback announcement, so keep it at Debug.
+		contextLogger.Debug().Msgf("Scanning raw files of Helm chart '%s' as a fallback after render failure", path)
```
MikaYuoadas, line 467: ❓ *Question:* Why remove the Chart.yaml guard? Do we want to emit this log-line for non Helm chart folder too?

MikaYuoadas (approved):
LGTM, just asked a question to clarify one change.

pkg/engine/provider/filesystem.go
whitemerch: The guard in resolveChartDir was already redundant, resolveChartDir is only reachable after checkConditions passes for a directory, and checkConditions already gates on Chart.yaml existing (lines 427-430). Any path reaching resolveChartDir is guaranteed to be a Helm chart root, so the Debug log can only fire there

PR #183, opened 2026-06-12
Enforce Rego formatting and bug linting

whitemerch:
## Motivation

With the Rego library files now free of linter-detected bugs, this PR adds tooling to keep them that way. It covers every `.rego` file in the repo, not just the libraries directory, and provides three layers of enforcement: a CI gate, an opt-in pre-commit hook, and format-on-save for VS Code and Cursor.

## Changes

**CI (`lint-rego.yml`)**

A new workflow triggers on any PR or push to `main` that touches a `.rego` file or the Regal config. It checks formatting with `opa fmt --fail` across the whole repo, then runs `regal lint` restricted to the `bugs` category. Style, idiomatic, performance, import, and testing findings are suppressed so the gate stays focused on correctness.

**Pre-commit hooks (`.pre-commit-config.yaml`)**

Two local hooks cover all staged `.rego` files:

1. `opa fmt --fail` — blocks the commit if any file is not properly formatted.
2. `regal lint --fail-level error` with non-bug categories disabled — blocks only on real correctness issues.

Install with `pip install pre-commit && pre-commit install`.

**Editor formatting (`.vscode/settings.json`)**

The workspace settings file sets `opa fmt` as the default Rego formatter and enables format-on-save. This applies automatically in VS Code and Cursor for anyone with the [OPA extension](https://marketplace.visualstudio.com/items?itemName=tsandall.opa) installed. `.vscode/` is unignored so this ships with the repo.

**Regal config (`.regal/config.yaml`)**

Added suppression for `testing/test-outside-test-package` since library tests intentionally live alongside their library files rather than in a separate test package.

## QA Instruction

1. `cd assets/libraries && opa test . -v` — all library tests pass.
2. `opa fmt --fail .` — exits 0.
3. `regal lint --fail-level error --disable-category style --disable-category idiomatic --disable-category performance --disable-category imports --disable-category testing .` — exits 0.
4. Wait for the new `lint-rego` CI job to go green.

## Impact

No runtime behavior changes. CI gains a new required job for all Rego files; the pre-commit hook is opt-in. Contributors using VS Code or Cursor get automatic Rego formatting on save.

## Future work

This PR intentionally starts with the `bugs` category only. Follow-up PRs will turn the rest of Regal back on for CI and pre-commit by dropping the matching `--disable-category` flags (and any other temporary suppressions) once the corresponding findings are cleared.

The `idiomatic`, `performance`, and `style` categories currently report a large backlog and are the main candidates for that rollout. A few rules are expected to stay suppressed because they conflict with deliberate contracts rather than indicating real issues: `directory-package-mismatch` (the engine requires every rule to live in `package datadog` regardless of directory), and the name-based rules such as `prefer-snake-case` and `avoid-get-and-list-prefix` on shared library functions (these names are a public API consumed by the detection rules and cannot be renamed without a coordinated change across repositories).

### DIFF
--- a/.github/workflows/lint-rego.yml
+++ b/.github/workflows/lint-rego.yml
@@ -1,0 +1,57 @@
+name: lint-rego
+
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - "**/*.rego"
+      - ".regal/config.yaml"
+      - ".github/workflows/lint-rego.yml"
+  push:
+    branches:
+      - main
+    paths:
+      - "**/*.rego"
+      - ".regal/config.yaml"
+
+permissions:
+  contents: read
+
+jobs:
+  lint-rego:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          persist-credentials: false
+
+      - name: Download and install OPA
+        run: |
+          OPA_VERSION="1.16.2"
+          curl -L -o opa "https://github.com/open-policy-agent/opa/releases/download/v${OPA_VERSION}/opa_linux_amd64_static"
+          chmod +x opa
+          sudo mv opa /usr/local/bin/
+          opa version
+
+      - name: Download and install Regal
+        run: |
+          REGAL_VERSION="0.41.1"
+          curl -L -o regal "https://github.com/StyraInc/regal/releases/download/v${REGAL_VERSION}/regal_Linux_x86_64"
+          chmod +x regal
+          sudo mv regal /usr/local/bin/
+          regal version
+
+      - name: Check Rego formatting
+        run: opa fmt --fail .
+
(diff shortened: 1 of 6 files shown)


.github/workflows/lint-rego.yml
```diff
@@ -0,0 +1,57 @@
...
+    paths:
+      - "**/*.rego"
+      - ".regal/config.yaml"
+      - ".github/workflows/lint-rego.yml"
+  push:
+    branches:
+      - main
+    paths:
+      - "**/*.rego"
+      - ".regal/config.yaml"
```
ChouraquiBen, line 15: The sound of the concurrency limits rule

.gitignore
```diff
@@ -20,7 +20,6 @@ bin
 *.out
 
 /.idea/
-/.vscode/
```
ChouraquiBen, line 23: nit: why though? I'm pretty sure that will lead to committing the folder 🤔

ChouraquiBen (dismissed):
LGTM

ChouraquiBen (approved):

PR #191, opened 2026-06-16
Add test-rules command and CI workflow

whitemerch:
## Motivation

When the scanner engine changes, there is no automated way to verify that existing rules still produce the expected findings. The rule test fixtures live in the default-rules repository and are now also stored in the backend API (alongside the rules themselves). This PR wires up the scanner to fetch and run those tests, so every PR gets a signal that the engine has not regressed against the full default rule suite.

## Changes

**API client**

The `Client` interface gains a `GetDefaultRulesetWithTests` method that calls the same endpoint as `GetDefaultRuleset` but with `include_tests=true`. Three new types mirror the backend model: `RuleTests`, `TestFile`, and `TestFinding`. The `Rule` struct gains a `Tests *RuleTests` field that is populated when tests are requested.

**test-rules command**

A new `test-rules` subcommand fetches the default ruleset (with tests) from the Datadog API, materialises each rule's fixture files into a temporary directory, and runs the scanner engine against them. Positive fixtures are compared against the stored expected findings; negative fixtures must produce no findings. The command exits non-zero if any rule fails, prints a per-rule pass/fail summary, and accepts `--type` and `--rule` flags to scope the run to a subset of platforms or rules.

The rule is converted to a `QueryMetadata` via a JSON:API round-trip through the scanner's own `datadog.Rule` type, meaning the test exercises the exact same rule representation the scanner uses in production.

**CI workflow**

`.github/workflows/test-rules.yml` runs the command on every pull request targeting `main` and on every push to `main`. It requires `DD_API_KEY` and `DD_APP_KEY` repository secrets. Fork pull requests will pass vacuously since secrets are not available in that context.

## QA Instruction

CI should pass

## Impact

Adds a new CLI subcommand (`test-rules`) and a CI workflow. No changes to scan behaviour, output formats, or existing commands.

### DIFF
--- a/.github/workflows/test-rules.yml
+++ b/.github/workflows/test-rules.yml
@@ -1,0 +1,85 @@
+name: test-rules
+
+# Runs the scanner's engine against the rule fixtures stored in the Datadog
+# backend. This catches engine regressions that break existing rule tests.
+#
+# Requires DD_API_KEY and DD_APP_KEY repository secrets (and optionally
+# DD_SITE if the target org is not on datadoghq.com).
+#
+# Runs on push to main and on manual dispatch. Secrets are not available on
+# fork pull requests, so automatic per-PR runs are not supported.
+
+on:
+  pull_request:
+    branches: [main]
+  push:
+    branches: [main]
+  workflow_dispatch:
+    inputs:
+      platforms:
+        description: "Comma-separated list of platforms to test (empty = all)"
+        required: false
+        default: ""
+      rules:
+        description: "Space-separated list of rule IDs to test (empty = all)"
+        required: false
+        default: ""
+
+permissions:
+  contents: read
+
+jobs:
+  test-rules:
+    name: test-rules
+    runs-on: ubuntu-latest
+    # Secrets are not available on fork pull requests; skip the job rather than
+    # failing with an unauthenticated API call.
+    if: >
+      github.event_name != 'pull_request' ||
+      github.event.pull_request.head.repo.full_name == github.repository
+    steps:
+      - name: Check out code
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          persist-credentials: false
+
+      - name: Set up Go
+        uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5
+        with:
+          go-version-file: go.mod
+          cache: false
+
+      - name: Get cache paths
+        id: go-cache-paths
+        run: echo "GO_BUILD=$(go env GOCACHE)" >>$GITHUB_OUTPUT
+
+      - name: Cache dependencies
+        uses: actions/cache@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
+        with:
+          path: ${{ steps.go-cache-paths.outputs.GO_BUILD }}
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.OS }}-build-
+            ${{ runner.OS }}-
+
+      - name: Build
+        run: go build -o bin/datadog-iac-scanner ./cmd/scanner
+
+      - name: Run rule tests
+        env:
+          DD_API_KEY: ${{ secrets.DD_API_KEY }}
+          DD_APP_KEY: ${{ secrets.DD_APP_KEY }}
+          DD_SITE: ${{ secrets.DD_SITE }}
+        run: |
+          ARGS=""
+          if [ -n "${{ inputs.platforms }}" ]; then
+            for p in $(echo "${{ inputs.platforms }}" | tr ',' ' '); do
+              ARGS="$ARGS --type $p"
+            done
+          fi
+          if [ -n "${{ inputs.rules }}" ]; then
+            for r in ${{ inputs.rules }}; do
+              ARGS="$ARGS --rule $r"
+            done
+          fi
+          ./bin/datadog-iac-scanner test-rules $ARGS
--- a/.github/workflows/test-rules.yml
+++ b/.github/workflows/test-rules.yml
@@ -2,12 +2,6 @@ name: test-rules
 
 # Runs the scanner's engine against the rule fixtures stored in the Datadog
 # backend. This catches engine regressions that break existing rule tests.
-#
-# Requires DD_API_KEY and DD_APP_KEY repository secrets (and optionally
-# DD_SITE if the target org is not on datadoghq.com).
-#
-# Runs on push to main and on manual dispatch. Secrets are not available on
-# fork pull requests, so automatic per-PR runs are not supported.
 
 on:
   pull_request:
@@ -21,7 +15,7 @@ on:
         required: false
         default: ""
       rules:
-        description: "Space-separated list of rule IDs to test (empty = all)"
+        description: "Comma-separated list of rule IDs to test (empty = all)"
         required: false
         default: ""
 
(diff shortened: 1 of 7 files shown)


.github/workflows/test-rules.yml
```diff
@@ -0,0 +1,85 @@
+name: test-rules
+
+# Runs the scanner's engine against the rule fixtures stored in the Datadog
+# backend. This catches engine regressions that break existing rule tests.
+#
+# Requires DD_API_KEY and DD_APP_KEY repository secrets (and optionally
```
jacobotb: Remove the DD_API_KEY and DD_APP_KEY and DD_SITE. They are not required for the default rules and avoids problems with forks and external PRs.

.gitignore
```diff
@@ -68,3 +68,6 @@ payload.json
 __pycache__
 
 module_support
+
+# local scanner binary
+/scanner
```
jacobotb, line 73: What's this one for?

pkg/datadog/client.go
```diff
@@ -139,7 +141,20 @@ type datadogClient struct {
...
+func (s *datadogClient) GetDefaultRulesetWithTests(ctx context.Context) (*Ruleset, error) {
+	return s.fetchDefaultRuleset(ctx, true)
+}
+
+func (s *datadogClient) fetchDefaultRuleset(ctx context.Context, includeTests bool) (*Ruleset, error) {
+	testsParam := "false"
+	if includeTests {
+		testsParam = "true"
+	}
+	path := "iac/rulesets/default-ruleset?include_tests=" + testsParam + "&include_testing_rules=true"
```
jacobotb: You can use `%t` with `fmt.Sprintf`:

```suggestion
	path := fmt.Sprintf("iac/rulesets/default-ruleset?include_tests=%t&include_testing_rules=true", includeTests)
```

cmd/scanner/test_rules.go
```diff
@@ -0,0 +1,518 @@
...
+		expected[i] = ruleFinding{
+			QueryName: ef.ShortDescription,
+			Severity:  ef.Severity,
+			Line:      ef.Line,
+			FileName:  ef.FileName,
+		}
+	}
+
+	query := source.ConvertRule(rule)
+	queryPtr := &query
```
jacobotb: You can just use `&query` in the two lines that call `testRunFiles`.

cmd/scanner/test_rules.go
```diff
@@ -0,0 +1,518 @@
...
+
+	platformStr := rule.Platform
+	resourcePlatforms := testResourcePlatformsFor(platformStr)
```
jacobotb:
```suggestion
	resourcePlatforms := testResourcePlatformsFor(rule.Platform)
```

Also, are these supposed to be the platforms in the rule, or the platforms specified in the command line? Because my original test code uses the command line, and the `validateResultShape` function then uses `resourcePlatforms` to ...

I'm not sure that `resourcePlatforms` is needed at all! It only checks that the rule's platform is in `resourcePlatforms` and, if so, it does extra checks. It doesn't fail anything if it's not there. But the platform _is_ going to be in `resourcePlatforms` whether we use the command line or the rule's platform, so it's irrelevant. Can you have a look?

cmd/scanner/test_rules.go
```diff
@@ -0,0 +1,518 @@
...
+	mismatches = append(mismatches, testCompareFindings(expected, positiveGot)...)
+	mismatches = append(mismatches, testCompareFindings([]ruleFinding{}, negativeGot)...)
+
+	return mismatches, nil
+}
+
+// testResourcePlatformsFor returns the set of platforms that require resourceType/resourceName.
+// Mirrors the logic used in the default-rules test tool.
+func testResourcePlatformsFor(platform string) map[string]struct{} {
+	resourcePlatforms := map[string]struct{}{
```
jacobotb: So, all of them?

cmd/scanner/test_rules.go
```diff
@@ -0,0 +1,518 @@
...
+		if _, ok := testAllowedIssueTypes[issueType]; !ok {
+			errs = append(errs, fmt.Sprintf("Result issueType %q is not in the allowed set", issueType))
+		}
+	}
+	return errs
+}
+
+// testCompareFindings returns mismatch messages between expected and actual findings.
+// An empty FileName in an expected finding acts as a wildcard.
+func testCompareFindings(expected, got []ruleFinding) []string {
```
jacobotb: Don't use `test` as a prefix for function names, please (here and elsewhere.)

cmd/scanner/test_rules.go
```diff
@@ -0,0 +1,518 @@
+package main
```
jacobotb, line 1: Suggestion (not necessarily for this PR): put the code that runs tests in a separate package, and then we can import it from the default-rules repo and use it there without having two copies of the same test code.

.github/workflows/test-rules.yml
```diff
@@ -0,0 +1,85 @@
...
+            for p in $(echo "${{ inputs.platforms }}" | tr ',' ' '); do
+              ARGS="$ARGS --type $p"
+            done
+          fi
+          if [ -n "${{ inputs.rules }}" ]; then
+            for r in ${{ inputs.rules }}; do
+              ARGS="$ARGS --rule $r"
+            done
+          fi
+          ./bin/datadog-iac-scanner test-rules $ARGS
```
jacobotb: You totally don't need to do this. You can use commas to separate the elements of a StringSliceFlag without having to break it down into multiple flags.

Also, I don't think this is the best way to collect arguments. You should use a Bash array for safety. Something like this:

```suggestion
          args=()
          if [ -n "${{ inputs.platforms }}" ]; then
            args+=(--type "${{ inputs.platforms }}")
          fi
          if [ -n "${{ inputs.rules }}" ]; then
            args+=(--rule "${{ inputs.rules }}")
          fi
          ./bin/datadog-iac-scanner test-rules "${args[@]}"
```

.gitignore
whitemerch: It prevents accidentally committing a locally-built binary. I ran go build ./cmd/scanner locally which dropped scanner in the root, and ended up committing it by accident

.github/workflows/test-rules.yml
whitemerch: I always confuse the config file and the rules aahhh

cmd/scanner/test_rules.go
whitemerch: I removed resourcePlatforms entirely safely

cmd/scanner/test_rules.go
whitemerch: Removed it, now it will directly look at `platformsRequiringResourceKeys`

cmd/scanner/test_rules.go
whitemerch: I will follow up in another PR

cmd/scanner/test_rules.go
```diff
@@ -0,0 +1,496 @@
...
+	"github.com/DataDog/datadog-iac-scanner/pkg/runner"
+	scanUtils "github.com/DataDog/datadog-iac-scanner/pkg/utils"
+	"github.com/google/uuid"
+	cli "github.com/urfave/cli/v3"
+)
+
+const (
+	defaultRuleTestQueryTimeoutSecs = 60
+	defaultRuleTestMaxResolverDepth = 15
+	fixtureDirPerm                  = 0750
```
jacobotb:
```suggestion
	fixtureDirPerm                  = 0700
```

jacobotb (dismissed):

jacobotb (approved):

PR #192, opened 2026-06-17
[K9CODESEC-2306] Resolve local Terraform modules during IaC scans

whitemerch:
> [!TIP]

## Motivation

Terraform scans previously treated local module calls as call-site blocks only. A rule could inspect a `module` block's inputs and map a subset of them back to expected resource attributes via `get_module_equivalent_key`, but the scanner never **materialized the resources inside a local module with the caller's values**. So a misconfiguration that only exists because of what the caller passed in was invisible to any rule that didn't carry a bespoke module-mapping branch.

This PR instantiates local modules during scans: it binds caller inputs to the module's variables (→ defaults → unknown), evaluates locals/functions/conditionals, recurses into nested local modules, and emits the resolved resources into the scan payload. Rules then run against real resources (e.g. `aws_s3_bucket`) with no module-awareness required.

Related Ticket: [K9CODESEC-2306](https://datadoghq.atlassian.net/browse/K9CODESEC-2306?atlOrigin=eyJpIjoiYmQ4MmVhNzQ0MzcxNGE0ZGFiY2VhNWRmNGEwMzE5MDUiLCJwIjoiaiJ9)

## Before / After

```hcl
# main.tf (root)
module "logs" {
  source      = "./modules/bucket"
  bucket_name = "app-logs"
  acl         = "public-read"   # insecure, set by the caller
}
```
```hcl
# modules/bucket/main.tf
variable "bucket_name" {}
variable "acl" { default = "private" }

resource "aws_s3_bucket" "this" {
  bucket = var.bucket_name
  acl    = var.acl
}
```

**Before:** the root saw the `module "logs"` call-site, and `modules/bucket/main.tf` was scanned standalone with `acl = var.acl` unresolved. The public ACL was only catchable if the S3 rule had a `get_module_equivalent_key` branch, most don't, so it was **missed**.

**After:** the module is instantiated with `acl = "public-read"`, materialized as `aws_s3_bucket.this`, and the normal "S3 bucket must not be public" rule fires on the resolved value. The called body file is suppressed and the local call-site stripped, so there is exactly one finding.

## Implementation

- New evaluator package `pkg/parser/terraform/tfeval` (module-instance tree: input binding, variable defaults, locals fixed-point, nested local modules with depth + cycle guards, output up-propagation).
- Engine wiring in `pkg/engine/module_resolve.go` + `Inspect`: instantiate local modules, inject synthetic resource documents, suppress called bodies, strip instantiated local call-sites.

## Things reviewers should know

- **Unknown values use a placeholder** (`__DD_TFEVAL_UNKNOWN__`) so attribute keys stay present. This prevents `MissingAttribute` false-positives from treating an unresolved expression as absent. Trade-off: value-equality rules see the sentinel rather than a concrete value (i.e. this swaps potential MissingAttribute FPs for potential IncorrectValue FPs).
- **Finding attribution shifts for local modules.** Findings now report at the module's *resource definition* (file + line) with the call chain, not at the call site. This changes findingId, which matters for suppression and ticket continuity in hosted scans.
- **Dedup model:** called module body files are suppressed; only *instantiated* local call-sites are stripped from root docs; remote/registry call-sites are left intact so existing Rego branches still fire; uncalled modules are still scanned standalone. Call-sites are only stripped when the module contributed synthetic docs (never remove coverage without a replacement).
- **Resilience:** evaluation is wrapped in `recover()` (a bad module logs and is skipped, scan continues); if every root fails to evaluate, no suppression/stripping is applied.
- **Root `*.tfvars` / `*.auto.tfvars`** feed module inputs; root-module resources themselves are unchanged (still scanned via the standard converter).

## Out of scope (follow-ups)

- **Remote / registry modules**: unchanged; still treated as call-site blocks. It will come in the future.
- **`count` / `for_each` expansion to N>1**: only the disable case (`count = 0`, `for_each = {}`) is handled; otherwise a block is materialized once, and `count.index` / `each.key` resolve to unknown. Incoming in the next PR.
- **Cross-resource references** (`x = aws_y.z.id`, data sources): resolve to unknown. Incoming in the next PR.
- **Multiple calls to the same module directory / per-instance attribution**: distinct instances can collapse to one synthetic doc (same defining file/line).
- **Retiring `get_module_equivalent_key`**: kept running in parallel; removal is a later PR once instantiated coverage is proven.

[Followup ticket](https://datadoghq.atlassian.net/browse/K9CODESEC-2307)

## Blast radius

All Terraform scans: local modules are evaluated and merged into the OPA payload. Remote and registry modules are unchanged.

## Testing / QA

1. `go test ./pkg/parser/terraform/tfeval/... ./pkg/engine/...`
2. `test-rules` (authoritative): confirm no regressions and net-new coverage on local-module fixtures; watch the unknown-placeholder FP rate.
3. Scan a fixture that calls a local module and confirm findings target instantiated resources with caller-resolved attributes; confirm remote/registry `module` call-sites are unchanged and there are no duplicate findings (stripped call-site + instantiated resource).

[K9CODESEC-2306]: https://datadoghq.atlassian.net/browse/K9CODESEC-2306?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/engine/module_resolve.go
+++ b/pkg/engine/module_resolve.go
@@ -18,39 +18,36 @@ import (
 )
 
 // instantiateLocalModules evaluates local modules and injects resolved resource
-// documents. Called module body files are suppressed (their resources are
-// emitted instantiated instead), and the "module" key is stripped from root
-// documents so call-site rule branches don't fire alongside the instantiated ones.
+// documents. In a called module body only the resource blocks are dropped (they
+// are re-emitted instantiated, with caller-resolved values); variable, output,
+// data and locals blocks are kept so their rules still fire exactly as in a
+// standalone scan. The instantiated local "module" call-sites are stripped from
+// root documents so call-site rule branches don't fire alongside the
+// instantiated resources.
 //
-// It mutates the input FileMetadata slice in place (clears Document /
-// LineInfoDocument for suppressed files; strips module blocks on other .tf files).
-func (c *Inspector) instantiateLocalModules(ctx context.Context, files model.FileMetadatas) (out []model.Document) {
-	defer func() {
-		if r := recover(); r != nil {
-			contextLogger := logger.FromContext(ctx)
-			contextLogger.Error().Interface("panic", r).Msg(
-				"instantiateLocalModules: recovered from panic; skipping synthetic module docs (input files may be partially updated)",
-			)
-			out = nil
-		}
-	}()
-
-	moduleDocs, suppressed, calledDirs, ok := resolveModuleDocuments(ctx, files, c.repoPath)
+// It mutates the input FileMetadata documents in place. Mutations are applied
+// only after evaluation succeeds (see resolveModulesSafely); a panic during
+// evaluation leaves the scan input untouched rather than removing coverage
+// without a synthetic replacement.
+func (c *Inspector) instantiateLocalModules(ctx context.Context, files model.FileMetadatas) []model.Document {
+	moduleDocs, suppressed, calledDirs, ok := c.resolveModulesSafely(ctx, files)
 	if !ok {
 		return nil
 	}
 	for _, f := range files {
-		if f == nil {
+		if f == nil || !isTerraformFile(f.FilePath) {
 			continue
 		}
 		if suppressed[f.ID] {
-			f.Document = model.Document{}
-			f.LineInfoDocument = nil
-		} else if isTerraformFile(f.FilePath) {
-			// Only remove local module call-sites that were instantiated; remote/registry
-			// module blocks must remain so the corresponding Rego branches can still fire.
-			stripLocalModuleCalls(f.Document, f.FilePath, c.repoPath, calledDirs)
+			// Resource blocks are re-emitted as instantiated synthetic docs, so
+			// drop only those to avoid double-counting; keep the rest of the body
+			// (variable/output/data/locals) so those rules still fire.
+			delete(f.Document, "resource")
+			delete(f.LineInfoDocument, "resource")
 		}
+		// Only remove local module call-sites that were instantiated; remote/registry
+		// module blocks must remain so the corresponding Rego branches can still fire.
+		stripLocalModuleCalls(f.Document, f.FilePath, c.repoPath, calledDirs)
 	}
 	contextLogger := logger.FromContext(ctx)
 	if len(moduleDocs) > 0 {
@@ -61,14 +58,34 @@ func (c *Inspector) instantiateLocalModules(ctx context.Context, files model.Fil
 	return moduleDocs
 }
 
+// resolveModulesSafely runs the panic-prone module evaluation under a recover so
+// a malformed module aborts only the synthetic-doc injection, not the scan. It
+// performs no in-place mutation of files, so returning ok=false on panic leaves
+// the scan input untouched and the caller skips all document mutations.
+func (c *Inspector) resolveModulesSafely(
+	ctx context.Context,
+	files model.FileMetadatas,
+) (moduleDocs []model.Document, suppressed, calledDirs map[string]bool, ok bool) {
+	defer func() {
+		if r := recover(); r != nil {
+			contextLogger := logger.FromContext(ctx)
+			contextLogger.Error().Interface("panic", r).Msg(
+				"instantiateLocalModules: recovered from panic; skipping synthetic module docs",
+			)
+			moduleDocs, suppressed, calledDirs, ok = nil, nil, nil, false
+		}
+	}()
+	return resolveModuleDocuments(ctx, files, c.repoPath)
+}
+
 // resolveModuleDocuments instantiates all local modules referenced by the
 // scanned files and returns synthetic resource documents (one per resolved
 // resource, attributed to its defining file) and the set of FileMetadata IDs
 // that should be suppressed to avoid double-scanning. Uncalled modules are
 // left untouched and continue to be scanned standalone.
 //
-// The boolean is true when at least one root module evaluated successfully and
-// the caller should apply call-site / suppression mutations; otherwise false
+// `ok` is true when at least one root module evaluated successfully and the
+// caller should apply call-site / suppression mutations; otherwise it is false
 // and the returned maps are nil.
 func resolveModuleDocuments(
 	ctx context.Context,
@@ -92,7 +109,7 @@ func resolveModuleDocuments(
 	contextLogger := logger.FromContext(ctx)
 	evaluator := tfeval.New()
 	seen := make(map[string]bool)
-	var rootEvalOK int
+	var rootEvalOK bool
 	actualCalledDirs := make(map[string]bool)
 
 	for dir := range dirsWithTf {
@@ -104,7 +121,7 @@ func resolveModuleDocuments(
 			contextLogger.Warn().Err(err).Msgf("tfeval: failed to evaluate root module %s", dir)
 			continue
 		}
-		rootEvalOK++
+		rootEvalOK = true
 		for d := range childDirs {
 			actualCalledDirs[d] = true
 		}
@@ -115,7 +132,7 @@ func resolveModuleDocuments(
 
 	// If every root evaluation failed, do not strip or suppress module bodies: that would
 	// remove coverage with no synthetic replacement.
-	if rootEvalOK == 0 {
+	if !rootEvalOK {
 		return nil, nil, nil, false
 	}
 
--- a/pkg/engine/module_resolve.go
+++ b/pkg/engine/module_resolve.go
@@ -1,0 +1,275 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache-2.0 License.
+ *
+ * This product includes software developed at Datadog (https://www.datadoghq.com)  Copyright 2024 Datadog, Inc.
+ */
+package engine
+
+import (
+	"context"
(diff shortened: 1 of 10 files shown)


pkg/engine/module_resolve.go
```diff
@@ -0,0 +1,275 @@
...
+		}
+		source, _ := call["source"].(string)
+		if source == "" {
+			continue
+		}
+		cleanSource := tfeval.StripGetterPrefix(source)
+		if !tfmodules.LooksLikeLocalModuleSource(cleanSource) {
+			continue
+		}
+		if calledDirs[filepath.Clean(filepath.Join(fileDir, cleanSource))] {
```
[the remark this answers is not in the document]
whitemerch: In Go, `filepath.Join("/some/dir", "/absolute/path")` returns `/absolute/path`, the absolute component wins. So the join produces the correct absolute key and matches `calledDirs` as expected

pkg/parser/terraform/tfeval/evaluator.go
```diff
@@ -0,0 +1,330 @@
...
+
+	for _, mb := range moduleBlocks {
+		label := blockLabel(mb)
+		if label == "" {
+			continue
+		}
+		source := knownString(mb.Body.Attributes["source"], evalCtx)
+		if source == "" || !tfmodules.LooksLikeLocalModuleSource(StripGetterPrefix(source)) {
+			continue // remote/registry sources are out of scope
+		}
```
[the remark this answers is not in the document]
whitemerch: Known gap, but will resolve in another PR, [ticket](https://datadoghq.atlassian.net/browse/K9CODESEC-2307)

pkg/engine/module_resolve.go
```diff
@@ -0,0 +1,275 @@
...
+// LineInfoDocument for suppressed files; strips module blocks on other .tf files).
+func (c *Inspector) instantiateLocalModules(ctx context.Context, files model.FileMetadatas) (out []model.Document) {
+	defer func() {
+		if r := recover(); r != nil {
+			contextLogger := logger.FromContext(ctx)
+			contextLogger.Error().Interface("panic", r).Msg(
+				"instantiateLocalModules: recovered from panic; skipping synthetic module docs (input files may be partially updated)",
+			)
+			out = nil
+		}
```
MikaYuoadas: ❓ **Question:** Shouldn’t we restore the suppressed `FileMetadatas` in this case? Else iiuc we could have cleared some `Document` and `LineInfoDocument` but still returns nil here, so we effectively just entirely removed some document from the scan with no synthetic replacement. 🤔
So imho either we restore the mutated `FileMetadatas` or we don’t mutate in place in the loop below and instead accumulate them and do the mutation at the end after the main loop.

pkg/engine/module_resolve.go
```diff
@@ -0,0 +1,275 @@
...
+	return moduleDocs
+}
+
+// resolveModuleDocuments instantiates all local modules referenced by the
+// scanned files and returns synthetic resource documents (one per resolved
+// resource, attributed to its defining file) and the set of FileMetadata IDs
+// that should be suppressed to avoid double-scanning. Uncalled modules are
+// left untouched and continue to be scanned standalone.
+//
+// The boolean is true when at least one root module evaluated successfully and
```
MikaYuoadas: 🤏 **Nitpick:** Which one? Phrasing is a little bit vague.

```suggestion
// `ok` is true when at least one root module evaluated successfully and
```

pkg/engine/module_resolve.go
```diff
@@ -0,0 +1,275 @@
...
+		abs := absPath(f.FilePath, repoPath)
+		dir := filepath.Dir(abs)
+		if prev, exists := byAbsPath[abs]; exists && prev != nil && prev.ID != f.ID {
+			contextLogger.Warn().Msgf(
+				"duplicate terraform file path in scan input %s (file ids %s and %s); using the latter",
+				abs, prev.ID, f.ID,
+			)
+		}
+		byAbsPath[abs] = f
+		filesByDir[dir] = append(filesByDir[dir], f)
```
MikaYuoadas, line 263: ❓ **Question:** In case of duplicates `byAbsPath` will use the latter `f` but `filesByDir[dir]` slice will have both `f` previous and current value appended making `byAbsPath` and `filesByDir` diverge in the total number of files they hold. Won’t that be an issue?

pkg/engine/module_resolve.go
```diff
@@ -0,0 +1,275 @@
...
+	for dir := range dirsWithTf {
+		for _, called := range tfeval.CalledLocalDirs(dir) {
+			staticCalledDirs[called] = true
+		}
+	}
+
+	contextLogger := logger.FromContext(ctx)
+	evaluator := tfeval.New()
+	seen := make(map[string]bool)
+	var rootEvalOK int
```
MikaYuoadas: 🤏 **Nitpick:** `rootEvalOk` is used as a flag only to check there was at least one successful root eval. Could have been a `bool` flag instead.

pkg/engine/module_resolve.go
```diff
@@ -0,0 +1,275 @@
...
+			contextLogger.Warn().Err(err).Msgf("tfeval: failed to evaluate root module %s", dir)
+			continue
+		}
+		rootEvalOK++
+		for d := range childDirs {
+			actualCalledDirs[d] = true
+		}
+		// evalRootDir distinguishes the same child module reached from different roots
+		// (different variable bindings) so synthetic documents are not incorrectly deduped.
+		extra = append(extra, instantiatedDocs(resources, byAbsPath, repoPath, seen, dir)...)
```
MikaYuoadas, line 130: ❓ **Question:** What happens to findings that are found on two different root stack call of the same module’s resource? Won’t they be dedup because they’ll have the same fingerprint? Is it what we want? And if so which resolved value will be shown to the end user?

pkg/engine/module_resolve.go
```diff
@@ -0,0 +1,275 @@
...
+		docs = append(docs, model.Document{
+			"id":   fm.ID,
+			"file": fm.FilePath,
+			"resource": map[string]interface{}{
+				r.Type: map[string]interface{}{
+					r.Name: tfeval.AttributesToDocument(r),
+				},
+			},
+		})
+	}
```
MikaYuoadas, line 201: ⛔ **Must fix:** The way this work currently silently drop everything that’s not a resource from the synthetic doc (variable, output, data, locals... blocks).

I tried running the current `iac-scanner` against one build from this PR on a simple example (I wasn’t original and just copied the two example files provided in this PR’s body).
Here’s the result:
| # | Finding | File | Line | `main` | PR |
|---|---------|------|------|--------|-----|
| 1 | S3 bucket ACL allows read or write to all users | `stack/main.tf` | 4 | ✅ | ❌ |
| 2 | CloudTrail log files S3 bucket is publicly accessible | `stack/main.tf` | 4 | ✅ | ❌ |
| 3 | Variable without description (×2) | `modules/bucket/main.tf` | 1, 2 | ✅ | ❌ |
| 4 | Variable without type (×2) | `modules/bucket/main.tf` | 1, 2 | ✅ | ❌ |
| 5 | **S3 bucket ACL allows read or write to all users** | **`modules/bucket/main.tf`** | **6** | ❌ | **✅ (new)** |
| 6 | S3 bucket logging disabled | `modules/bucket/main.tf` | 6 | ✅ | ✅ |
| 7 | S3 bucket without versioning | `modules/bucket/main.tf` | 6 | ✅ | ✅ |
| 8 | Resource not using tags | `modules/bucket/main.tf` | 6 | ✅ | ✅ |
| 9 | Team tag missing on AWS resource | `modules/bucket/main.tf` | 6 | ✅ | ✅ |

We see # 1 correctly is now reported as # 5 inside the module with this PR’s new local module resolution.
We can ignore # 2 here (it’s the rule that’s actually faulty and its a false positive in the `get_module_equivalent_key` case).

What’s relevant here is # 3 and # 4: as we drop every non-resource from the synthetic docs rules matching on variable, output, data or locals won’t work anymore.

MikaYuoadas (changes requested):
Great feature overall, just pointed a few area of concern that need to be addressed.

pkg/engine/module_resolve.go
whitemerch: Indeed, these rules weren't on my radar

pkg/engine/module_resolve.go
whitemerch: These rules weren't on my radar indeed

pkg/engine/module_resolve.go
whitemerch: On a duplicate abs path ,`byAbsPath` keeps the latter `f` while `filesByDir[dir]` holds both, so both IDs get suppressed but the synthetic doc is attributed only to the latter. In practice both IDs point to identical content so no finding is lost, just attributed to one of the two. What do you think?

pkg/engine/module_resolve.go
whitemerch: Follow up [[K9VULN-16354] Differentiate local module findings per caller to prevent fingerprint collisions](https://github.com/DataDog/datadog-iac-scanner/pull/197)

MikaYuoadas (approved):
LGTM, all my comments were addressed 👍

PR #193, opened 2026-06-17
[K9CODESEC-2307] Expand local module tfeval for count, for_each, and same-module refs

whitemerch:
## Motivation

This builds on [PR #192](https://github.com/DataDog/datadog-iac-scanner/pull/192), which instantiates local Terraform modules into the scan payload. Two gaps from that rollout still weakened rules: known `count` / `for_each` only disabled the zero case while a single materialized block left `count.index` and `each.*` unknown, and sibling `resource` references such as `aws_kms_key.x.description` stayed unknown because nothing in the eval context represented evaluated resource attributes.

**Before / after.** A module with `count = 3` and `bucket = "logs-${count.index}"` used to yield one bucket attribute that stayed unknown; it now yields up to three instances with concrete names when the count is known. A resource that sets an attribute from another resource in the same file can resolve when that attribute is known from the prior evaluation pass.

**JIRA ticket:** [K9CODESEC-2307](https://datadoghq.atlassian.net/browse/K9CODESEC-2307)

## Changes

**tfeval (expansion).** Known numeric `count` expands into separate instances (cap 10) with per-instance `count.index` and names like `this[0]`. Known `for_each` over map/object/set expands one instance per element (same cap) with `each.key` / `each.value`. `count = 0` and `for_each = {}` still emit no instances. Unknown expansions keep one placeholder instance so attributes remain evaluable.

**tfeval (same-module references).** Up to three passes evaluate resources, inject resolved attributes under `evalCtx.Variables[<resource type>][<name>]`, then re-run the locals fixed-point so references and locals can pick up values from sibling blocks.

## Author Checklist

## QA Instruction

1. Run `go test ./pkg/parser/terraform/tfeval/... ./pkg/engine/...`.
2. Optional: scan a fixture where a local module uses literal `count` or `for_each` and confirm instantiated synthetic resources reflect per-index or per-key values where rules read those attributes.

## Blast Radius

Terraform scans that use local module instantiation: synthetic documents may include more than one resolved resource per block and instance-style names in the inner resource map. Remote and registry `module` call sites and non-local-module paths are unchanged.

## Additional Notes

**Follow-ups / limits.** `data` sources are still not evaluated; `count` / `for_each` that are not known at eval time still use a single placeholder instance; expansion and ref passes are intentionally capped (`maxCountExpansion`, `resourceRefPasses`); reference lookup registers one object per bare resource name when multiple instances share a type and label.

[K9CODESEC-2307]: https://datadoghq.atlassian.net/browse/K9CODESEC-2307?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/engine/inspector.go
+++ b/pkg/engine/inspector.go
@@ -276,12 +276,10 @@ func (c *Inspector) Inspect(
 	contextLogger := logger.FromContext(ctx)
 	contextLogger.Debug().Msg("engine.Inspect()")
 
-	// Instantiate local Terraform modules into resolved-resource documents and
-	// drop the called module bodies from the standalone scan to avoid duplicates.
+	// Instantiate local modules into synthetic resource docs; suppress duplicate module bodies.
 	moduleDocs := c.instantiateLocalModules(ctx, files)
 
-	// instantiateLocalModules mutates files in place (clears suppressed bodies).
-	// Combine skips empty documents, so this must stay before Combine.
+	// Must run before Combine: instantiateLocalModules clears suppressed file bodies in place.
 	combinedFiles := files.Combine(ctx, false)
 
 	vulnerabilities := make([]model.Vulnerability, 0)
--- a/pkg/engine/module_resolve.go
+++ b/pkg/engine/module_resolve.go
@@ -143,11 +143,7 @@ func resolveModuleDocuments(
 		}
 	}
 
-	// Only strip call sites for dirs that contributed synthetic documents (i.e.,
-	// their files are present in the scan input). If a module was read from disk
-	// by EvaluateModule but its files are absent from the scan, instantiatedDocs
-	// drops its resources; stripping the call site would remove the only payload
-	// those files provide without any synthetic replacement.
+	// Only strip call sites when that module's files are in the scan (avoid dropping sole coverage).
 	strippedDirs := make(map[string]bool, len(actualCalledDirs))
 	for dir := range actualCalledDirs {
 		if len(filesByDir[dir]) > 0 {
@@ -158,13 +154,8 @@ func resolveModuleDocuments(
 	return extra, suppressed, strippedDirs, true
 }
 
-// instantiatedDocs builds a synthetic resource document for each resolved
-// module resource, attributed to its defining file. Root resources and those
-// whose file is out of scope are skipped. seen dedupes across multiple roots.
-//
-// v1 limitation: count/for_each and multiple calls to the same module directory
-// are not expanded into separate instances; dedupe and document id attribution
-// use the defining file only, so distinct instances can collapse to one synthetic doc.
+// instantiatedDocs builds one synthetic document per resolved module resource (skips root
+// and out-of-scope). seen dedupes across eval roots; dedupe key includes type, name, line, module path.
 func instantiatedDocs(
 	resources []tfeval.ResolvedResource,
 	byAbsPath map[string]*model.FileMetadata,
@@ -189,12 +180,14 @@ func instantiatedDocs(
 		}
 		seen[key] = true
 
+		// Use the bare resource label (not the expanded name like k[0]) so that
+		// Rego rules matching input.document[_].resource.TYPE.NAME find the resource.
 		docs = append(docs, model.Document{
 			"id":   fm.ID,
 			"file": fm.FilePath,
 			"resource": map[string]interface{}{
 				r.Type: map[string]interface{}{
-					r.Name: tfeval.AttributesToDocument(r),
+					tfeval.ResourceBaseName(r.Name): tfeval.AttributesToDocument(r),
 				},
 			},
 		})
@@ -202,13 +195,9 @@ func instantiatedDocs(
 	return docs
 }
 
-// stripLocalModuleCalls removes module entries from doc whose source resolves
-// to one of calledDirs. Remote/registry sources are left intact so existing
-// Rego branches that match call-site attributes continue to work.
-//
-// The Terraform parser stores nested HCL blocks as model.Document (a named map
-// type), so we accept both model.Document and map[string]interface{} to handle
-// either representation safely.
+// stripLocalModuleCalls drops local module blocks whose source dir is in calledDirs.
+// Remote/registry sources stay so existing Rego call-site rules still match.
+// Accepts model.Document or map[string]interface{} for nested module maps.
 func stripLocalModuleCalls(doc model.Document, filePath, repoPath string, calledDirs map[string]bool) {
 	modules := docAsMap(doc["module"])
 	if modules == nil {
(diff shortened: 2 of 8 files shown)


pkg/parser/terraform/tfeval/parse.go
```diff
@@ -219,6 +220,44 @@ func LoadRootVars(dir string) map[string]cty.Value {
...
+			return name[:idx], unquoted, false, true
+		}
+	}
+	// count: name ends with [N] where N is an integer
+	if idx := strings.LastIndex(name, "["); idx >= 0 {
+		inner := strings.TrimSuffix(name[idx+1:], "]")
+		if _, err := strconv.Atoi(inner); err == nil {
+			return name[:idx], inner, true, true
+		}
+	}
```
MikaYuoadas, line 249: 🤏 **Nitpick:** Not sure if done on purpose but we never actually check there’s a closing bracket. e.g. `foo[0` would still be considered a valid "count" instance key.

pkg/parser/terraform/tfeval/evaluator.go
```diff
@@ -4,16 +4,17 @@
...
  */
 
-// Package tfeval evaluates local Terraform modules by binding caller inputs to
-// module variables, resolving locals, recursing into nested modules, and
-// returning a flat list of resources with concrete attribute values.
-// Unresolvable references (data sources, cross-resource refs, etc.) are left
-// as cty.UnknownVal rather than sentinel strings.
+// Package tfeval evaluates local Terraform modules: bind inputs, resolve locals,
+// recurse into nested modules, emit resources with concrete cty values where possible.
+// Anything still unresolvable (e.g. data sources) stays cty.UnknownVal, not sentinels.
```
MikaYuoadas, line 9: 💬 **Remark:** Not sure why the PR rewords this comment, it was fine already 🤷

pkg/parser/terraform/tfeval/evaluator.go
```diff
@@ -26,12 +27,14 @@ import (
 )
 
 const (
-	// defaultMaxDepth bounds nested-module recursion to protect against
-	// pathological or cyclic local module graphs.
+	// defaultMaxDepth caps nested module recursion (cycles, deep graphs).
 	defaultMaxDepth = 15
-	// localsMaxPasses bounds the fixed-point iteration used to resolve locals
-	// that reference one another.
+	// localsMaxPasses caps fixed-point resolution of locals that reference each other.
```
MikaYuoadas, line 33: 💬 **Remark:** Same, current comments were fine already 🤷

pkg/parser/terraform/tfeval/evaluator.go
```diff
@@ -220,27 +347,213 @@ func (e *Evaluator) evalResourceBlocks(
...
+	makeOne func(string, *hcl.EvalContext) ResolvedResource,
+) ([]ResolvedResource, bool) {
+	countAttr := rb.Body.Attributes["count"]
+	if countAttr == nil {
+		return nil, false
+	}
+	cv, diags := countAttr.Expr.Value(evalCtx)
+	if !diags.HasErrors() && cv.IsKnown() && !cv.IsNull() && cv.Type() == cty.Number {
+		f, _ := cv.AsBigFloat().Float64()
+		n := int(f)
```
MikaYuoadas: 💡 **Suggestion:** I think the idiomatic cty conversion to int is:

```suggestion
		var n int
		_ := gocty.FromCtyValue(cv, &n)
```

Bonus: it also ensure cv actually had no fractionnal part (instead of silently dropping it).

pkg/engine/inspector.go
```diff
@@ -276,12 +276,10 @@ func (c *Inspector) Inspect(
...
 	contextLogger.Debug().Msg("engine.Inspect()")
 
-	// Instantiate local Terraform modules into resolved-resource documents and
-	// drop the called module bodies from the standalone scan to avoid duplicates.
+	// Instantiate local modules into synthetic resource docs; suppress duplicate module bodies.
 	moduleDocs := c.instantiateLocalModules(ctx, files)
 
-	// instantiateLocalModules mutates files in place (clears suppressed bodies).
-	// Combine skips empty documents, so this must stay before Combine.
+	// Must run before Combine: instantiateLocalModules clears suppressed file bodies in place.
```
MikaYuoadas, line 282: 💬 **Remark:** Not sure why this needed to change, although shorter the new comment but it drop relevant context for understanding. 🤔

pkg/engine/module_resolve.go
```diff
@@ -143,11 +143,7 @@ func resolveModuleDocuments(
 		}
 	}
 
-	// Only strip call sites for dirs that contributed synthetic documents (i.e.,
-	// their files are present in the scan input). If a module was read from disk
-	// by EvaluateModule but its files are absent from the scan, instantiatedDocs
-	// drops its resources; stripping the call site would remove the only payload
-	// those files provide without any synthetic replacement.
+	// Only strip call sites when that module's files are in the scan (avoid dropping sole coverage).
```
MikaYuoadas, line 146: 🔥 **Nice:** This one is indeed both shorter and clearer.

pkg/engine/module_resolve.go
```diff
@@ -189,26 +180,24 @@ func instantiatedDocs(
...
-// stripLocalModuleCalls removes module entries from doc whose source resolves
-// to one of calledDirs. Remote/registry sources are left intact so existing
-// Rego branches that match call-site attributes continue to work.
-//
-// The Terraform parser stores nested HCL blocks as model.Document (a named map
-// type), so we accept both model.Document and map[string]interface{} to handle
-// either representation safely.
+// stripLocalModuleCalls drops local module blocks whose source dir is in calledDirs.
+// Remote/registry sources stay so existing Rego call-site rules still match.
+// Accepts model.Document or map[string]interface{} for nested module maps.
```
MikaYuoadas, line 200: 💬 **Remark:** Seems like just a reword.

MikaYuoadas (dismissed):
LGTM just added a nit.
Not sure if all doc changes are necessary, but they don’t do any harm 🤷

pkg/parser/terraform/tfeval/evaluator.go
whitemerch: That's me telling my cursor that he is writing too much comments lmao

MikaYuoadas (approved):

pkg/parser/terraform/tfeval/evaluator.go
MikaYuoadas: Indeed AI tend to spend too be too verbose in comments (they also like tokenmaxxing it seems).

PR #198, opened 2026-06-18
Remove SimilarityID and legacy exclude-results suppression

whitemerch:
## Motivation

`SimilarityID` was computed for every finding but is not part of SARIF output, and it confused tooling and authors. Suppression via `exclude-results` in the legacy `dd-iac-scan.config` file is deprecated and not used; current IaC configuration uses `code-security.datadog.yaml` and does not expose similarity-based exclusion ([IaC Security configuration](https://docs.datadoghq.com/security/code_security/iac_security/configuration/)). This change removes the field and the unused suppression path tied to it, which cleans a lot of code.

## Changes

**Model and engine**

`SimilarityID` / `OldSimilarityID` are removed from vulnerabilities and summary rows. The engine no longer computes similarity hashes or accepts `LegacyExcludeResults` / `excludeResults` in `NewInspector`. 

[GetDatadogFingerprintHash](https://github.com/DataDog/datadog-iac-scanner/blob/f0c1748fe1d01b63e9e17e91d6db70d44117f275/pkg/report/model/sarif.go#L687), which is the one used for the SARIF, is the preferred centralized option.

## Author Checklist

## QA Instruction

1. From the repo root: `go test ./pkg/... ./internal/... ./cmd/...`
2. Run a scan with SARIF output on a small fixture and confirm results still include fingerprints/locations as before, with no regression in counts for typical rules.

## Blast Radius

It wasn't used, so it should be safe.

## Additional Notes


pkg/scan/scan.go
```diff
@@ -79,14 +79,12 @@ func (c *Client) initScan(ctx context.Context) (*executeScanParameters, error) {
 		engine.DefaultVulnerabilityBuilder,
 		c.Tracker,
 		queryFilter,
-		c.ExcludeResultsMap,
 		c.ScanParams.Config.RuleConfigs,
 		c.ScanParams.RepoPath,
```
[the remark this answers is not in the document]
whitemerch: It's the legacy config, and this field has never been used in particularly

pkg/report/model/gitlab_sast.go
```diff
@@ -128,7 +128,7 @@ func initGitlabSASTScan(start, end time.Time) gitlabSASTScan {
 func (glsr *gitlabSASTReport) BuildGitlabSASTVulnerability(issue *model.QueryResult, file *model.VulnerableFile) {
 	if len(issue.Files) > 0 {
 		vulnerability := gitlabSASTVulnerability{
-			ID:       file.SimilarityID,
+			ID:       fmt.Sprintf("%s:%s:%d", issue.QueryID, file.FileName, file.Line),
```
MarshalX: Should we use something from the vulnerability itself here to avoid ID collisions that could hit because of the same file and line?

pkg/report/model/code_climate.go
```diff
@@ -43,18 +47,21 @@ func BuildCodeClimateReport(summary *model.Summary) []CodeClimateReport {
 
 	for i := range summary.Queries {
 		for j := range summary.Queries[i].Files {
+			qr := summary.Queries[i]
+			vf := qr.Files[j]
+			fp := fmt.Sprintf("%s:%s:%d", qr.QueryID, vf.FileName, vf.Line)
```
MarshalX: Same question like in GItLab SAST

whitemerch: Yes the thing is that these files are not actually used so I just did a minimal fix. These are remnants from KICS. I will use the same `GetDatadogFingerprintHash` instead to have a more similar structure as the sarif just for safety

pkg/report/model/asff.go
```diff
@@ -122,7 +122,7 @@ func (a *AwsAccountInfo) getFinding(query *model.QueryResult, file *model.Vulner
 		CreatedAt:    *aws.String(timeFormatted),
 		Description:  *aws.String(getDescription(query, "asff")),
 		GeneratorID:  *aws.String(query.QueryID),
-		ID:           *aws.String(fmt.Sprintf("%s/%s/%s", awsRegion, awsAccountID, file.SimilarityID)),
+		ID:           *aws.String(fmt.Sprintf("%s/%s/%s/%s/%d", awsRegion, awsAccountID, query.QueryID, file.FileName, file.Line)),
```
MarshalX: To be honest I am not quite sure about these new IDs. They are dynamic length now. Before that they have been hashes with pre-defined len. Now they are even human readable. I am still limited in my knowledges where and how these reports could be imported/uploaded, but I suspect that these IDs would be visible the the users.

whitemerch: I should have mentioned it in the PR, but these reports aren't actually used. Our only format exported and accessible with the binary is the SARIF, which is using [GetDatadogFingerprintHash](https://github.com/DataDog/datadog-iac-scanner/blob/f0c1748fe1d01b63e9e17e91d6db70d44117f275/pkg/report/model/sarif.go#L687). I will use the same hash as the SARIF for correctness

PR #200, opened 2026-06-19
Unify Datadog fingerprint across findings, reports, and storage

whitemerch:
## Motivation

Finding identity was split between an in-engine legacy hash, per-format logic, and in-memory deduplication keys, which made counts and SARIF partial fingerprints easy to disagree. This change makes one stable Datadog fingerprint the source of truth from summary creation through exports and storage deduplication, and removes unused legacy similarity and exclude-results plumbing.

## Changes

**Model and engine**

Introduce `GetDatadogFingerprintHash`, stamp each `VulnerableFile` in `CreateSummary`, and remove `SimilarityID` / related counters and the old similarity computation package. The inspector and vulnerability builder no longer compute or suppress by legacy similarity IDs.

**Scan and storage**

Summary counters drop failed-similarity metrics; the scan client no longer builds exclude-results maps; in-memory deduplication keys findings with the same fingerprint hash used in summaries.

**Configuration**

Remove the legacy `exclude-results` option from parsing, migration comments, and `legacy_config.md` because suppression no longer keys off similarity IDs.

**Reports and fixtures**

All exporters read the stamped `Fingerprint` field (including CSV column rename). E2E comparison order and sample JSON under `test/assets` are updated for the new summary shape.

## Author Checklist

## QA Instruction

1. Run `go test -count=1 ./...` from the repository root (full suite passes locally).
2. Optional: `make build` then `./bin/datadog-iac-scanner scan` on a small Terraform fixture and confirm SARIF `partialFingerprints.DATADOG_FINGERPRINT` matches the fingerprint in JSON summary output for the same finding.

## Blast Radius

Affects the open-source CLI and library surfaces: JSON summary schema (removed similarity-related fields and counter), CSV column `fingerprint` instead of `similarity_id`, SARIF and other report formats, and in-memory deduplication granularity (aligned with the fingerprint hash rather than line plus search key).

## Additional Notes

Downstream, findings with the **same fingerprint hash are collapsed** into one item. That makes the fingerprint definition load-bearing for what customers see after ingestion. This PR aligns the scanner, in-memory dedupe, and SARIF on one hash so behavior matches that contract; **revisiting whether the hash is granular enough** (for example across tricky Terraform or multi-rule scenarios) is intentionally **follow-up work**, not in scope here.

For integrations we care about most today, **SARIF is the format that is consumed downstream**. Other report outputs (JSON summary, CSV, JUnit, etc.) are updated to stay consistent with the same fingerprint field, but they are not the primary integration surface in practice.


MikaYuoadas (dismissed):
Great, I like that we have now a convenient way to have specific fingerprinting algorithm per platform.

ChouraquiBen (approved):
No changes since Akim's latest review, only the branch rebased. Reapproving in his stead

PR #202, opened 2026-06-19
[K9CODESEC-2398] Add RunCustomRegoQuery and ValidateCustomRegoQuery

whitemerch:
## Motivation

Custom rule authors writing Rego policies need to know immediately whether their code is valid, not after a failed deployment. This PR adds two new exported functions, `ValidateCustomRegoQuery` and `RunCustomRegoQuery`, that give callers IDE-quality diagnostics and evaluation results for custom Rego rules without running a full scan.

Related Ticket: [K9CODESEC-2398](https://datadoghq.atlassian.net/browse/K9CODESEC-2398?atlOrigin=eyJpIjoiNGVmMjAwMWU4Y2QwNDRiMjhjMmEwNmFmNTlmN2NhMmMiLCJwIjoiaiJ9)

## Changes

**Validation pipeline**

`ValidateCustomRegoQuery` runs three phases in a single call and returns all errors simultaneously, so authors see the full picture at once rather than fixing one error at a time:

1. **Parse**: OPA parse errors are rewritten into actionable messages. Bare `.document` references (missing `input` prefix), missing package declarations, missing commas inside objects, and unbalanced parentheses each get a precise line/column location and a human-readable explanation instead of OPA's internal error text.
2. **Static AST checks**: verifies that the package is `datadog`, that a `DatadogPolicy` rule is present, that all required result fields (`documentId`, `resourceType`, `resourceName`, `searchKey`) are assigned, that `sprintf` call matches its format string, and that any aliased library has a corresponding import statement.
3. **OPA compilation**: compiles the rule against the platform libraries and surfaces type errors and unresolved references. Errors that are purely a consequence of an earlier structural problem (wrong package, missing rule, missing import) are filtered out so they do not drown out the root cause.

If the module is missing its `package` declaration, validation recovers by temporarily prepending `package datadog`, runs the static checks on the recovered AST, and shifts error line numbers back so locations still point at the user's original code.

**Evaluation**

`RunCustomRegoQuery` writes the Rego content and the IaC file to temporary files with the correct platform-specific extension (`.tf`, `.yaml`, `.json`, …), runs the scan engine in-memory, and returns findings with precise resource locations alongside any runtime errors.

**CLI**

Two new subcommands are registered under `custom`:

`datadog-iac-scanner custom validate --platform <platform> --rego <base64>` — outputs JSON `{"errors": [...]}`.

`datadog-iac-scanner custom evaluate --platform <platform> --rego <base64> --file <base64>` — runs validation first; if errors are found they are returned without scanning. Otherwise outputs JSON `{"findings": [...], "errors": [...]}`.

Both commands accept base64-encoded inputs so binary-safe content can be passed from any calling process without shell escaping concerns.

> [!NOTE]
> I have a branch with an UI and a server, let me know if you need it to better test.

## Author Checklist

## QA Instruction

1. `go test ./pkg/scan/ -run TestValidateCustomRegoQuery -v` — all scenarios (missing package, wrong package, missing rule, missing result fields, missing imports, missing commas, unbalanced parens, missing `input`, undefined variables) should pass.
2. Build: `go build ./cmd/scanner/...`
3. Validate a known-good rule end-to-end:
```
REGO=$(base64 -i <path-to-valid-rule.rego>)
./bin/datadog-iac-scanner custom validate --platform terraform --rego "$REGO"
# expect: {"errors": []}
```
4. Introduce a deliberate error (remove `resourceType` from the result object) and re-run validate — expect a `missing_result_field` error with the correct line number.
5. Evaluate against a sample Terraform file:
```
FILE=$(base64 -i <path-to-sample.tf>)
./bin/datadog-iac-scanner custom evaluate --platform terraform --rego "$REGO" --file "$FILE"
# expect: {"findings": [...], "errors": []}
```

## Blast Radius

This PR only adds new exported functions and CLI subcommands. No existing scan logic, query loading, or result reporting paths are modified. The only change to existing files is wiring the new `custom` command into the CLI entry point and a minor fix to pass the IaC file path through the scan client.

## Additional Notes

[K9CODESEC-2398]: https://datadoghq.atlassian.net/browse/K9CODESEC-2398?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/scan/custom_scan.go
+++ b/pkg/scan/custom_scan.go
@@ -37,6 +37,7 @@ const (
 	scanTargetJSON     = "scan-target.json"
 	platformARM        = "azureresourcemanager"
 	codeMissingPackage = "missing_package"
+	datadogPackage     = "package datadog"
 )
 
 // platformTempFileName returns a temp filename whose extension the scanner's file-type
@@ -70,7 +71,7 @@ func RunCustomRegoQuery(
 	}
 	defer func() { _ = os.RemoveAll(tmpDir) }()
 
-	const ownerReadWritePerm = 0o600
+	const ownerReadWritePerm = 0600
 	tmpFile := filepath.Join(tmpDir, platformTempFileName(platform))
 	if err := os.WriteFile(tmpFile, fileContent, ownerReadWritePerm); err != nil {
 		return nil, nil, fmt.Errorf("writing temp file: %w", err)
@@ -87,7 +88,6 @@ func RunCustomRegoQuery(
 		Platform:                []string{platform},
 		ChangedDefaultQueryPath: true,
 		MaxFileSizeFlag:         5,
-		QueryExecTimeout:        10, // shorter than the CLI default; custom rules should be fast
 		ScanID:                  "console",
 		MaxResolverDepth:        15,
 		FlagEvaluator:           featureflags.NewLocalEvaluator(),
@@ -136,7 +136,7 @@ func ValidateCustomRegoQuery(
 	module, parseErrs := parseRegoModule(regoContent)
 	if module == nil {
 		if hasPackageExpectedError(parseErrs) {
-			if recoveredModule, _ := parseRegoModule("package datadog\n\n" + regoContent); recoveredModule != nil {
+			if recoveredModule, _ := parseRegoModule(datadogPackage + "\n\n" + regoContent); recoveredModule != nil {
 				recoveredErrs := staticChecks(recoveredModule)
 				shiftErrorLines(recoveredErrs, -2)
 				return append(parseErrs, recoveredErrs...), nil
@@ -219,10 +219,10 @@ func hasPackageExpectedError(errs []RegoValidationError) bool {
 func shiftErrorLines(errs []RegoValidationError, delta int) {
 	for i := range errs {
 		if errs[i].StartLine > 0 {
-			errs[i].StartLine += delta
+			errs[i].StartLine = max(1, errs[i].StartLine+delta)
 		}
 		if errs[i].EndLine > 0 {
-			errs[i].EndLine += delta
+			errs[i].EndLine = max(1, errs[i].EndLine+delta)
 		}
 	}
 }
@@ -599,7 +599,7 @@ func enrichParseErrors(regoContent string, errs []RegoValidationError) []RegoVal
 		switch {
 		case e.Code == ast.ParseErr && e.Message == "package expected":
 			e.Code = codeMissingPackage
-			e.Message = "expected `package datadog` at the start of the module."
+			e.Message = "expected `" + datadogPackage + "` at the start of the module."
 			result = append(result, e)
 		case e.Code == ast.ParseErr && e.Message == "unexpected identifier token: expected number":
 			rewriteMissingInputRoot(regoContent, &e)
--- a/pkg/scan/custom_scan.go
+++ b/pkg/scan/custom_scan.go
@@ -1,0 +1,911 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache-2.0 License.
+ *
+ * This product includes software developed at Datadog (https://www.datadoghq.com)  Copyright 2024 Datadog, Inc.
+ */
+package scan
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/open-policy-agent/opa/v1/ast"
+	"github.com/open-policy-agent/opa/v1/rego"
+
+	"github.com/DataDog/datadog-iac-scanner/pkg/engine/source"
+	"github.com/DataDog/datadog-iac-scanner/pkg/featureflags"
+	"github.com/DataDog/datadog-iac-scanner/pkg/model"
+	"github.com/DataDog/datadog-iac-scanner/pkg/printer"
+	"github.com/DataDog/datadog-iac-scanner/pkg/utils"
+)
+
+// RegoValidationError is a compile-time or static-analysis error with source location.
+type RegoValidationError struct {
+	Code      string `json:"code"`
+	Message   string `json:"message"`
+	StartLine int    `json:"start_line"`
+	StartCol  int    `json:"start_col"`
+	EndLine   int    `json:"end_line"`
+	EndCol    int    `json:"end_col"`
+}
+
+const (
+	scanTargetJSON     = "scan-target.json"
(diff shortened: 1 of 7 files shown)


pkg/scan/custom_scan.go
```diff
@@ -0,0 +1,887 @@
...
+	regoContent string,
+	fileContent []byte,
+) ([]model.Vulnerability, map[string]error, error) {
+	tmpDir, err := os.MkdirTemp("", "iac-custom-scan-*")
+	if err != nil {
+		return nil, nil, fmt.Errorf("creating temp dir: %w", err)
+	}
+	defer func() { _ = os.RemoveAll(tmpDir) }()
+
+	const ownerReadWritePerm = 0o600
```
ChouraquiBen: nit: cf test_rules.go: the permissions are written `0600` and `0700`, it would be good to keep the same format for consistency

pkg/scan/custom_scan.go
```diff
@@ -0,0 +1,887 @@
...
+	params := &Parameters{
+		Path:                    []string{tmpFile},
+		QueriesPath:             []string{"."},
+		LibrariesPath:           source.LibrariesDefaultBasePath,
+		PreviewLines:            3,
+		CloudProvider:           []string{""},
+		Platform:                []string{platform},
+		ChangedDefaultQueryPath: true,
+		MaxFileSizeFlag:         5,
+		QueryExecTimeout:        10, // shorter than the CLI default; custom rules should be fast
```
ChouraquiBen: Question: Is it like the timeout discussed yesterday? The one that made the scanner non-deterministic?

pkg/scan/custom_scan.go
```diff
@@ -0,0 +1,887 @@
...
+func ValidateCustomRegoQuery(
+	ctx context.Context,
+	platform string,
+	regoContent string,
+) ([]RegoValidationError, error) {
+	// Phase 1: parse.
+	module, parseErrs := parseRegoModule(regoContent)
+	if module == nil {
+		if hasPackageExpectedError(parseErrs) {
+			if recoveredModule, _ := parseRegoModule("package datadog\n\n" + regoContent); recoveredModule != nil {
```
ChouraquiBen: nit: The linter probably did not catch it because of the trailing new lines, but it ay be interesting to make `'package datadog'` a const that could be reused

pkg/scan/custom_scan.go
```diff
@@ -0,0 +1,887 @@
...
+func hasPackageExpectedError(errs []RegoValidationError) bool {
+	for _, e := range errs {
+		if e.Code == codeMissingPackage {
+			return true
+		}
+	}
+	return false
+}
+
+func shiftErrorLines(errs []RegoValidationError, delta int) {
```
ChouraquiBen, line 219: nit: The use case for this function is that you make the lines go back by two because of the two new lines added at the beginning, but the function could be used to add a number. I would either check that delta is negative or that the StartLine and EndLine are not going beyond the actual line number

pkg/scan/custom_scan.go
```diff
@@ -0,0 +1,887 @@
...
+
+// platformTempFileName returns a filename with the right extension for the given platform
+// so the scanner's extension filter accepts the file.
+func platformTempFileName(platform string) string {
+	switch strings.ToLower(platform) {
+	case "terraform":
+		return "scan-target.tf"
+	case "cloudformation", platformARM:
+		return scanTargetJSON
+	case "kubernetes", "ansible":
```
ChouraquiBen, line 51: must-fix: using `'kubernetes"`as platform, the `k8s.rego` library won't be loaded as it won't be found. Using the `k8s` platform, the target becomes `JSON` as described below. More on this line 840.

pkg/scan/custom_scan.go
```diff
@@ -0,0 +1,887 @@
...
+		if precise.Start.Line >= v.VulnerabilityLocation.Start.Line &&
+			precise.End.Line <= v.VulnerabilityLocation.End.Line {
+			v.VulnerabilityLocation = precise
+			clampLocation(&v.VulnerabilityLocation)
+		}
+	}
+}
+
+func libraryFilesystemSource(ctx context.Context, platform string) *source.FilesystemSource {
+	return source.NewFilesystemSource(ctx, []string{"."}, []string{platform}, []string{""}, source.LibrariesDefaultBasePath, false)
```
ChouraquiBen, line 864: The platform could be normalized here using a map or something so that platform can be kubernetes and links to k8s

pkg/scan/custom_scan.go
```diff
@@ -0,0 +1,887 @@
...
+	}
+}
+
+func enrichCompileErrors(regoContent string, errs []RegoValidationError) []RegoValidationError {
+	out := make([]RegoValidationError, 0, len(errs))
+	for _, e := range errs {
+		varName, ok := unsafeVarName(e)
+		if !ok || varName == resultVar {
+			// "result is unsafe" is always a cascade of other unsafe vars; skip it.
+			continue
```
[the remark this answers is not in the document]
ChouraquiBen: Is it not still skipping? It is expected behaviour?

ChouraquiBen (commented):
A few comments 🙏

pkg/scan/custom_scan.go
whitemerch: Ah i forgot about these, let me align them

pkg/scan/custom_scan.go
whitemerch: Yes i noticed this before you added comments, that's part of why I force-pushed ahha. Sorry for that, now I am deriving it from libraryPlatform

pkg/scan/custom_scan.go
whitemerch: yes using libraryPlatform now

pkg/scan/custom_scan.go
whitemerch: indeed hahaha, i did this PR a few days ago but took time to test it with the UI (new release => binary => dd-source => deploy), which obviously I forgot about it since the discussion around the timeout happened yesterday afternoon. That's a good comment, thank you Jeff

pkg/scan/custom_scan.go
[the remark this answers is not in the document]
whitemerch: I refined and force pushed. That's part of why I resolved the comment

ChouraquiBen (approved):
LGTM 🙏

PR #205, opened 2026-06-22
Add the queries-path flag to the scanner binary

barbiergaspard:
## Motivation

This PR allows the user to define a local directory from which the IaC rules are fetched. This is useful for testing purposes when changing the rules.

## Changes

- Added a --queries-path flag
- Changed the values of QueriesPath and ChangedDefaultQueryPath scan parameters if the flag is used
- Modified the README.md to reflect previous changes 

## Author Checklist

## QA Instruction

- Run a scan on your chosen repository
- Clone the [datadog-iac-scanner-default-rules](https://github.com/DataDog/datadog-iac-scanner-default-rules)
- Either do some relevant changes to a rule, or choose a subdirectory from the rules repo (not containing every rules that flagged a violation in the first scan)
- Rerun the scan using -q QUERYDIR
- Verify that the findings are in line with the flag you just used
## Blast Radius

This change impacts the query fetching of the IaC scanner

## Additional Notes

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,18 @@ By default, the output file is named `datadog-iac-scanner-result.sarif`. Use `--
 datadog-iac-scanner scan -p REPODIR -o OUTPUTDIR --output-name OUTPUTFILE.sarif
 ```
 
+You can run the scanner using your own local rules, written in the QUERYDIR directory, with the -q flag:
+
+```bash
+datadog-iac-scanner scan -p REPODIR -o OUTPUTDIR -q QUERYDIR
+```
+
+The rule directories in the QUERYDIR directory must contain a 'query.rego' and a 'metadata.json' file. You can also specify multiple paths:
+
+```bash
+datadog-iac-scanner scan -p REPODIR -o OUTPUTDIR -q QUERYDIR1 -q QUERYDIR2
+```
+
 Run `datadog-iac-scanner scan --help` to see all available flags.
 
 ### Configuring the scan
--- a/cmd/scanner/scan.go
+++ b/cmd/scanner/scan.go
@@ -210,7 +210,7 @@ func runScan(ctx context.Context, c *cli.Command) error {
 	queriesPath := c.StringSlice("queries-path")
 	queriesPath, err = validateQueriesPaths(queriesPath)
 	if err != nil {
-		return errorWithExitCode(fmt.Errorf("Path parsing exited with error: %q", err), constants.InvalidConfigErrorCode)
+		return errorWithExitCode(fmt.Errorf("path parsing exited with error: %q", err), constants.InvalidConfigErrorCode)
 	}
 
 	changedDefaultQueryPath := len(queriesPath) > 0
--- a/cmd/scanner/scan.go
+++ b/cmd/scanner/scan.go
@@ -91,6 +91,22 @@ var scanAction = &cli.Command{
 			Usage:  "(experimental, will be removed soon) scan terraform plans",
 			Value:  false,
 		},
+		&cli.StringSliceFlag{
+			Name:    "queries-path",
+			Aliases: []string{"q"},
+			Usage:   "a list of query directories paths",
+		},
+
+		// NOTE: --x-parallelparsing flag disabled due to pre-existing race conditions
+		// in concurrent query workers (SetupLogs shared state write, docker detector
+		// shallow slice copy) that cause non-deterministic violation counts.
+		// See K9VULN-13746 for the follow-up to fix and re-enable.
+		// &cli.BoolFlag{
+		// 	Name:   "x-parallelparsing",
+		// 	Hidden: true,
+		// 	Usage:  "(experimental, will be removed soon) parse files in parallel",
+		// 	Value:  false,
+		// },
 	},
 	Action: runScan,
 }
@@ -99,6 +115,24 @@ const (
 	filePerms = 0644
 	dirPerms  = 0755
 )
+
+func validateQueriesPaths(paths []string) ([]string, error) {
+	absolutePaths, err := getAbsolutePaths(paths)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, path := range absolutePaths {
+		info, err := os.Stat(path) // follows symlink
+		if err != nil {
+			return nil, fmt.Errorf("invalid queries path %q", path)
+		}
+		if !info.IsDir() {
+			return nil, fmt.Errorf("queries path %q is not a directory", path)
+		}
+	}
+	return absolutePaths, nil
+}
 
 // nolint:gocyclo
 func runScan(ctx context.Context, c *cli.Command) error {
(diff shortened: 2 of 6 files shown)


barbiergaspard:
* **#206** <a href="https://app.graphite.com/github/pr/DataDog/datadog-iac-scanner/206?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#217](https://github.com/DataDog/datadog-iac-scanner/pull/217) <a href="https://app.graphite.com/github/pr/DataDog/datadog-iac-scanner/217?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#238](https://github.com/DataDog/datadog-iac-scanner/pull/238) <a href="https://app.graphite.com/github/pr/DataDog/datadog-iac-scanner/238?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#205** <a href="https://app.graphite.com/github/pr/DataDog/datadog-iac-scanner/205?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/datadog-iac-scanner/205?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

cmd/scanner/scan.go
```diff
@@ -164,26 +170,33 @@ func runScan(ctx context.Context, c *cli.Command) error {
 		cfg.RuleConfigs[ruleID] = rc
 	}
 
+	queriesPath := c.StringSlice("queries-path")
+	changedDefaultQueryPath := len(queriesPath) > 0
+	if !changedDefaultQueryPath {
+		queriesPath = []string{"./assets/queries"}
+	}
```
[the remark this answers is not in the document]
whitemerch: True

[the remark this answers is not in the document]
barbiergaspard: Just updated the branch to fix this. The scan will now fail in (one of) the provided path(s) is invalid or not a directory

barbiergaspard:
@codex review

cmd/scanner/scan.go
whitemerch, line 1: --queries-path paths are not resolved to absolute paths, which is wrong. 

Maybe you could do like the other params, and do something like:

`
queriesPath, err = getAbsolutePaths(queriesPath)
if err != nil {
    return errorWithExitCode(fmt.Errorf("invalid queries path: %w", err), constants.InvalidConfigErrorCode)
}
`

README.md
```diff
@@ -56,6 +56,16 @@ By default, the output file is named `datadog-iac-scanner-result.sarif`. Use `--
 datadog-iac-scanner scan -p REPODIR -o OUTPUTDIR --output-name OUTPUTFILE.sarif
 ```
 
+You can run the scanner using your own local rules, written in the QUERYDIR directory, with the -q flag:
+
+```bash
+datadog-iac-scanner scan -p REPODIR -o OUTPUTDIR -q QUERYDIR
+```
+The rule directories in the QUERYDIR directory must contain a 'query.rego' and a 'metadata.json' file. You can also specify multiple paths:
+```bash
```
whitemerch, line 67: There should be a blank line between the paragraph and the fence

pkg/engine/source/filesystem.go
```diff
@@ -291,24 +291,38 @@ func (s *FilesystemSource) GetQueries(ctx context.Context, queryParameters *Quer
...
 
-// localQueryDirs collects one-level-deep sub-directories from each path in s.Source.
+// localQueryDirs recursively collects sub-directories from each path in s.Source.
+// It does not collect sub-directories that do not contain a query.rego and metadata.json file.
 func (s *FilesystemSource) localQueryDirs(ctx context.Context) []string {
 	contextLogger := logger.FromContext(ctx)
 	var dirs []string
 	for _, src := range s.Source {
-		entries, err := os.ReadDir(src)
+		err := filepath.WalkDir(src, func(path string, d os.DirEntry, err error) error {
```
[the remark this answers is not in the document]
whitemerch: As we discussed on slack, should fix

whitemerch (commented):
It might be worth it overall some tests, either a table-driven test like in scan_test.go (or an E2E fixture similar to test/e2e/e2e_test.go) covering at least: valid dir, non-existent path, path-is-a-file, and symlink-to-dir. It's exhaustive, feel free to add more test gaps, but that's what code coverage is made for [here](https://github.com/DataDog/datadog-iac-scanner/pull/205#issuecomment-4767337086), which is to make you assess if the changes you are introducing are safe enough

pkg/engine/source/filesystem.go
```diff
@@ -291,24 +291,43 @@ func (s *FilesystemSource) GetQueries(ctx context.Context, queryParameters *Quer
...
+// It does not collect sub-directories that do not contain a query.rego and metadata.json file.
 func (s *FilesystemSource) localQueryDirs(ctx context.Context) []string {
 	contextLogger := logger.FromContext(ctx)
 	var dirs []string
 	for _, src := range s.Source {
-		entries, err := os.ReadDir(src)
+		evaluated, err := filepath.EvalSymlinks(src)
 		if err != nil {
-			contextLogger.Debug().Msgf("localQueryDirs: skipping %s: %v", src, err)
-			continue
```
barbiergaspard, line 302: This part puzzled me. Apparently we used to just skip a directory if the path was invalid. To me, it makes more sense to stop there instead of just ignoring the directory, so I changed this part. Lmk if I should revert this

cmd/scanner/scan.go
```diff
@@ -167,25 +188,46 @@ func runScan(ctx context.Context, c *cli.Command) error {
...
+			return errorWithExitCode(fmt.Errorf("invalid queries path %q", path), constants.InvalidConfigErrorCode)
+		}
+		if !info.IsDir() {
+			return errorWithExitCode(fmt.Errorf("queries path %q is not a directory", path), constants.InvalidConfigErrorCode)
+		}
+	}
+	changedDefaultQueryPath := len(queriesPath) > 0
+	if !changedDefaultQueryPath {
+		queriesPath = []string{"./assets/queries"}
+	}
```
whitemerch, line 219: `validateQueriesPaths` already calls `getAbsolutePaths`, then `os.Stat` on each result, then checks `IsDir`. The block immediately after in `runScan` repeats exactly those two checks on the already-validated queriesPath slice, so you are basically performing a second dead loop

cmd/scanner/scan.go
```diff
@@ -167,25 +188,46 @@ func runScan(ctx context.Context, c *cli.Command) error {
 		cfg.RuleConfigs[ruleID] = rc
 	}
 
+	queriesPath := c.StringSlice("queries-path")
+	queriesPath, err = validateQueriesPaths(queriesPath)
+	if err != nil {
+		return errorWithExitCode(fmt.Errorf("invalid queries path %q", err), constants.InvalidConfigErrorCode)
```
whitemerch: `err` here is the error value returned by `validateQueriesPaths`, not a path string. `%q` formats its argument as a quoted Go string literal, so this would produce something like invalid queries path "invalid queries path \"/some/path\"", quoting the inner error message rather than wrapping it

pkg/engine/source/filesystem.go
whitemerch: With return nil, if you do -q path1 -q path2 and path1 fails `EvalSymlinks`, you lose all queries from path2 with no error propagated upward. That's the worst of both worlds: it silently drops valid sources without telling the caller anything failed

pkg/engine/source/filesystem.go
```diff
@@ -291,24 +291,43 @@ func (s *FilesystemSource) GetQueries(ctx context.Context, queryParameters *Quer
...
+			contextLogger.Debug().Msgf("localQueryDirs: error evaluting %s: %v", src, err)
```
whitemerch: evaluating, not evaluting

cmd/scanner/scan_test.go
```diff
@@ -68,3 +71,53 @@ func TestApplyPlatformFilters(t *testing.T) {
...
+				assert.Contains(t, err.Error(), tt.wantErr)
+				return
+			}
+
+			require.NoError(t, err)
+			require.Len(t, got, len(tt.paths))
+			assert.True(t, filepath.IsAbs(got[0]))
+		})
+	}
+}
```
whitemerch: Missing newline at EOF, and there is an error on the diff actually

cmd/scanner/scan_test.go
```diff
@@ -68,3 +71,53 @@ func TestApplyPlatformFilters(t *testing.T) {
 		})
 	}
 }
+
+
```
whitemerch, line 112: nit: keep it one line in between the func, it's the convention

pkg/engine/source/filesystem_test.go
```diff
@@ -571,3 +571,100 @@ func TestCheckQueryIncludeWithLegacyId(t *testing.T) {
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			root := t.TempDir()
+			src, want := tt.set(t, root)
+
+			source := NewFilesystemSource(ctx, src, []string{""}, []string{""}, "", true)
+
+			got, _ := source.localQueryDirs(ctx)
```
whitemerch: The new `localQueryDirs` tests discard the returned error, so invalid-path and mixed-path cases don’t actually assert the error propagation this PR relies on. Add wantErr assertions, ideally through GetQueries if you want to lock public behavior

whitemerch (commented):
We are almost there!

An existing but empty `--queries-path` disables default Datadog rules and completes successfully with zero rules. I reproduced this locally with scanner, it exited 0 with 0 rules found 0 violations. That can create false-clean scans when users point at the wrong directory or an empty checkout. Maybe the custom path flow should fail when discovery finds no valid rule directories

whitemerch:
@codex review

cmd/scanner/scan.go
```diff
@@ -164,26 +188,38 @@ func runScan(ctx context.Context, c *cli.Command) error {
...
+		OutputPath:              outputPath,
+		OutputName:              c.String("output-name"),
+		PreviewLines:            3,
+		RepoPath:                repoDir,
+		Path:                    inputPaths,
+		QueriesPath:             queriesPath,
+		ChangedDefaultQueryPath: changedDefaultQueryPath,
+		LibrariesPath:           "./assets/libraries",
+		ReportFormats:           []string{"sarif"},
+		Platform:                selectPlatforms(c.StringSlice("type")),
```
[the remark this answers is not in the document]
barbiergaspard: Thoughts on this @whitemerch ? Not sure this is relevant. It's not really in the scope of this PR, but we could use it to decide if we want to gate the platforms to only the ones WE support, or if we want to allow for arbitrary platform. I'd personally choose the former, but I'm not sure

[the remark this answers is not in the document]
whitemerch: Not relevant

whitemerch (approved):

PR #207, opened 2026-06-23
Improve scan performance

whitemerch:
## TL;DR

> **~7.3× faster wall-clock time. 696→700 findings (intentional, not a regression).  
> No timeouts. Correctness preserved. Memory footprint stable (~1.4–1.5 GB RSS throughout).**

All benchmarks below use **`cloud-inventory` with `-t Terraform`** (28,914 files).  
Peak RSS measured with `/usr/bin/time -l`.

The scan goes from **337 s on `main`** to **46 s** (with `--x-parallelparsing`) with all improvements stacked.
The dominant individual win is a single change, the **Terraform per-directory variable cache**,
which alone cuts wall time from 134 s to 49 s (2.7×).

---

## Wall Time

```
scanner-00-main           ████████████████████████████████████████████████████  337 s  (baseline, serial)
scanner-01-parallel       █████████████████                                     137 s  (−59%)
scanner-02-inspector      █████████████████                                     134 s  (−60%)
scanner-03-tf-parser      ██████                                                 49 s  (−85%)  ← biggest leap
scanner-04-modules        ██████                                                 48 s  (−86%)
scanner-05-runner         ██████                                                 46 s  (−86%)  ← final (includes terraform.rego fix)
```

---

## Results at a Glance

| # | Binary | Parallel? | Wall (s) | vs baseline | Violations | Rules | Peak RSS |
|---|--------|:---------:|:--------:|:-----------:|:----------:|:-----:|:---------:|
| 00 | `main` (baseline)         | ✗ | **337** | —      | 696 | 106 | **1.40 GB** |
| 01 | + race fix + parallel     | ✓ | **137** | 2.5×   | 696 | 106 | **1.48 GB** |
| 02 | + OPA v1 / inspector      | ✓ | **134** | 2.5×   | 696 | 106 | **1.44 GB** |
| 03 | + terraform parser cache  | ✓ |  **49** | **6.9×** | 696 | 106 | **1.43 GB** |
| 04 | + module HCL cache        | ✓ |  **48** | **7.0×** | 697 | 106 | **1.46 GB** |
| 05 | + runner pool + `terraform.rego` fix | ✓ | **46** | **~7.3×** | 701 | 107 | **1.47 GB** |

---

## Phase Breakdown (CPU time)

CPU time is aggregate across all goroutines; wall time for a phase can be shorter when work is parallelised.

### `prepare_sources`: file parsing phase

```
scanner-00  ████████████████████████████████████████████████████████████  593 s  serial
scanner-01  ████████████████████████████████████████████████████████████████████████████████████  859 s  (more CPU, less wall — parallel)
scanner-02  ████████████████████████████████████████████████████████████████████████████████  823 s
scanner-03  ████████                                                                           86 s  ← dirVarsCache lands
scanner-04  ███████                                                                            74 s
scanner-05  ███                                                                                28 s  ← HCL body cache
```

### `start_scan`: OPA rule evaluation phase

```
scanner-00  ██████████████████████████████████████████████████████  54.4 s
scanner-01  ████████████████████████████████████████████████████████████  60.6 s  (more queries, parallel parse overhead)
scanner-02  █████████████████████████████████████████████████████████████  60.9 s
scanner-03  █████████████████████████████████████████████████████████  57.5 s
scanner-04  ██████████████████████████████████████████████           46.3 s  ← precomputed stores start helping
scanner-05  ████████████████████████████████                         32.8 s  ← runner buffer pool
```

### Full phase table (CPU seconds)

| Binary | file_type | get_queries | prepare_sources | start_scan | generate_report |
|--------|----------:|------------:|----------------:|-----------:|----------------:|
| 00-main | 8.66 | 0.29 | 593.35 | 54.39 | 0.48 |
| 01-parallel | 11.86 | 0.23 | 858.87 | 60.64 | 0.07 |
| 02-inspector | 11.70 | 0.39 | 823.03 | 60.86 | 0.78 |
| 03-tf-parser | 9.88 | 0.25 | 85.63 | 57.45 | 0.07 |
| 04-modules | 8.66 | 0.27 | 73.73 | 46.25 | 0.08 |
| 05-runner | 7.32 | 0.23 | 27.89 | 32.75 | 0.08 |

---

## Memory Profile (peak RSS)

Measurements are peak resident set size from `/usr/bin/time -l`.

```
scanner-00-main      ████████████████████████████████████████████████████████████████████████  1.40 GB  (baseline)
scanner-01-parallel  ████████████████████████████████████████████████████████████████████████  1.48 GB
scanner-02-inspector ████████████████████████████████████████████████████████████████████████  1.44 GB
scanner-03-tf-parser ████████████████████████████████████████████████████████████████████████  1.43 GB
scanner-04-modules   ████████████████████████████████████████████████████████████████████████  1.46 GB
scanner-05-runner    ████████████████████████████████████████████████████████████████████████  1.47 GB
```

**Memory is flat across all steps (~1.4–1.5 GB).** The caches (dirVarsCache, HCL body cache)
trade CPU time for speed, not memory — the saved allocations are offset by retaining parsed
structures, leaving net RSS unchanged. The dominant cost is the OPA input data (one document
per Terraform resource) held in memory during `start_scan`; that does not change across steps.

---

### 01: Race fix + `--x-parallelparsing`
**Wall: 337 s → 137 s (−59%)**

Parallel file parsing cuts wall time significantly by spreading the I/O and parse load across cores.
`prepare_sources` CPU *increases* (more total work due to thread overhead and contention), but wall
time drops because the work runs concurrently across cores instead of serially.

The underlying race fixes are necessary prerequisites:
- `vulnerability_builder.go`: removed `SetupLogs` shared-state write under concurrency
- `docker_detect.go`: changed shallow slice alias to a deep copy preventing concurrent mutations
- `inspector.go`: removed the decode-phase wall-clock deadline that was silently dropping findings

### 02: OPA v1 + precomputed stores + pre-parsed Rego
**Wall: 137 s → 134 s (−2%)**

Wall time barely moves at this step because `prepare_sources` still dominates. The OPA changes
precompute one `inmem.Store` per platform from the merged input data once per scan (instead of
rebuilding it from JSON for every query), and pre-parse the 82 KB of shared Rego library text once
at startup via `rego.ParsedModule()`. The benefit materialises in `start_scan` CPU time — it
sets up the gains visible in later steps.

### 03: Terraform per-directory variable cache (`dirVarsCache`)
**Wall: 134 s → 49 s (−63%)**

**The single biggest improvement in the suite.**

Without the cache, every `.tf` file in a directory triggers `getInputVariables` +
`getDataSourcePolicy`, which re-reads and re-parses *all* `.tf` files in that directory to
build the variable/locals/data-source map. In a large repository with many
files per directory, this is O(N²) re-parsing. The `dirVarsCache` + `singleflight.Group`
reduces it to one parse per directory, shared (read-only cloned) by all files in that directory.

### 04: Module HCL body cache + parallel `parseHCLBodies`
**Wall: 49 s → 48 s (−2%)**

Two benefits:
1. **Body cache** (`parsedBodyCache sync.Map`): the same `.tf` bytes are never lexed/parsed more
   than once per process lifetime. In multi-phase scans (module discovery, then engine inspection)
   this eliminates a full second pass over all files.
2. **Parallel parse**: `parseHCLBodies` now dispatches parse work to a worker pool (GOMAXPROCS
   workers), then merges in original file order for deterministic locals/variable shadowing.

### 05: Runner buffer pool + `terraform.rego` fix
**Wall: 48 s → 46 s (−4%)**

Two distinct sub-changes:

**Runner buffer pool** (`scanReadBufferPool sync.Pool`): reuses the 1 MiB read buffers across
files instead of allocating one per file. Reduces allocation pressure in `prepare_sources`
(CPU: 74 s → 28 s) and `start_scan` (CPU: 46 s → 33 s).

**`terraform.rego` fix** (`resolve_reference_name`): changes the rule to collect all matching
candidates into a set before selecting one, avoiding OPA's "multiple-output" error when the same
resource name appears across multiple module instances in separate documents. This is why
violations tick up from 696 to 700 — findings that were being silently suppressed by OPA
evaluation errors are now correctly reported (see Correctness section).

---

## Correctness Check

All binaries produce the **same exit code (60 = CRITICAL findings present)** and scan the same
28,914 Terraform files (`-t Terraform` on `cloud-inventory`).

| Metric | 00–04 | 05 | Notes |
|--------|:-----:|:-----:|-------|
| Files scanned | 28,914 | 28,914 | ✅ identical |
| Rules evaluated | 106 | 107 | ✅ see below |
| Violations | 696–697 | 701 | ✅ see below |
| Timed-out rules | 0 | 0 | ✅ none |
| Exit code | 60 | 60 | ✅ identical |

**+4 violations, +1 rule (05 onwards):** This is the `terraform.rego` fix for
`resolve_reference_name`. Before the fix, OPA's multi-value rule evaluation silently failed when
the same resource appeared across multiple module instance documents, suppressing findings. The
fix makes that function deterministic, surfacing 4 previously suppressed findings and making 1
additional rule produce results. This is a **correctness improvement, not a regression**.

---

## Key Takeaways

1. **~7.3× end-to-end speedup** (337 s → 46 s with `--x-parallelparsing`) over `main` on
   `cloud-inventory` / `-t Terraform`.

2. **The terraform `dirVarsCache` is the most impactful single change**: cuts wall time from
   134 s to 49 s (−63%). The root cause was O(N²) re-parsing of directory contents for every
   file in that directory.

3. **Memory footprint is flat (~1.4–1.5 GB RSS) across all steps.** The caches trade CPU cycles
   for speed without adding net memory. 

4. **No regressions.** All scans complete, finding counts either match or increase for documented
   correctness reasons, and no rules time out on any improved binary.

## Author Checklist

### DIFF
--- a/cmd/scanner/main.go
+++ b/cmd/scanner/main.go
@@ -17,10 +17,13 @@ import (
 
 const defaultFailCode = 126
 
-// gcPercent sets the garbage collection target percentage. A lower value makes
-// the GC run more frequently, reclaiming transient allocations from OPA query
-// evaluation faster and reducing peak memory usage.
-const gcPercent = 50
+// gcPercent is the GC target (GOGC). The scan is a short-lived batch job whose
+// live heap stays small (~150 MB on large repos), but OPA evaluation and HCL
+// parsing churn huge volumes of short-lived objects. A low GOGC collects very
+// frequently and, in profiles, GC accounts for ~40% of scan CPU. Letting the
+// heap grow ~3x between collections trades a modest amount of peak memory for a
+// large reduction in GC CPU. Override with GOGC; cap peak memory with GOMEMLIMIT.
+const gcPercent = 200
 
 func main() {
 	if _, ok := os.LookupEnv("GOGC"); !ok {
--- a/cmd/scanner/main.go
+++ b/cmd/scanner/main.go
@@ -17,13 +17,10 @@ import (
 
 const defaultFailCode = 126
 
-// gcPercent is the GC target (GOGC). The scan is a short-lived batch job whose
-// live heap stays small (~150 MB on large repos), but OPA evaluation and HCL
-// parsing churn huge volumes of short-lived objects. A low GOGC collects very
-// frequently and, in profiles, GC accounts for ~40% of scan CPU. Letting the
-// heap grow ~3x between collections trades a modest amount of peak memory for a
-// large reduction in GC CPU. Override with GOGC; cap peak memory with GOMEMLIMIT.
-const gcPercent = 200
+// gcPercent sets the garbage collection target percentage. A lower value makes
+// the GC run more frequently, reclaiming transient allocations from OPA query
+// evaluation faster and reducing peak memory usage.
+const gcPercent = 50
 
 func main() {
 	if _, ok := os.LookupEnv("GOGC"); !ok {
--- a/pkg/engine/inspector.go
+++ b/pkg/engine/inspector.go
@@ -32,11 +32,12 @@ import (
 	"github.com/DataDog/datadog-iac-scanner/pkg/utils"
 	"github.com/hashicorp/hcl/v2"
 	"github.com/hashicorp/hcl/v2/hclsyntax"
-	"github.com/open-policy-agent/opa/ast"           // nolint:staticcheck
-	"github.com/open-policy-agent/opa/cover"         // nolint:staticcheck
-	"github.com/open-policy-agent/opa/rego"          // nolint:staticcheck
-	"github.com/open-policy-agent/opa/storage/inmem" // nolint:staticcheck
-	"github.com/open-policy-agent/opa/topdown"       // nolint:staticcheck
+	"github.com/open-policy-agent/opa/v1/ast"
+	"github.com/open-policy-agent/opa/v1/cover"
+	"github.com/open-policy-agent/opa/v1/rego"
+	"github.com/open-policy-agent/opa/v1/storage"
+	"github.com/open-policy-agent/opa/v1/storage/inmem"
+	"github.com/open-policy-agent/opa/v1/topdown"
 	"github.com/pkg/errors"
 	"github.com/zclconf/go-cty/cty"
 )
@@ -72,6 +73,13 @@ type QueryLoader struct {
 	platformLibraries map[string]source.RegoLibraries
 	querySum          int
 	QueriesMetadata   []model.QueryMetadata
+	// parsedCommon is the Common library module parsed once at startup.
+	// Passed via rego.ParsedModule() to skip re-parsing 40 KB of Rego text on
+	// every PrepareForEval call; each call still compiles its own fresh module
+	// set so there is no shared mutable compiler state across goroutines.
+	parsedCommon *ast.Module
+	// parsedGeneric holds the per-platform Generic library module, also parsed once.
+	parsedGeneric map[string]*ast.Module
 }
 
 // VulnerabilityBuilder represents a function that will build a vulnerability
@@ -155,7 +163,7 @@ func NewInspector(
 	}
 	platformLibraries := getPlatformLibraries(ctx, queriesSource, queries)
 
-	queryLoader := prepareQueries(queries, commonLibrary, platformLibraries, tracker)
+	queryLoader := prepareQueries(ctx, queries, commonLibrary, platformLibraries, tracker)
 
 	failedQueries := make(map[string]error)
 
@@ -224,7 +232,7 @@ func (c *Inspector) createInspectionJobs(jobs chan<- InspectionJob, queries []mo
 func (c *Inspector) performInspection(ctx context.Context, scanID string, filesMap map[string]*model.FileMetadata,
 	astPayload ast.Value,
 	jobs <-chan InspectionJob, results chan<- QueryResult, queries []model.QueryMetadata,
-	modules []tfmodules.ParsedModule) {
+	modules []tfmodules.ParsedModule, baseStores map[string]storage.Store) {
 	for job := range jobs {
 		select {
 		case <-ctx.Done():
@@ -233,8 +241,12 @@ func (c *Inspector) performInspection(ctx context.Context, scanID string, filesM
 		default:
 		}
 
-		queryOpa, err := c.QueryLoader.LoadQuery(ctx, &queries[job.queryID], modules)
+		loadStart := time.Now()
+		queryOpa, err := c.QueryLoader.LoadQuery(ctx, &queries[job.queryID], modules, baseStores)
+		loadDur := time.Since(loadStart)
 		if err != nil {
+			contextLogger := logger.FromContext(ctx)
+			contextLogger.Warn().Err(err).Msgf("failed to load query %s", queries[job.queryID].Query)
 			continue
 		}
 
@@ -252,7 +264,12 @@ func (c *Inspector) performInspection(ctx context.Context, scanID string, filesM
 			FlagEvaluator: c.flagEvaluator,
 		}
 
+		evalStart := time.Now()
 		vuls, err := c.doRun(ctx, queryContext)
+		evalDur := time.Since(evalStart)
+		contextLogger := logger.FromContext(ctx)
+		contextLogger.Debug().Msgf("query timing: load=%s eval=%s query=%s",
+			loadDur.Round(time.Millisecond), evalDur.Round(time.Millisecond), queries[job.queryID].Query)
 		if err == nil {
 			c.tracker.TrackQueryExecution(query.Metadata.Aggregation)
 		}
@@ -310,6 +327,10 @@ func (c *Inspector) Inspect(
 
 	queries := c.getQueriesByPlat(platforms)
 
+	// Pre-build one inmem.Store per platform so LoadQuery does not re-parse the
+	// same payload for every PrepareForEval call.
+	baseStores := precomputeBaseStores(c.QueryLoader.precomputeBaseInputData(ctx, enrichedModules))
+
 	// Compute the file map once and share it (read-only) across all workers
 	filesMap := files.ToMap()
 
@@ -328,7 +349,7 @@ func (c *Inspector) Inspect(
 		go func() {
 			// Decrement the counter when the goroutine completes
 			defer wg.Done()
-			c.performInspection(ctx, scanID, filesMap, astPayload, jobs, results, queries, enrichedModules)
+			c.performInspection(ctx, scanID, filesMap, astPayload, jobs, results, queries, enrichedModules, baseStores)
 		}()
 	}
 	// Start a goroutine to create inspection jobs
@@ -369,6 +390,7 @@ func processResult(ctx context.Context, result *QueryResult,
 	queries []model.QueryMetadata, c *Inspector) {
 	contextLogger := logger.FromContext(ctx)
 	if result.err != nil {
+		contextLogger.Warn().Err(result.err).Msgf("query failed to execute: %s", queries[result.queryID].Query)
 		c.failedQueriesMu.Lock()
 		c.failedQueries[queries[result.queryID].Query] = result.err
 		c.failedQueriesMu.Unlock()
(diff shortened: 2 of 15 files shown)


cmd/scanner/main.go
```diff
@@ -17,10 +17,13 @@ import (
...
-// the GC run more frequently, reclaiming transient allocations from OPA query
-// evaluation faster and reducing peak memory usage.
-const gcPercent = 50
+// gcPercent is the GC target (GOGC). The scan is a short-lived batch job whose
+// live heap stays small (~150 MB on large repos), but OPA evaluation and HCL
+// parsing churn huge volumes of short-lived objects. A low GOGC collects very
+// frequently and, in profiles, GC accounts for ~40% of scan CPU. Letting the
+// heap grow ~3x between collections trades a modest amount of peak memory for a
+// large reduction in GC CPU. Override with GOGC; cap peak memory with GOMEMLIMIT.
+const gcPercent = 200
```
MarshalX: Should we gradually increase it? For example we could back to default 100

pkg/engine/inspector.go
```diff
@@ -745,28 +773,108 @@ func ShouldSkipVulnerability(command model.CommentsCommands, queryID, legacyQuer
...
+
+	// Pre-parse the Common and platform-Generic library modules once so that
+	// each LoadQuery call can pass them via rego.ParsedModule() and skip
+	// re-tokenizing ~82 KB of Rego text on every PrepareForEval call.
+	// Each PrepareForEval still compiles its own independent module set, so
+	// there is no shared mutable compiler state and no concurrency hazard.
+	parsedCommon, err := ast.ParseModuleWithOpts("Common", commonLibrary.LibraryCode,
+		ast.ParserOptions{RegoVersion: ast.RegoV1})
+	if err != nil {
+		contextLogger.Warn().Err(err).Msg("Failed to pre-parse Common Rego library; will re-parse per query")
```
MarshalX: Love this shared parsing! I made the same during my hot path optimizations and it gain nicely. 

Does it make sense to fallback to slow path if we can not compile common library? I mean why we want to re-parse it lately. Is it possible that the parsing of the same rego code failed now, but will succeed later? I would prefer to hard fail here the entire scan

pkg/parser/terraform/modules/modules.go
```diff
@@ -71,110 +73,214 @@ const (
...
-					if !diag.HasErrors() &&
-						val.Type().Equals(cty.String) &&
-						!val.IsNull() {
-						localsMap[name] = val.AsString()
-					}
-				}
-			case "variable":
-				if len(block.Labels) != 1 {
+	bodies := make([]*hclsyntax.Body, len(tfFiles))
+	numWorkers := runtime.GOMAXPROCS(0)
```
MarshalX: I would love to be able to control it. We do have the similar thing already in codebase. `c.ScanParams.ParallelScanFlag`. I think we should reuse it.

If you decide to create your own flag here, consider to use `AdjustNumWorkers` which is a wrapper for `runtime.GOMAXPROCS`

cmd/scanner/main.go
MarshalX: where do we set GOMEMLIMIT btw?

pkg/engine/source/source.go
MarshalX, line 115: here AbsSource is not used

so we should rebuild the map key the same way like in modules.go

smth like this?
```go
key := module.Source
if module.IsLocal && module.AbsSource != "" {
    key = module.AbsSource
}
providersMap[key] = attrData
```

pkg/parser/terraform/modules/modules.go
```diff
@@ -71,110 +73,214 @@ const (
...
+
+		// Use the resolved absolute path as the key for local modules so that
+		// two roots with identical relative sources (e.g. "./module") are stored
+		// as separate entries rather than one silently overwriting the other.
+		key := mod.Source
+		if mod.IsLocal && mod.AbsSource != "" {
+			key = mod.AbsSource
+		}
+		if _, exists := modules[key]; !exists {
+			modules[key] = mod
```
MarshalX, line 239: you are using `mod.AbsSource` as key here which is great to prevent collisions, but during the merge in `source.go` you are missing this case around key building with `AbsSource`

line 115, `providersMap[module.Source]`. could you pls verify?

pkg/parser/terraform/modules/modules.go
```diff
@@ -71,110 +73,214 @@ const (
...
-		filePath := file.FilePath
-		if !isTerraformFile(filePath) {
-			continue
-		}
-		baseDir := filepath.Dir(filePath)
+// parsedBodyCache memoizes parsed *hclsyntax.Body keyed by content. The same
+// Terraform bytes can be parsed more than once per scan (module discovery,
+// enrichment, etc.); caching avoids redundant lex/parse work. Parsed bodies are
+// read-only after construction, so sharing them across goroutines is safe.
+var parsedBodyCache sync.Map // bodyCacheKey -> *hclsyntax.Body
```
MarshalX: pls correctly me if i am wrong, but i do not see any caps or logic around keeping this cache inbond. to have control over memory growth and prevent any memory leaks here 

also, it lives at package level. which will be not cleaned up in case of long-living process. for example our HTTP server for IDE. is it expected to be shared between scans? if not, pls move to per-scan scope

pkg/parser/terraform/terraform.go
```diff
@@ -35,6 +38,15 @@ type Parser struct {
...
 	terraformVarsPath string
 	sciInfo           model.SCIInfo
+
+	// dirVarsCache memoizes the per-directory variable/locals/data-source
+	// resolution. getInputVariables + getDataSourcePolicy read and parse every
+	// *.tf file in the directory, so without this cache a directory of N files
+	// re-parses all N files N times (O(N²)). The result depends only on the
+	// directory contents and the (constant) terraformVarsPath, so it is shared
+	// read-only across all files in the same directory.
+	dirVarsCache sync.Map // dir -> converter.VariableMap
```
MarshalX, line 44: probably it will introduce problems for upcoming HTTP server for IDEs as well if the Parser gonna be reused between scans. pls check

pkg/engine/inspector.go
```diff
@@ -315,6 +332,10 @@ func (c *Inspector) Inspect(
 
 	queries := c.getQueriesByPlat(platforms)
 
+	// Pre-build one inmem.Store per platform so LoadQuery does not re-parse the
+	// same payload for every PrepareForEval call.
+	baseStores := precomputeBaseStores(c.QueryLoader.precomputeBaseInputData(ctx, enrichedModules))
```
MarshalX, line 335: that would be awesome to cover this with `-race` test. not critical

MarshalX (commented):
Great job! Awesome performance gains! 

The rest is only about correctness of shared cache between scans, better control of memory consumption and cache hash collisions

pkg/engine/inspector.go
whitemerch: Verified and it looks ok

pkg/engine/source/source.go
whitemerch: The inconsistency is moot: `get_module_equivalent_key` will be called by zero rules in the repo, the full legacy enrichment path (ParseAllModuleVariables → source.go) is effectively dead, and I'll remove it in a dedicated follow-up PR, which I already have locally. This is following this big change in terraform modules:

- https://github.com/DataDog/datadog-iac-scanner/pull/192

- https://github.com/DataDog/datadog-iac-scanner/pull/193 

- https://github.com/DataDog/datadog-iac-scanner/pull/197

pkg/parser/terraform/modules/modules.go
whitemerch: Same answer as above

cmd/scanner/main.go
whitemerch: - Given this is a short-lived CLI batch job, i have put it as 200 as it is defensible, and the data came up as positive for the performance. I don't have a strong opinion there though.

- We never set `GOMEMLIMIT` anywhere; we only set GOGC. The Go runtime already honors `GOMEMLIMIT` from the environment if an operator sets it. I initially did it differently as I was testing few things for the performance, but I missed the comment left on the const which isn't adequate to the actual state anymore

MikaYuoadas: I’d be very careful with tampering with GOGC, it was lowered [here](https://github.com/DataDog/datadog-iac-scanner/pull/88) because we were hitting frequent OOM error on large repo.
The impact isn’t at all a few MB like stated in the comment. On my benchmark I was going from 2.8GB to 1.8GB peak heap size using `GOGC=50` instead of the default of `100` (so a 800MB peak reduction).
Setting it to 200 would in the same benchmark put the peak live heap to around 4.5GB....
Also I don’t know where you got the 40% GC CPU overhead, but my benchmark found only a 4% increase scan time with `GOGC=50` vs `100`

`GOMEMLIMIT` is not set, I wanted to but with the scan concurrency of 3 we potentially have 3 separate go binary running in the same pod, so what do we set the `GOMEMLIMIT` to?
* `GOMEMLIMIT = pod memory limit` risk having concurrent scan collectively crossing the pod memory limit and killing it (OOM)
* `GOMEMLIMIT = 1/3 pod memory limit` is a bit wasteful as we can’t have scan bursts more memory while other concurrent scan are not using it.

So I’d advise to be reaaaally careful modifying that value and this should be backed by numbers, run some actual benchmarking before changing it.

whitemerch: As discussed async, there were indeed no improvements changing GOGC, so I removed this change. It was a mistake done in a batch of other improvements that made me think it changed the performance a lot

pkg/parser/terraform/modules/modules.go
```diff
@@ -71,110 +73,212 @@ const (
...
-		body, ok := hclFile.Body.(*hclsyntax.Body)
-		if !ok {
-			contextLogger.Error().Msgf("Unexpected body type in %s", filePath)
-			continue
-		}
+	bodies := make([]*hclsyntax.Body, len(tfFiles))
+	numWorkers = utils.AdjustNumWorkers(numWorkers)
+	if numWorkers < minParseWorkers {
+		numWorkers = minParseWorkers
+	}
```
MarshalX: we are doing workers normalization from the params using `AdjustNumWorkers` and then making `floor` to already normalized value. so even if someone will pass only 2 workers, the 4 will be spawn. i do feel lack of control and transparency here. WDYT?

whitemerch: You're right, that's a bug in the logic. Let me fix it

MarshalX (approved):
🏎️

PR #210, opened 2026-06-23
Replace per-query timeout with slow-rule logging; deprecate `--timeout` option

MarshalX:
## Motivation

A per-query wall-clock timeout (`--timeout`) was bounding rego eval + result decode. Under worker contention this cut queries off at a timing-dependent point, producing non-deterministic finding counts and silently dropping findings (the same class of bug addressed for decode in #209). Rather than tuning the timeout, we want queries to run to completion and instead **surface slow rules via logs** so they can be debugged/optimized later — without failing the scan.

This deprecates and disables `--timeout`, removes query time-bounding, and adds slow-rule logging.

JIRA Ticket: https://datadoghq.atlassian.net/browse/K9CODESEC-2452 (follow-up to #209 / K9CODESEC-2411)

## Changes

- **`pkg/engine/inspector.go`**
  - Removed the per-query timeout: `doRun` no longer wraps eval in `context.WithTimeout`; eval runs on the scan context and is only stopped by real scan cancellation. Removed the `queryExecTimeout` field/computation/log.
  - Added non-enforcing **slow-rule logging**: times eval + decode and, past `slowQueryWarnThreshold` (60s), emits a structured `Warn` (`event=slow_rule`, `queryID`, `platform`, `evalMs`, `decodeMs`, `totalMs`, `thresholdMs`). The query always completes — it is never failed or dropped for being slow.
- **`cmd/scanner/scan.go`** — `--timeout` is now **hidden** and marked `(DEPRECATED)`; logs a deprecation warning when set; no longer wired into scan parameters.
- **`pkg/scan/client.go`** — `Parameters.QueryExecTimeout` marked `Deprecated: ignored` (k**ept for struct/CLI compatibility**).

## Author Checklist

No new unit test: the change removes a code path (timeout enforcement) and adds observability logging; behavior is covered by existing engine/scan tests, and the removed `e2e-cli-056` no longer applies.

## QA Instruction

- Scan a large repo several times — finding counts are stable run-to-run (no timeout-driven variance), and no query is aborted.
- Run with `--timeout 5` (or any value) — you should see a deprecation warning (`the --timeout flag is deprecated and no longer has any effect…`) and **identical results** to running without it.
- If a rule is genuinely slow (eval + decode > 60s), confirm a single structured log line with `event=slow_rule` and the `*Ms` fields appears; the scan still completes normally.

## Blast Radius

DataDog IaC Scanner only (query engine + CLI). Behavioral change: queries are no longer time-bounded, so a single pathological rule could run longer than before (mitigated by logging, and bounded by overall scan/process lifecycle). `--timeout` remains accepted for backward compatibility but has no effect. No change to which findings are produced for normal rules.

## Additional Notes

### DIFF
--- a/pkg/engine/inspector.go
+++ b/pkg/engine/inspector.go
@@ -128,7 +128,6 @@ func NewInspector(
 	queryParameters *source.QueryInspectorParameters,
 	ruleConfigs map[string]config.IacRuleConfig,
 	repoPath string,
-	queryTimeout int,
 	useOldSeverities bool,
 	needsLog bool,
 	numWorkers int,
--- a/pkg/engine/inspector.go
+++ b/pkg/engine/inspector.go
@@ -55,6 +55,10 @@ const (
 	regoQuery = utils.RegoQuery
 )
 
+// slowQueryWarnThreshold is how long a single query (rego eval + result decode)
+// may take before it is logged as slow. We surface slow rules so they can be optimized later.
+const slowQueryWarnThreshold = 60 * time.Second
+
 // ErrNoResult - error representing when a query didn't return a result
 var ErrNoResult = errors.New("query: not result")
 
@@ -92,7 +96,6 @@ type Inspector struct {
 	repoPath             string
 	enableCoverageReport bool
 	coverageReport       cover.Report
-	queryExecTimeout     time.Duration
 	useOldSeverities     bool
 	numWorkers           int
 	flagEvaluator        featureflags.FlagEvaluator
@@ -161,12 +164,6 @@ func NewInspector(
 		Add(docker.DetectKindLine{}, model.KindDOCKER).
 		Add(terraform.DetectKindLine{}, model.KindTerraform)
 
-	queryExecTimeout := time.Duration(queryTimeout) * time.Second
-
-	if needsLog {
-		contextLogger.Info().Msgf("Query execution timeout=%v", queryExecTimeout)
-	}
-
 	return &Inspector{
 		QueryLoader:      &queryLoader,
 		vb:               vb,
@@ -175,7 +172,6 @@ func NewInspector(
 		ruleConfigs:      ruleConfigs,
 		detector:         lineDetector,
 		repoPath:         repoPath,
-		queryExecTimeout: queryExecTimeout,
 		useOldSeverities: useOldSeverities,
 		numWorkers:       utils.AdjustNumWorkers(numWorkers),
 		flagEvaluator:    flagEvaluator,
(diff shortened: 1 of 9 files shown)


MikaYuoadas (dismissed):

pkg/engine/inspector.go
```diff
@@ -445,11 +438,13 @@ func (c *Inspector) doRun(ctx context.Context, qCtx *QueryContext) (vulns []mode
 		options = append(options, rego.EvalQueryTracer(cov))
 	}
 
-	results, err := qCtx.Query.OpaQuery.Eval(timeoutCtx, options...)
+	evalStart := time.Now()
+	results, err := qCtx.Query.OpaQuery.Eval(qCtx.Ctx, options...)
```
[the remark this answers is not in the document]
whitemerch: This looks correct to me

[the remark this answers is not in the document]
whitemerch: we could keep a bounded context only in the test-rules code path?

pkg/engine/inspector.go
```diff
@@ -445,11 +438,13 @@ func (c *Inspector) doRun(ctx context.Context, qCtx *QueryContext) (vulns []mode
...
+	evalDuration := time.Since(evalStart)
 	qCtx.payload = nil
 	if err != nil {
 		if topdown.IsCancel(err) {
-			return nil, errors.Wrap(err, "query executing timeout exited")
+			return nil, errors.Wrap(err, "query evaluation canceled (scan aborting)")
```
whitemerch, line 446: eval runs on the scan context, which means when a user cancels a scan for eg., the condition will fire and the message "query executing timeout exited" will be logged, but it won't be a timeout

pkg/scan/client.go
```diff
@@ -23,20 +23,21 @@ import (
...
+	Path                []string
+	PayloadPath         string
+	PreviewLines        int
+	QueriesPath         []string
+	LibrariesPath       string
+	ReportFormats       []string
+	Platform            []string
+	TerraformVarsPath   string
+	// Deprecated: ignored. Queries are no longer time-bounded.
 	QueryExecTimeout            int
```
whitemerch: dead code no?

whitemerch (dismissed):

pkg/engine/inspector.go
MarshalX: not sure that i get it right. i did change the message to canceled already

pkg/scan/client.go
MarshalX: i thought about it as a public code api, so i leave temporary. but now i left backward compatibility for our CLI interface only and reworked our rules test workflow

pkg/engine/inspector.go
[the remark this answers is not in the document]
MarshalX: i reworked our test rules command; also i ve added additional time guard for GHA

pkg/engine/inspector.go
whitemerch: Nevermind I don't know how I read this wrong, I still saw the "query executing timeout exited" 🤦‍♂️

whitemerch (approved):

PR #211, opened 2026-06-23
Add a long-lived `serve` HTTP server

MarshalX:
## Motivation

Bring IaC scanning into the Datadog IDE extensions, the same way SAST and secret scanning already work.

Today the scanner is a one-shot CLI (requires a git remote, fetches rules from the API per scan, reads files from disk, writes SARIF) — unsuitable for an IDE. The extension needs a long-lived, **pure-engine** HTTP server it can drive by pushing rules + file content over HTTP, mirroring the `datadog-static-analyzer-server` contract so it can reuse the same binary-manager lifecycle (keep-alive ping, graceful shutdown, version discovery).

JIRA Ticket: [K9CODESEC-2453](https://datadoghq.atlassian.net/browse/K9CODESEC-2453)

## Changes

`serve` is purely additive. The `scan` CLI is unchanged, and the virtual-FS refactor is behavior-preserving (the CLI uses a `DiskFS` passthrough).

- **`serve` command + lifecycle** mirroring the static-analyzer server: flags `--port` / `--address` / `--keep-alive-timeout` / `--enable-shutdown` (`--use-rules-cache` accepted as a no-op); `GET /ping`→`pong`, `GET /version` & `/revision`, `GET|POST /shutdown` (204/403), keep-alive idle-exit goroutine, `X-iac-scanner-server-*` + CORS headers.
- **`GET /ide/v1/iac/supported-files`** — static `{patterns, strategy}` map (directory vs single_file), derived from `analyzer.PossibleFileTypes()` with a drift test.
- **`POST /ide/v1/iac/analyze`** — content-push scan. Request `{files:[{path,content}], rules, config}` (raw content); response `{findings, missing_files, failed_queries}`. No git, no network, no SARIF.
- **Content-push + virtual FS** — new `pkg/vfs` (`FS` / `DiskFS` / `MemFS`). `MemFS` resolves Terraform siblings, vars, and local modules in memory and records referenced-but-unpushed paths as workspace-relative `missing_files` (the hybrid-escalation signal; CWD-independent, so one server can serve many workspaces). New `MemorySourceProvider` feeds pushed files to the engine instead of walking disk. `vfs.FS` is threaded through the Terraform parser/data-source/modules, the inspector, and extension detection.
- **`scan.Client` seams** — `Client.Scan(ctx)` (in-memory results) plus `WithFS` / `WithQuerySourceFactory` / `WithInMemoryScan` options, and an in-memory `initScan` branch that bypasses the disk analyzer.

## Author Checklist

## QA Instruction

The server needs the Rego support libraries on disk; rules and file content are pushed per request.

```
go build -o /tmp/iac-scanner ./cmd/scanner
/tmp/iac-scanner serve --port 49200 --enable-shutdown --libraries-path ./assets/libraries
```

- `curl 127.0.0.1:49200/ping` → `pong`; `GET /ide/v1/iac/supported-files` → the `{patterns,strategy}` array.
- `POST /ide/v1/iac/analyze`: fetch rules anonymously from `GET https://api.datadoghq.com/api/v2/static-analysis/iac/rulesets/default-ruleset` (base64-decode each `rego_query`, lowercase `platform`). Push a Terraform `main.tf` that uses `var.x` defined in a sibling `variables.tf` → expect findings, with `var.x` resolved (proves in-memory sibling resolution). Push a `module { source = "../mod" }` whose directory you *don't* send → expect `"mod"` to appear in `missing_files` as a workspace-relative path.
- **CLI regression**: `datadog-iac-scanner scan …` SARIF output must be unchanged (DiskFS path is behavior-preserving).
- **Concurrency**: `go build -race`, then fire several parallel `/analyze` requests — no `DATA RACE` in the log and the process must not exit `66`.

## Blast Radius

**Only `datadog-iac-scanner`.** `serve` is additive; the `scan` CLI and its SARIF output are unchanged (the `vfs` refactor is a DiskFS passthrough, covered by the existing suite). No backend/API changes.

## Additional Notes

- **Deferred**: Helm rendering is disabled in content-push mode (the Helm SDK needs a real chart directory) — pushed charts fall back to raw-YAML scanning; the `$ref`/include resolver isn't virtualized yet (niche OpenAPI/Ansible cross-file refs).

[K9CODESEC-2453]: https://datadoghq.atlassian.net/browse/K9CODESEC-2453?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/engine/provider/filesystem.go
+++ b/pkg/engine/provider/filesystem.go
@@ -92,6 +92,7 @@ func NewFileSystemSourceProvider(ctx context.Context, paths, excludes, onlyPaths
 func (s *FileSystemSourceProvider) addExcluded(ctx context.Context, excludePaths []string) error {
 	contextLogger := logger.FromContext(ctx)
 	for _, excludePath := range excludePaths {
+		excludePath = filepath.Clean(excludePath)
 		info, err := os.Stat(excludePath)
 		if err != nil {
 			if os.IsNotExist(err) {
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -1,0 +1,286 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache-2.0 License.
+ *
+ * This product includes software developed at Datadog (https://www.datadoghq.com)  Copyright 2024 Datadog, Inc.
+ */
+
+// Package server implements a long-lived HTTP server that analyzes IaC files on
+// demand. It mirrors the contract of the Rust datadog-static-analyzer-server so
+// the Datadog VS Code extension can manage both binaries through the same
+// lifecycle (keep-alive ping, graceful shutdown, version discovery).
+//
+// The server is a pure engine: it never reads git, calls the Datadog API, or
+// writes SARIF. Rules and file content are pushed over HTTP by the extension.
+//
+// Lifecycle endpoints (this file):
+//
+//	GET  /ping      health / keep-alive (resets the idle timer), returns "pong"
+//	GET  /version   running binary version (plain text)
+//	GET  /revision  running binary commit (plain text)
+//	POST /shutdown  graceful shutdown (204 when --enable-shutdown, else 403)
+//	GET  /shutdown  same as POST
+//
+// IaC endpoints (GET /ide/v1/iac/supported-files, POST /ide/v1/iac/analyze) are
+// registered here and implemented in sibling files.
+package server
+
+import (
+	"context"
+	"crypto/rand"
+	"encoding/hex"
+	"encoding/json"
+	"errors"
+	"net"
+	"net/http"
+	"strconv"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/DataDog/datadog-iac-scanner/internal/constants"
+	"github.com/DataDog/datadog-iac-scanner/pkg/logger"
+	"github.com/rs/zerolog/log"
+)
+
+const (
+	// readHeaderTimeout / readTimeout are explicit slowloris guards
+	readHeaderTimeout = 10 * time.Second
+	readTimeout       = 60 * time.Second
+	// shutdownGraceTimeout matches the static-analyzer server: its keep-alive
+	// loop notifies a graceful shutdown, then allows 10s before escalating to
+	// process abort. We give in-flight requests the same window via http.Server.Shutdown.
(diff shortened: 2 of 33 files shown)


pkg/engine/provider/filesystem.go
```diff
@@ -92,7 +92,8 @@
 func (s *FileSystemSourceProvider) addExcluded(ctx context.Context, excludePaths []string) error {
 	contextLogger := logger.FromContext(ctx)
 	for _, excludePath := range excludePaths {
-		info, err := os.Stat(excludePath)
+		excludePath = filepath.Clean(excludePath)
+		info, err := os.Stat(excludePath) //nolint:gosec
```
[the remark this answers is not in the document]
MarshalX: FP because addExcluded processes paths from the --exclude-paths CLI flag. The person running the scanner intentionally provides these paths. There's no server, no untrusted user, no injection surface. The same reasoning applies to --path (the scan target itself)

pkg/server/server.go
```diff
@@ -0,0 +1,330 @@
...
+	mux.HandleFunc("POST /ide/v1/iac/analyze", s.handleAnalyze)
+
+	s.http = &http.Server{
+		Addr:              net.JoinHostPort(cfg.Address, strconv.Itoa(cfg.Port)),
+		Handler:           s.middleware(mux),
+		ReadHeaderTimeout: readHeaderTimeout,
+		ReadTimeout:       readTimeout,
+		IdleTimeout:       idleTimeout,
+		// No WriteTimeout: a cold scan recompiles the rule corpus and can take
+		// several seconds until the prepared-query cache lands.
```
MikaYuoadas, line 156: 💡 **Suggestion:** Why not put a generous write timeout still?
Without it a client that stops reading can hold the goroutine indefinitely.

MikaYuoadas (approved):
LGTM 👍

pkg/server/server.go
MarshalX: I would love to, but I have no clear vision on what value to use... It could take minutes on huuuge repos. Should I just add and option for `serve ...` sub command?

MarshalX: I will have a follow up PR with rules caching etc. I will not bother it here

PR #212, opened 2026-06-24
Fix data race on `Inspector.failedQueries` (concurrent map writes -> fatal crash)

MarshalX:
## Motivation

`Inspector.failedQueries` (`pkg/engine/inspector.go`) is a plain `map[string]error` written concurrently during `Inspect` with no synchronization: the collector goroutine writes it on a query error (`processResult`), while N worker goroutines write it on a vulnerability-build failure (`getVulnerabilitiesFromQuery`). In Go, concurrent map writes are not a "maybe corrupt" race, the runtime throws an unrecoverable `fatal error: concurrent map writes` that crashes the whole process.

It stayed hidden because the map is only written on failures, so it's empty on healthy scans — which is why `-race` never caught it across the parallel real-repo runs. 

JIRA Ticket https://datadoghq.atlassian.net/browse/K9CODESEC-2487

## Changes

- Added `failedQueriesMu sync.Mutex` to `Inspector` to guard `failedQueries`.
- Guarded both writers (`processResult` and `getVulnerabilitiesFromQuery`); the lock is only taken on the failure path, so healthy scans pay nothing.
- `GetFailedQueries` now returns a locked copy via `maps.Clone` so callers can read safely while a scan is still running.
- Added `TestInspector_FailedQueriesConcurrentWrites`, which drives both writer paths concurrently and reproduces the data race under `-race` on the unfixed code.

## Author Checklist

## QA Instruction

- Run the new test against the **fix reverted**: `go test -race -run TestInspector_FailedQueriesConcurrentWrites ./pkg/engine/` should report `WARNING: DATA RACE` pointing at `inspector.go:371` (`processResult`) and `inspector.go:596-597` (`getVulnerabilitiesFromQuery`).
- With the fix in place: `go test -race ./pkg/engine/` passes cleanly (exit 0).
- Confirm `failed_queries` output is unchanged, no vb-build errors are dropped (we kept the worker-side write rather than deleting it).

## Blast Radius

DataDog IaC Scanner only, specifically `pkg/engine`. No public API or output-format change; the lock is on the failure path only, so the happy path is unaffected. Removes an unrecoverable process crash that affects the CLI (under parallel parsing) and, more severely, server mode (one panic takes down all in-flight requests).

### DIFF
--- a/pkg/engine/inspector.go
+++ b/pkg/engine/inspector.go
@@ -10,6 +10,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"maps"
 	"reflect"
 	"runtime/debug"
 	"strings"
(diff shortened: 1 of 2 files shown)


MikaYuoadas (approved):
LGTM

PR #214, opened 2026-06-26
Nathan.gallet/helm deterministic rendering

NathanGallet-dd:
## Summary

Helm charts using non-deterministic Sprig functions (`randAlphaNum`, `uuidv4`, `now`, etc.) in resource names produce a different rendered name on every scan, generating a new `DATADOG_FINGERPRINT` each time and re-reporting the same finding as new indefinitely.

**Root cause:** `engine.Engine` (used internally by `action.Install.Run`) pulls its funcmap from `sprig.TxtFuncMap()` with no injection point — `funcMap()` and `initFunMap()` are unexported, and `Engine` has no `FuncMap` field.

**Fix:** Pre-process each chart template's raw bytes before rendering, replacing non-deterministic function calls with stable, line-number-seeded stubs:

| Function | Stub format |
|---|---|
| `randAlphaNum N`, `randAlpha N`, `randAscii N` | `"ddscan{lineNum:04d}"` |
| `randNumeric N` | `"{lineNum:08d}"` |
| `uuidv4` | `"00000000-0000-0000-{lineNum:04d}-{lineNum:012d}"` |
| `now` | `(toDate "2006-01-02" "2000-01-01")` — returns `time.Time`, works in any pipeline |

Stubs are seeded by line number so two resources in the same file using the same function produce distinct, stable names across every scan. Replacements are applied back-to-front by byte offset so earlier match positions stay valid during mutation. The `now` guard skips matches preceded by `$` or `.` to avoid corrupting `$now` variable declarations and `.Values.now` field accesses.

Crypto functions (`genSignedCert`, `genCA`, etc.) are deferred — they return structs, not strings, and rarely appear in resource names.

## Changes

- `pkg/resolver/helm/helm.go` — adds `deterministicPattern` struct, `deterministicPatterns` table, `applyDeterministicSubstitutions`, and `makeDeterministic` (mirrors `setID` pattern, recurses into sub-charts). `makeDeterministic` is called before `setID` in `runInstall`.
- `pkg/resolver/helm/helm_deterministic_test.go` — unit tests for each function type + two-lines distinctness + no-match passthrough; integration test calls `Resolve` twice and asserts byte-identical output.
- `test/fixtures/test_helm_nondeterministic/` — minimal fixture chart exercising all covered function types.

## Testing

```sh
go test ./pkg/resolver/helm/... -v -run "TestApplyDeterministic|TestHelm_"
```

All existing `TestHelm_Resolve` cases remain green. `TestHelm_DeterministicRendering` calls `Resolve` twice on the new fixture and asserts `require.Equal(first.File, second.File)`.

### DIFF
--- a/pkg/resolver/helm/helm.go
+++ b/pkg/resolver/helm/helm.go
@@ -190,35 +190,54 @@ func newClient(ctx context.Context) *action.Install {
 	return client
 }
 
-// deterministicPatterns maps each non-deterministic sprig function regex to a replacement generator.
-// The generator receives the line number of the match.
-var deterministicPatterns = []struct {
+// deterministicPattern represents a non-deterministic sprig function to replace.
+type deterministicPattern struct {
 	re      *regexp.Regexp
 	replace func(lineNum int) string
-}{
+	// guard is an optional function that can reject a match. Returns true to accept, false to skip.
+	// If nil, all matches are accepted.
+	guard func(s string, matchStart int) bool
+}
+
+// deterministicPatterns maps each non-deterministic sprig function regex to a replacement generator.
+// The generator receives the line number of the match.
+var deterministicPatterns = []deterministicPattern{
 	{
 		re:      regexp.MustCompile(`\brandAlphaNum\s+\d+`),
 		replace: func(lineNum int) string { return fmt.Sprintf(`"ddscan%04d"`, lineNum) },
+		guard:   nil,
 	},
 	{
 		re:      regexp.MustCompile(`\brandAlpha\s+\d+`),
 		replace: func(lineNum int) string { return fmt.Sprintf(`"ddscan%04d"`, lineNum) },
+		guard:   nil,
 	},
 	{
 		re:      regexp.MustCompile(`\brandAscii\s+\d+`),
 		replace: func(lineNum int) string { return fmt.Sprintf(`"ddscan%04d"`, lineNum) },
+		guard:   nil,
 	},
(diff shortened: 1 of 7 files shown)


pkg/resolver/helm/helm.go
whitemerch, line 1: We are missing

- {{ randAlphaNum .Values.suffixLen }}
- {{ randBytes 16 }}

pkg/resolver/helm/helm.go
```diff
@@ -187,6 +190,192 @@ func newClient(ctx context.Context) *action.Install {
...
+	re      *regexp.Regexp
+	replace func(lineNum int) string
+	// guard is an optional function that can reject a match. Returns true to accept, false to skip.
+	// If nil, all matches are accepted.
+	guard func(s string, matchStart int) bool
+}
+
+// deterministicPatterns maps each non-deterministic sprig function regex to a replacement generator.
+// The generator receives the line number of the match.
+var deterministicPatterns = []deterministicPattern{
```
whitemerch, line 293: I don't think a chart using `{{ randAlphaNum .Values.len }} `will be matched here

pkg/resolver/helm/helm.go
whitemerch, line 1: I think Helm template comments will be matched too, even though they shouldnt

whitemerch (commented):
Nice job!

whitemerch (approved):

PR #215, opened 2026-06-26
Resolve Terraform plan line numbers from searchKey

whitemerch:
## Motivation

Terraform plan JSON is remapped under a top-level `resource` key, so line detection cannot rely on text matching for attribute-level findings. Rules today duplicate location hints via a `searchLine` field alongside `searchKey`. This change lets plan files resolve lines from `searchKey` alone so rules can drop `searchLine` for attribute-level paths.

## Changes

**File kind**

Introduce `KindTerraformPlan` and set it when the JSON parser recognizes Terraform plan content, so plan files are routed to the correct line-detection path.

**Line detection**

For `KindTerraformPlan`, convert bracket-style `searchKey` values (e.g. `aws_s3_bucket[b].acl`) into a structural path under `resource.<type>.<name>.<attr>` and resolve the line via the existing gjson `_dd_lines` lookup. Attribute-level keys are resolved structurally; resource-only keys (no attribute segment) still fall back to text matching so existing expected lines are preserved.

**Tests**

Add unit tests for path conversion and an integration test that parses a minimal plan and asserts the resolved line.

## Author Checklist

## QA Instruction

1. `go test ./pkg/detector/... ./pkg/parser/...`
2. `go test ./...`
3. Point the default-rules repo at this branch (`go mod edit -replace github.com/DataDog/datadog-iac-scanner=../datadog-iac-scanner`) and run `./tool.sh test`.
4. Remove `searchLine` from a Terraform rule with a plan JSON fixture (e.g. `terraform-alicloud-rds-instance-address-publicly-accessible`) and confirm plan line numbers still match expected results.

## Blast Radius

Datadog IaC Scanner line detection for Terraform plan JSON files. No change to `.tf` HCL detection, CloudFormation, Kubernetes, or other JSON/YAML formats.

## Additional Notes

This PR does not read the rego `searchLine` field. The engine still honors `searchLine` when rules provide it (`searchLineCalculator` in `vulnerability_builder.go`); follow-up work in the default-rules repo can remove those fields once this ships. Resource-only `searchKey` values on plans continue to use text matching rather than structural lookup.

### DIFF
--- a/pkg/detector/default_detect.go
+++ b/pkg/detector/default_detect.go
@@ -17,7 +17,19 @@ import (
 
 const (
 	undetectedVulnerabilityLine = -1
+	// resource, type, name, and at least one attribute segment.
+	tfPlanMinAttributePathLen = 4
+	// resource, type, and name — used for types where plan line metadata
+	// lives on the resource object rather than a specific attribute.
+	tfPlanMinResourcePathLen = 3
 )
+
+// tfPlanResourceLevelTypes are Terraform resource types whose plan JSON line
+// metadata is anchored on the resource object (not an attribute). Text matching
+// on the plan address field is wrong for these; resolve structurally instead.
+var tfPlanResourceLevelTypes = map[string]struct{}{
+	"aws_api_gateway_deployment": {},
+}
 
 type defaultDetectLine struct {
 }
@@ -34,14 +46,23 @@ func (d defaultDetectLine) DetectLine(ctx context.Context, file *model.FileMetad
 		ResolvedFiles:   d.prepareResolvedFiles(file.ResolvedFiles),
 	}
 
+	lines := *file.LinesOriginalData
+
+	// Terraform plan JSON is remapped under a top-level "resource" key, so the
+	// resource block has no source text to match. Resolve the searchKey path
+	// structurally (mirroring the legacy searchLine) before any text matching.
+	if file.Kind == model.KindTerraformPlan {
+		if result := detectTerraformPlanLine(searchKey, file, detector.ResolvedFile, outputLines, lines); result != nil {
+			return *result
+		}
+	}
+
 	var extractedString [][]string
 	extractedString = GetBracketValues(searchKey, extractedString, "")
 	sanitizedSubstring := searchKey
 	for idx, str := range extractedString {
 		sanitizedSubstring = strings.ReplaceAll(sanitizedSubstring, str[0], `{{`+strconv.Itoa(idx)+`}}`)
 	}
-
-	lines := *file.LinesOriginalData
 	splitSanitized := strings.Split(sanitizedSubstring, ".")
 
 	// Re-join $ref segments split by dot (e.g. "$ref=#/schemas/v1.0.Foo" → ["v1","0","Foo"])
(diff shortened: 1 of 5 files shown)


pkg/detector/default_detect.go
```diff
@@ -103,6 +124,85 @@ func (d defaultDetectLine) DetectLine(ctx context.Context, file *model.FileMetad
...
+// cannot be resolved so the caller can fall back to text matching.
+func detectTerraformPlanLine(
+	searchKey string,
+	file *model.FileMetadata,
+	resolvedFile string,
+	outputLines int,
+	lines []string,
+) *model.VulnerabilityLines {
+	path := terraformPlanPath(searchKey)
+	if len(path) < tfPlanMinResourcePathLen {
```
ChouraquiBen: Must-fix: there is a conflict between this block and the one below

pkg/detector/default_detect.go
```diff
@@ -34,14 +46,23 @@ func (d defaultDetectLine) DetectLine(ctx context.Context, file *model.FileMetad
 		ResolvedFiles:   d.prepareResolvedFiles(file.ResolvedFiles),
 	}
 
+	lines := *file.LinesOriginalData
+
+	// Terraform plan JSON is remapped under a top-level "resource" key, so the
+	// resource block has no source text to match. Resolve the searchKey path
+	// structurally (mirroring the legacy searchLine) before any text matching.
+	if file.Kind == model.KindTerraformPlan {
+		if result := detectTerraformPlanLine(searchKey, file, detector.ResolvedFile, outputLines, lines); result != nil {
```
ChouraquiBen, line 45: Remark: nice, this will work the same as it currently does with `searchKey` used as fallback if `searchLine` fails.

pkg/detector/default_detect.go
```diff
@@ -17,8 +17,20 @@ import (
...
+	// resource, type, and name — used for types where plan line metadata
+	// lives on the resource object rather than a specific attribute.
+	tfPlanMinResourcePathLen = 3
 )
 
+// tfPlanResourceLevelTypes are Terraform resource types whose plan JSON line
+// metadata is anchored on the resource object (not an attribute). Text matching
+// on the plan address field is wrong for these; resolve structurally instead.
+var tfPlanResourceLevelTypes = map[string]struct{}{
+	"aws_api_gateway_deployment": {},
```
ChouraquiBen: Question: So, this will need to be updated with all the resources that match the description above?

pkg/parser/parser.go
```diff
@@ -152,9 +158,16 @@ func (c *Parser) Parse(
 			cont = string(fileContent)
 		}
 
+		kind := c.Parsers.GetKind()
+		if ck, ok := c.Parsers.(contentKindParser); ok {
```
ChouraquiBen, line 162: Remark: This is stylish as it allows not to create many changes in the parsers but it creates an abstraction that is used by only one parser. Are there any other parsers that could benefit from it?

pkg/detector/default_detect.go
ChouraquiBen: Question: Also, what if customers create rules for resources that are not in the list? The rule will fail without them knowing there was an issue here.

Isn't there any other way?

ChouraquiBen (commented):
A few comments but the core of the idea is good!

pkg/parser/parser.go
whitemerch: I don’t see another parser with the same “same extension, different semantic kind” case today, but it could be possible in the future yes

ChouraquiBen: Okay, let's leave it as is then

pkg/parser/json/parser.go
```diff
@@ -76,6 +76,15 @@ func (p *Parser) GetKind() model.FileKind {
 	return model.KindJSON
 }
 
+// KindForContent overrides the static kind for Terraform plan JSON so line
+// detection can resolve plan resources structurally instead of by text match.
+func (p *Parser) KindForContent(content []byte) (model.FileKind, bool) {
+	if contentIsTerraformPlan(content) {
```
ChouraquiBen: Remark: this function calls `parseTFPlan` and raises the total number of `parseTFPlan` usage to 3 per TFPlan. There should be 1 or 2 TFPlan scanned per repo but this is to be noticed

pkg/detector/default_detect.go
```diff
@@ -34,14 +36,23 @@ func (d defaultDetectLine) DetectLine(ctx context.Context, file *model.FileMetad
...
+	// resource block has no source text to match. Resolve the searchKey path
+	// structurally (mirroring the legacy searchLine) before any text matching.
+	if file.Kind == model.KindTerraformPlan {
+		if result := detectTerraformPlanLine(searchKey, file, detector.ResolvedFile, outputLines, lines); result != nil {
+			return *result
+		}
+	}
+
 	var extractedString [][]string
 	extractedString = GetBracketValues(searchKey, extractedString, "")
```
ChouraquiBen, line 51: Remark: here, the bracket values are being taken away

ChouraquiBen: See below

pkg/detector/default_detect.go
ChouraquiBen, line 204: Question: Here, the array index is being read but in [[1]](https://github.com/DataDog/datadog-iac-scanner/pull/215/changes#r3490860369), the bracket values have been taken away. Is the behaviour broken or am I missing the place where they are being populated back?

ChouraquiBen (commented):
Two comments, one is about performance and can imo be overlooked, the other one is a bit more important 🙏 
Thanks for the time you're taking to tackle my comments

pkg/detector/default_detect.go
whitemerch: Not broken, the values are restored in `extractStructuralPath`. `GetBracketValues` captures `[name]` as a pair `["[name]", "name"]` in `extractedString`. The sanitization loop replaces the full bracket expression `[name]` with `{{0}}` so the dot-split works cleanly. Then `extractStructuralPath` replaces `{{0}} `back with `ext[1] = "name"` (the value without brackets) before passing the path to `GetLineBySearchLine`. gjson expects plain dotted keys, not bracket notation, so stripping [] while preserving the value is intentional

pkg/parser/json/parser.go
```diff
@@ -114,21 +114,11 @@ func (p *Parser) StringifyContent(content []byte) (string, error) {
...
-	if _, err := parseTFPlan(doc); err != nil {
-		return false
-	}
-	return true
+// looksLikeTerraformPlan is a fast byte-level heuristic that avoids a full
+// JSON parse. Both keys are required by the Terraform plan JSON spec, so their
+// co-presence is a reliable signal. Used by KindForContent and StringifyContent
+// where a full parseTFPlan call would be redundant (Parse already does it).
+func looksLikeTerraformPlan(content []byte) bool {
+	return bytes.Contains(content, []byte(`"format_version"`)) &&
```
ChouraquiBen, line 122: big

ChouraquiBen (approved):
LGTM

ChouraquiBen:
Thanks for going through all the comments 🙏

PR #219, opened 2026-06-29
Deduplicate synthetic module docs by resolved content to shrink OPA input on large repos

whitemerch:
## Motivation

On repositories with many Terraform local modules (e.g. monorepos with shared infrastructure modules called from multiple stacks), the scanner instantiates one synthetic document per resolved resource **per caller** and appends every one of them to `input.document`. Rules that iterate the full document list, including several S3 and IAM rules that call `has_target_resource` or nest `input.document[_]` inside `input.document[i]`, scale roughly as O(N²) in document count. With enough callers this pushes individual rules well past the 60-second slow-rule threshold, and in the worst case drives total scan time over the outer 1-hour kill limit, producing a full failure instead of partial results.

## Changes

**Module deduplication**

When the same module resource is re-reached through different call chains and resolves to identical attributes (same type, name, definition line, and resolved configuration), only one synthetic document is emitted into `input.document`. Additional callers are recorded as *extras* in a side map instead of adding redundant documents.

**Finding expansion**

After OPA evaluation, findings on a deduplicated document are cloned, one copy per extra caller, each carrying the correct `ModuleCallChain` and `FileID`. This preserves per-caller fingerprints and file attribution so no coverage is lost.

**Content key**

Dedup uses a SHA-256 digest of the JSON-serialized resolved attributes. The key also includes the module address, so distinct module instances within a single configuration (`module.a` vs `module.b`, `count`/`for_each` indices) always get distinct keys and are never merged. Two callers that pass different variable values likewise produce different keys and emit separate documents.

## Author Checklist

## QA Instruction

1. `go test ./pkg/engine/...` — all existing module resolution tests pass unchanged. `TestInspect_TwoCallersGetDistinctCallChains` confirms per-caller findings are preserved (N callers → N findings via cloning); `TestInspect_AggregateRuleCardinalityPreservedWithinConfig` confirms a count-based rule still sees every distinct module instance in a configuration.
2. Scan a Terraform repo that uses the same local module from multiple stacks with identical inputs. Check scanner logs for the `Instantiated … unique OPA docs … deduplicated callers` line and confirm the unique-doc count is lower than the total caller count.
3. Verify findings still appear attributed to each call-site (distinct `moduleCallChain` per finding).

## Blast Radius

Affects the scan pipeline only — specifically module instantiation during Terraform scans. No change to rule assets, CLI flags, SARIF output schema, or non-Terraform platforms. Repositories that do not use local Terraform modules see no behaviour change.

## Additional Notes

This is a targeted mitigation for large-repo scan timeouts. It reduces document count, and therefore O(N²) rule eval time, without changing rule logic or the per-rule timeout model.

**Scope of dedup (cardinality).** Per-resource rules (the vast majority, including the S3 / `has_target_resource` family this targets) are unaffected: the finding fires on the primary document and is cloned per caller. Aggregate rules that count resources are also unaffected within a configuration, because distinct module instances keep distinct documents (covered by `TestInspect_AggregateRuleCardinalityPreservedWithinConfig`). The only behavior change is for a hypothetical rule that counts the *same* module resource reached identically from multiple independent root configurations — counting across separate Terraform states is not a meaningful check, and no bundled rule does this (the only `count()` rules aggregate within a single document).

### DIFF
--- a/pkg/engine/module_resolve.go
+++ b/pkg/engine/module_resolve.go
@@ -7,8 +7,6 @@ package engine
 
 import (
 	"context"
-	"crypto/sha256"
-	"encoding/hex"
 	"encoding/json"
 	"path/filepath"
 	"strconv"
@@ -18,6 +16,7 @@ import (
 	"github.com/DataDog/datadog-iac-scanner/pkg/model"
 	tfmodules "github.com/DataDog/datadog-iac-scanner/pkg/parser/terraform/modules"
 	"github.com/DataDog/datadog-iac-scanner/pkg/parser/terraform/tfeval"
+	"github.com/cespare/xxhash/v2"
 )
 
 // extraCallerInfo records a deduplicated module caller so its findings can be
@@ -260,17 +259,14 @@ func newInstanceFileMetadata(fm *model.FileMetadata, id, callChain string) *mode
 	return &clone
 }
 
-// resourceAttrKey returns a SHA-256 hex digest of a resolved resource's attributes,
-// used as the content key for dedup. SHA-256 matches the hash strength used by
-// GetDatadogFingerprintHash and avoids collisions from a weaker polynomial hash.
+// resourceAttrKey hashes resolved attributes for in-scan content dedup only (not fingerprints).
 func resourceAttrKey(r *tfeval.ResolvedResource) string {
 	attrs := tfeval.AttributesToDocument(r)
 	b, err := json.Marshal(attrs)
 	if err != nil {
 		return ""
 	}
-	sum := sha256.Sum256(b)
-	return hex.EncodeToString(sum[:])
+	return strconv.FormatUint(xxhash.Sum64(b), 16)
 }
 
 // callChainKey is repo-relative outer caller + "|" + module address (no line numbers, to keep fingerprints stable).
--- a/pkg/engine/module_resolve.go
+++ b/pkg/engine/module_resolve.go
@@ -221,8 +221,12 @@ func instantiatedDocs(
 		cck := callChainKey(r, repoPath)
 		docID := strings.Join([]string{fm.ID, cck, r.Name}, "\x00")
 
-		// Dedupe by resource content: identical resolved attributes from different callers collapse
-		// to a single OPA doc (saving eval cost on quadratic rules); findings are cloned after eval.
+		// Dedupe by resource content: identical resolved attributes reached through different
+		// call chains collapse to a single OPA doc (saving eval cost on quadratic rules); findings
+		// are cloned per caller after eval. The key includes ModuleAddress, so distinct module
+		// instances within one configuration (module.a vs module.b, count/for_each indices) always
+		// get distinct keys and are never merged. Cardinality is therefore preserved within a
+		// configuration; only the same module resource re-reached from another root collapses.
 		contentKey := strings.Join([]string{abs, r.ModuleAddress, r.Type, r.Name, strconv.Itoa(r.DefLine), resourceAttrKey(r)}, "\x00")
 		if primaryDocID, duplicate := seen[contentKey]; duplicate {
 			extras[primaryDocID] = append(extras[primaryDocID], extraCallerInfo{callChain: cck, docID: docID})
--- a/pkg/engine/module_resolve.go
+++ b/pkg/engine/module_resolve.go
@@ -7,6 +7,9 @@ package engine
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
 	"path/filepath"
 	"strconv"
 	"strings"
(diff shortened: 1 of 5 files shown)


whitemerch:
@codex review

pkg/engine/module_resolve_integration_test.go
```diff
@@ -198,6 +198,94 @@ resource "aws_s3_bucket" "this" {
...
+	}
+}
+`
+
+// TestInspect_AggregateRuleCardinalityPreservedWithinConfig verifies that an
+// aggregate rule that counts resources still sees every distinct module instance
+// within a single configuration. Two module calls in one root resolve to the same
+// attributes but distinct module addresses, so content-dedup keeps both documents
+// and the count-based rule fires, no cardinality is lost.
+func TestInspect_AggregateRuleCardinalityPreservedWithinConfig(t *testing.T) {
```
MarshalX: it uses `bucket_a` and `bucket_b`. module address is part of the key. not sure that this test tests deduplication

pkg/engine/module_resolve.go
```diff
@@ -215,6 +260,19 @@ func newInstanceFileMetadata(fm *model.FileMetadata, id, callChain string) *mode
...
+// resourceAttrKey returns a SHA-256 hex digest of a resolved resource's attributes,
+// used as the content key for dedup. SHA-256 matches the hash strength used by
+// GetDatadogFingerprintHash and avoids collisions from a weaker polynomial hash.
+func resourceAttrKey(r *tfeval.ResolvedResource) string {
+	attrs := tfeval.AttributesToDocument(r)
+	b, err := json.Marshal(attrs)
+	if err != nil {
+		return ""
+	}
+	sum := sha256.Sum256(b)
```
MarshalX: why `sha256` has been chose here? why not faster `xxhash` like in prev PRs?

pkg/engine/module_resolve.go
```diff
@@ -180,15 +217,23 @@ func instantiatedDocs(
...
+		// are cloned per caller after eval. The key includes ModuleAddress, so distinct module
+		// instances within one configuration (module.a vs module.b, count/for_each indices) always
+		// get distinct keys and are never merged. Cardinality is therefore preserved within a
+		// configuration; only the same module resource re-reached from another root collapses.
+		contentKey := strings.Join([]string{abs, r.ModuleAddress, r.Type, r.Name, strconv.Itoa(r.DefLine), resourceAttrKey(r)}, "\x00")
+		if primaryDocID, duplicate := seen[contentKey]; duplicate {
+			extras[primaryDocID] = append(extras[primaryDocID], extraCallerInfo{callChain: cck, docID: docID})
+			continue
+		}
+		seen[contentKey] = docID
```
MarshalX, line 234: Hot take from our Optimus Prime: 

> The core concern is that the dedup decision is made before OPA evaluation, but the finding-expansion can only clone findings that already exist after evaluation. This breaks any rule that aggregates over input.document.

<details>
<summary>Details</summary>

The keying (pkg/engine/module_resolve.go:230) collapses identical module resources reached from different roots into a single OPA document. expandModuleFindings (pkg/engine/inspector.go:413) then clones whatever findings exist per extra caller — but it can't synthesize a finding that didn't fire, nor suppress one that shouldn't have.

Concrete failure modes:
- count([r | input.document[_].resource.aws_s3_bucket[_]]) >= 2 — stops firing when two identical cross-root instances collapse to one doc (under-count → false negative).
- count(...) == 1 — can start false-positive, then get incorrectly cloned (over-report).

Suggested directions: gate the optimization to rules proven per-document/per-resource, or preserve logical multiplicity in a form arbitrary Rego can still observe (a side-channel caller count isn't enough unless every affected rule reads it).
</details>

Megatron take:

> Content-dedup collapses identical module resources reached from different root configurations into one input.document entry before OPA eval. expandModuleFindings clones findings post-eval, so it preserves per-resource rules but cannot preserve aggregate semantics: a custom rule doing count(input.document[_].resource...), a cross-document comparison, or an absence check will under- or over-count when collapse happens. This is safe for all bundled rules (none aggregate across roots) but is an unguarded assumption about rule content.

Should be okay now, but with custom rules idk?

pkg/engine/module_resolve_integration_test.go
whitemerch: Yes true, funny thing is that i was first testing with the same test, and ended up diverging to rather testing this case. I will add one in addition to this one

pkg/engine/module_resolve.go
whitemerch: Indeed, I copied the datadog fingerprint hash function initially, which is why I ended putting sha256. But you are totally correct, it's totally valid and it's an oversight. I will correct it

pkg/engine/module_resolve.go
whitemerch: Yes I will see what I can do, knowing we will shortly support custom rules

whitemerch: On cross-document cardinality: I looked at the default-rules corpus, the rules that scan input.document globally are either count == 0 (presence checks, dedup-safe) or pair different resource types by reference. Nothing does cross-root “how many identical module instances exist across independent stacks,” and that scenario is pretty niche for custom rules too, given also how custom rules are niche too, and combine it with modules, this is a pretty rare case for something I tried to find a solution for without success

MarshalX (approved):
LGTM!

PR #221, opened 2026-06-29
Optimize privilege-escalation rules by pre-indexing IAM permission sets

whitemerch:
## Motivation

The `*_unrecommended_permission_policy_scenarios` helpers in `assets/libraries/common.rego` were performing a full `input.document[_]` scan for **every principal** (role, user, group) they were called on. With Terraform module instantiation in place, `input.document` can contain many synthetic module documents, making each helper call O(principals × docs × policies). Because ~64 near-identical privilege-escalation rules call these helpers independently, the total cost was O(64 × principals × docs × policies) per scan, the primary contributor to multi-minute scan times on module-heavy repositories.

Benchmarks on a simulated module-inflated input (3,000 roles, 750 policies across 50 documents) measured **3,216 ms → 45 ms per rule** (~71× speedup).

## Changes

**Rule library (`assets/libraries/common.rego`)**

Three helper functions (`role_unrecommended_permission_policy_scenarios`, `user_unrecommended_permission_policy_scenarios`, `group_unrecommended_permission_policy_scenarios`) are replaced with an indexed design.

Two new helpers support the index:

`_to_action_array` normalises an `Action` field (string or array) to an array for uniform iteration.

`_policy_wildcard_actions` returns the set of lowercased `Allow`-on-`"*"` actions for a given policy resource object. OPA memoizes complete-rule calls per unique argument, so `json_unmarshal` is paid at most once per unique policy within a query.

Three complete set rules (`_role_dangerous_pairs`, `_user_dangerous_pairs`, `_group_dangerous_pairs`) pre-compute `[principalName, permission]` pairs across both scenarios (inline policies and policy attachments) for all principals at once. OPA evaluates and caches complete rules on first access; all subsequent lookups within the same query are O(1) set membership checks. This is the key difference from an incremental/partial-rule approach, which OPA specializes per bound key and does not cache across callers.

## Author Checklist

## QA Instruction

1. `opa test assets/libraries/` — all 324 existing tests must pass (no new tests are needed since the function signatures are unchanged and existing tests cover both scenarios).
2. `go test ./pkg/engine/... -timeout 120s` — engine tests must pass.
3. Optionally benchmark one slow rule before and after:

```
opa bench \
  -d assets/libraries/common.rego \
  -d assets/libraries/terraform.rego \
  -d <path-to-role_with_privilege_escalation_by_actions_iam_AttachRolePolicy/query.rego> \
  --input <module-heavy-input.json> \
  'data.datadog.DatadogPolicy'
```

Verify findings count is identical between the patched and unpatched library.

## Blast Radius

DataDog IaC Scanner only. The change is limited to three helper functions in `assets/libraries/common.rego`.

## Additional Notes

The ~64 `role_with_privilege_escalation_by_actions_*`, `user_with_privilege_escalation_by_actions_*`, and `group_with_privilege_escalation_by_actions_*` rules in the default rule corpus all benefit from this change without modification.

### DIFF
--- a/assets/libraries/common.rego
+++ b/assets/libraries/common.rego
@@ -722,91 +722,82 @@ unrecommended_permission_policy(resourcePolicy, permission) if {
 	equalsOrInArray(statement.Action, lower(permission))
 }
 
-group_unrecommended_permission_policy_scenarios(targetGroup, permission) if {
-	# get the IAM group policy
+# Normalises Action (string or array) to an array for uniform iteration.
+_to_action_array(x) := [x] if is_string(x)
+
+_to_action_array(x) := x if is_array(x)
+
+# Returns the set of lowercased Allow-on-"*" actions for a policy resource object.
+# Complete rules are memoized by OPA per unique argument, so json_unmarshal
+# is paid at most once per unique policy object within a query.
+_policy_wildcard_actions(policyObj) := {lower(a) |
+	policy := json_unmarshal(policyObj.policy)
+	statement := get_statement(policy)[_]
+	is_allow_effect(statement)
+	equalsOrInArray(statement.Resource, "*")
+	a := _to_action_array(statement.Action)[_]
+}
+
+# Pre-computed sets of [principalName, permission] pairs, one per principal type.
+#
+# Using complete set rules (not incremental/partial rules) guarantees that OPA
+# evaluates and caches each set ONCE per top-level query, regardless of how
+# many times the caller loops over principal names. The 64 privilege-escalation
+# rules that call *_unrecommended_permission_policy_scenarios share these sets
+# and avoid repeating an O(roles × docs × policies) document scan per role.
+
+_role_dangerous_pairs := {[roleName, permission] |
+	rolePolicy := input.document[_].resource.aws_iam_role_policy[_]
+	roleName := split(rolePolicy.role, ".")[1]
+	permission := _policy_wildcard_actions(rolePolicy)[_]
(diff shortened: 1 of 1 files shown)


assets/libraries/common.rego
```diff
@@ -722,91 +722,82 @@ unrecommended_permission_policy(resourcePolicy, permission) if {
...
-	group == targetGroup
+# Normalises Action (string or array) to an array for uniform iteration.
+_to_action_array(x) := [x] if is_string(x)
 
-	# confirm that policy associated is unrecommended
-	policy := split(attachment.policy_arn, ".")[1]
-
-	policies := {"aws_iam_role_policy", "aws_iam_user_policy", "aws_iam_group_policy", "aws_iam_policy"}
-	resourcePolicy := input.document[_].resource[policies[_]][policy]
+_to_action_array(x) := x if is_array(x)
```
ChouraquiBen, line 728: Question: That will make queries fail if x is neither an array or a string, is that ok or should there be a default case?

ChouraquiBen (approved):
LGTM

assets/libraries/common.rego
whitemerch: Good question, in practice it's fine as-is. IAM Action is only ever a string or array per the AWS policy schema, and if it's something else (e.g. null in a malformed policy), `_to_action_array` is undefined and that statement is simply skipped in the comprehension, same outcome as having no actions. No query failure, no false positives. Happy to add else := [] if you prefer making that explicit.

PR #226, opened 2026-06-30
Gate Terraform local module instantiation behind a feature flag

whitemerch:
## Motivation

Local Terraform module instantiation changes scan behavior and adds runtime cost on large repositories. To unblock the v1.6.0 release while slow-rule performance work continues, module resolution should be opt-in rather than always on.

## Changes

**Feature flags**

Add `k9-iac-enable-local-module-eval`, defaulting to `false` in the local evaluator.

**Scan pipeline**

When the flag is off, skip `instantiateLocalModules` and `expandModuleFindings`. Module files are scanned as parsed HCL with unresolved `var.*` references, matching the pre-module-instantiation behavior.

**Tests**

Enable the flag explicitly in existing module integration tests. Add a regression test confirming the default-off path does not inject synthetic module documents.

## QA Instruction

1. `go test ./pkg/engine/ -run TestInspect -count=1`
2. Build the scanner and run a Terraform repo with local modules; confirm no synthetic module documents are added by default.
3. Enable `k9-iac-enable-local-module-eval` and confirm module instantiation findings are restored.

## Impact

CLI and server scans default to the prior module scanning behavior until the flag is enabled remotely.

### DIFF
--- a/cmd/scanner/scan.go
+++ b/cmd/scanner/scan.go
@@ -87,6 +87,12 @@ var scanAction = &cli.Command{
 			Value:  false,
 		},
 		&cli.BoolFlag{
+			Name:   "x-local-module-eval",
+			Hidden: true,
+			Usage:  "(experimental) resolve Terraform local module variables before scanning",
+			Value:  false,
+		},
+		&cli.BoolFlag{
 			Name:   "x-terraform-plan",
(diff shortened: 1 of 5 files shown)


MarshalX (approved):
LGTM!

PR #227, opened 2026-06-30
Speed up IAM privilege-escalation rules on large Terraform repos

whitemerch:
## Motivation

On large Terraform repositories, the IAM privilege-escalation rule family was a major scan bottleneck. Each rule rebuilt dangerous-permission sets by resolving `policy_arn` attachments with an O(attachments × documents) document rescan. This is a followup of https://github.com/DataDog/datadog-iac-scanner/pull/221, which improved these queries, but there was a bigger improvement possible.

## Changes

**IAM helpers (`common.rego`)**

Build `_policy_name_wildcard_perms` once per query in a single pass over managed policy resources. Attachment branches for roles, users, and groups now resolve `policy_arn` references via membership in that set instead of rescanning every document for each attachment.

**Tests**

Add regression coverage for inline policies, cross-document attachment resolution, and non-wildcard policy exclusion.

## Author Checklist

## QA Instruction

1. `opa test assets/libraries/ --v1-compatible`
2. Build the scanner and scan a large Terraform repo with many IAM roles and policy attachments.
3. Confirm privilege-escalation rules no longer appear in slow-rule logs and total scan time drops materially.

On `cloud-inventory` (29,788 files, full rule corpus): scan time dropped from ~8m to ~1m53s, with 0 slow IAM privilege-escalation rules (previously 19 rules at ~3 min each).

## Blast Radius

Affects Terraform IAM privilege-escalation rules that use `*_unrecommended_permission_policy_scenarios` in `common.rego`. No CLI or engine behavior changes.

## Additional Notes

N/A

### DIFF
--- a/assets/libraries/common.rego
+++ b/assets/libraries/common.rego
@@ -738,6 +738,18 @@ _policy_wildcard_actions(policyObj) := {lower(a) |
 	a := _to_action_array(statement.Action)[_]
 }
 
+_managed_policy_types := {"aws_iam_role_policy", "aws_iam_user_policy", "aws_iam_group_policy", "aws_iam_policy"}
+
+# [policyResourceName, wildcardAction] pairs across every managed policy type,
+# built in one document pass. Only policies granting an Allow on "*" contribute,
+# so this set stays small. The attachment branches below resolve a policy_arn
+# reference via membership here instead of rescanning every document per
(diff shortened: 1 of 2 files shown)


MarshalX (dismissed):
Awesome! 

Nit: we do have tests for role and user attachment, but not for group

MarshalX (approved):

PR #229, opened 2026-06-30
Add a shared CPU budget to fix negative scaling at high core counts

MarshalX:
## Motivation

A scan runs several worker pools, and they **nest**. The scanner fans out one goroutine *per service*, and inside each service the query-evaluation pool spins up its own `GOMAXPROCS` workers. There were five such pools in total (query eval, file reading, HCL body parsing, module-variable parsing, file-type detection), each independently sizing itself to roughly the core count, with no shared notion of how much CPU work was allowed to run at once.

The result: in-flight CPU-bound goroutines scaled as `Nservices × GOMAXPROCS`, well over 100 goroutines competing for 16 cores (on my laptop). Past a point, **adding cores made scans slower, not faster** (negative scaling): the pools oversubscribed the machine and the extra goroutines added scheduler churn, cache thrashing, and lock contention. A `GOMAXPROCS` sweep confirmed it, latency got worse going from 8 to 16 cores, and concurrent scans showed almost no throughput gain.

On top of that, the five pools were inconsistent: only the file reader clamped its worker count to `[4, 64]`; the module-variable pool ignored the auto-sizing helper and hardcoded `4`; the file-type analyzer is I/O-bound (it reads file content) but was capped at `GOMAXPROCS` while the otherwise-identical file reader fanned out to 64; and all five were hand-rolled copies of the same channel + `WaitGroup` + closer boilerplate.

JIRA Ticket: [K9CODESEC-2731](https://datadoghq.atlassian.net/browse/K9CODESEC-2731)

- Introduced a **process-wide shared CPU budget** (weighted semaphore sized to the available CPU count). Every CPU-bound pool (Rego eval, HCL parsing, module-variable parsing) acquires one slot per unit of work, so nesting no longer multiplies, at most `AvailableCPUs()` CPU-bound goroutines run at once regardless of how deep the pools stack.
- **Unified the five worker pools** behind one generic helper (`utils.ForEach`, errgroup-backed), replacing the duplicated fan-out/`WaitGroup`/closer boilerplate.
- **Separated CPU-bound from I/O-bound work**: I/O-bound pools (file reading, file-type detection) do *not* consume the CPU budget (blocking on disk shouldn't hold a CPU slot) and fan out wider than the core count via shared `IOMinWorkers`/`IOMaxWorkers` (`[4, 64]`) constants in `workers.go`. Fixes the analyzer being throttled to `GOMAXPROCS`.
- Removed the hardcoded `4` in module-variable parsing and made that pass **deterministic** (sorted iteration order).
- Single source of truth for core count (`utils.AvailableCPUs()`) used everywhere instead of ad-hoc `GOMAXPROCS` reads.
- Added unit tests for the new worker helper and CPU budget; added a benchmark harness for the `GOMAXPROCS` sweep.

## Author Checklist

## QA Instruction

The change is internal (concurrency), so the observable behavior must be **identical** while scaling improves:

- **Correctness:** scan a repo before and after this PR; the findings count and SARIF output must be unchanged.
- **Scaling:** run the same scan at `GOMAXPROCS=4`, `8`, `16` (e.g. `GOMAXPROCS=16 datadog-iac-scanner scan -p <repo> -o <out>`). Higher core counts should be the same or faster — in particular 16 cores should no longer be slower than 8 (the previous negative-scaling regression).
- **Race safety:** `go test -race ./pkg/engine/... ./pkg/utils/ ./pkg/analyzer/ ./pkg/parser/terraform/modules/` passes clean.

## Blast Radius

Datadog IaC Scanner only. The change is confined to the scanner's internal worker/concurrency layer (`pkg/utils`, `pkg/engine`, `pkg/analyzer`, `pkg/parser/terraform/modules`, `pkg/engine/provider`). No public API, CLI flag, rule, or output-format change. No impact on other services or documentation.

[K9CODESEC-2731]: https://datadoghq.atlassian.net/browse/K9CODESEC-2731?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/engine/inspector.go
+++ b/pkg/engine/inspector.go
@@ -296,9 +296,14 @@ func (c *Inspector) Inspect(
 	// Must run after module mutations (which suppress module bodies in place).
 	combinedFiles := files.Combine(ctx, false)
 
-	// Step 1: Parse Terraform modules
+	// Step 1: Parse Terraform modules. A genuine per-file HCL parse failure is
+	// non-fatal (logged, scan continues), but a context cancellation must abort
+	// the scan rather than proceed with partial module data.
 	parsedModules, err := tfmodules.ParseTerraformModules(ctx, c.fsys, files, c.numWorkers)
 	if err != nil {
+		if ctxErr := ctx.Err(); ctxErr != nil {
+			return nil, ctxErr
+		}
 		contextLogger.Warn().Err(err).Msg("Failed to parse Terraform modules")
 	}
 	contextLogger.Info().Msgf("Found %d modules", len(parsedModules))
--- a/pkg/engine/inspector.go
+++ b/pkg/engine/inspector.go
@@ -308,9 +308,14 @@ func (c *Inspector) Inspect(
 	}
 	contextLogger.Info().Msgf("Found %d modules", len(parsedModules))
 
-	// Step 2: Enrich modules with parsed variables
+	// Step 2: Enrich modules with parsed variables. As with Step 1, a context
+	// cancellation must abort the scan rather than proceed with partial module
+	// data; per-module parse failures are non-fatal and handled internally.
 	rootDir := c.repoPath
-	enrichedModules := tfmodules.ParseAllModuleVariables(ctx, c.fsys, parsedModules, rootDir)
+	enrichedModules, err := tfmodules.ParseAllModuleVariables(ctx, c.fsys, parsedModules, rootDir)
+	if err != nil {
+		return nil, err
+	}
 
 	queries := c.getQueriesByPlat(platforms)
 
@@ -359,13 +364,10 @@ func (c *Inspector) executeQueries(
 			results[i] = c.evalQuery(ctx, scanID, filesMap, payloads, queries, i, enrichedModules, baseStores)
 			return nil
 		})
+	// The closure never returns a non-nil error itself, so ForEach only reports
+	// context cancellation here; surfacing it keeps a canceled scan from being
+	// reported as a successful scan with partial/empty results.
 	if err != nil {
-		return vulnerabilities, err
-	}
-
-	// A canceled scan must surface the cancellation rather than be reported as a
-	// successful scan with partial/empty results.
-	if err := ctx.Err(); err != nil {
 		return vulnerabilities, err
 	}
 
--- a/pkg/engine/inspector.go
+++ b/pkg/engine/inspector.go
@@ -224,75 +224,56 @@ func getPlatformLibraries(ctx context.Context, queriesSource source.QueriesSourc
 	return platformLibraries
 }
 
-type InspectionJob struct {
-	queryID int
-}
-
 type QueryResult struct {
 	vulnerabilities []model.Vulnerability
 	err             error
 	queryID         int
 }
 
-// This function creates an inspection task and sends it to the jobs channel
-func (c *Inspector) createInspectionJobs(jobs chan<- InspectionJob, queries []model.QueryMetadata) {
-	defer close(jobs)
-	for i := range queries {
-		jobs <- InspectionJob{queryID: i}
+// evalQuery loads and evaluates a single query, returning its result. A load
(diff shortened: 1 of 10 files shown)


pkg/engine/inspector.go
```diff
@@ -357,50 +338,35 @@ func (c *Inspector) executeQueries(
...
+	// draws from the process-wide CPU budget: when this scan runs as one of N
+	// concurrent per-platform services, all their query pools share the same
+	// budget and cannot oversubscribe the machine. Results land in an
+	// index-aligned slice (no shared mutable state between workers); the actual
+	// aggregation happens serially below.
+	results := make([]QueryResult, len(queries))
+	err := utils.ForEach(ctx, queries, utils.PoolOptions{Workers: c.numWorkers, CPUBound: true},
+		func(ctx context.Context, _ model.QueryMetadata, i int) error {
+			results[i] = c.evalQuery(ctx, scanID, filesMap, payloads, queries, i, enrichedModules, baseStores)
+			return nil
```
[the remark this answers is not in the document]
MarshalX: fixed in https://github.com/DataDog/datadog-iac-scanner/pull/229/commits/0ada3090961c1778ff3257536acc216bb22023a7

pkg/parser/terraform/modules/modules.go
```diff
@@ -117,38 +113,22 @@ func parseHCLBodies(ctx context.Context, files model.FileMetadatas, numWorkers i
...
-					contextLogger.Warn().Msgf("Skipping file %s due to HCL parse errors: %s", file.FilePath, diags.Error())
-					continue
-				}
-				if body == nil {
-					contextLogger.Error().Msgf("Unexpected body type in %s", file.FilePath)
-					continue
-				}
-				bodies[i] = body
+	// HCL parsing is CPU-bound, so the pool draws from the shared CPU budget.
+	_ = utils.ForEach(ctx, tfFiles, utils.PoolOptions{Workers: numWorkers, CPUBound: true},
```
whitemerch: Should we propagate `ctx.Err()` here? Right now a canceled scan can finish with partial HCL/module results and no error surfaced to the caller

pkg/utils/workers.go
```diff
@@ -5,13 +5,143 @@
...
+// hand-rolled "jobs channel + N goroutines + WaitGroup + closer" pattern that
+// was duplicated across the engine, analyzer and Terraform-module parsers.
+//
+// fn receives the original slice index so callers can write results into a
+// preallocated, index-aligned output slice without sharing mutable state.
+//
+// If opts.CPUBound is set, each invocation of fn is wrapped in CPUBound so the
+// work draws from the shared process-wide CPU budget. The first non-nil error
+// cancels the remaining work and is returned.
+func ForEach[T any](ctx context.Context, items []T, opts PoolOptions, fn func(ctx context.Context, item T, index int) error) error {
```
whitemerch, line 139: Worth a brief comment on why we need both `g.SetLimit` and `CPUBound`: errgroup limits goroutines per pool, semaphore shares the process-wide CPU budget across nested pools...

whitemerch (dismissed):

pkg/utils/workers.go
MarshalX: addressed in https://github.com/DataDog/datadog-iac-scanner/pull/229/commits/1c7a279ed9a84e01b84c3d5988dfc47e5f461354

pkg/parser/terraform/modules/modules.go
MarshalX: addressed in https://github.com/DataDog/datadog-iac-scanner/pull/229/commits/1c7a279ed9a84e01b84c3d5988dfc47e5f461354

whitemerch (dismissed):

whitemerch (approved):
Re-verified, good to go

MikaYuoadas (commented):
Great changes 🔥

PR #231, opened 2026-07-02
Remove embedded Rego libraries and rely on backend delivery

whitemerch:
## Motivation

Rego libraries were previously embedded in the scanner binary via `go:embed` and served as a fallback when the backend was unavailable. The libraries are now managed and pushed to the backen, making the embedded copy redundant and a source of drift. This PR removes the embedded libraries entirely and makes the backend the sole delivery path, with `--libraries-path` as an explicit local override for development or custom deployments.

## Changes

**Library delivery**

Removed all files under `assets/` (18 `.rego` files, `common.json`, and 4 `*_test.rego` files). The `assets` package no longer embeds anything. The OPA unit tests for library helper functions have been ported to the default-rules repository.

**Backend source**

`DatadogSource.GetQueryLibrary` now normalizes platform names to lowercase on both storage and lookup, fixing a case mismatch (`cloudFormation` vs `cloudformation`) that was previously masked by the embedded fallback. The auto-fallback to embedded libraries in `NewDatadogSource` is removed; the only library fallback path is when `--libraries-path` is explicitly set.

**Filesystem source**

`FilesystemSource.GetQueryLibrary` now reads `.rego` and `.json` files directly from disk instead of delegating to the embedded assets package.

**Scan wiring**

`pkg/scan/scan.go` only adds a local library fallback source when `ChangedDefaultLibrariesPath` is true (i.e. `--libraries-path` was explicitly passed). Without that flag, the scanner always fetches libraries from the backend.

**CLI commands**

`list-queries`, `test-rules`, and custom Rego evaluation (`custom.go`) all use `DatadogSource` for library access instead of the embedded filesystem.

**E2E tests**

Removed `useLocalLibraries` helper and the `testdata/libraries/` snapshot. E2E tests now fetch libraries from the public backend endpoint (`/api/v2/static-analysis/iac/libraries`), which requires no authentication.

**Tooling**

Removed `.github/scripts/update_versions/`, the `update_software_versions` workflow, `rules_generator/classes/CodeProcessor.py` methods that read/wrote `common.json` now raise `NotImplementedError`  (are they even used? Anyways if so they would need to be ported separetly, and in a different way)

## Author Checklist

## QA Instruction

Build the binary and run a scan against any Terraform fixture, libraries are now fetched from the backend automatically:

```bash
go build -o bin/datadog-iac-scanner ./cmd/scanner
./bin/datadog-iac-scanner scan -p <path-to-terraform-dir>
```

Confirm no `library not found` errors appear. To test the explicit local override path:

```bash
./bin/datadog-iac-scanner serve --libraries-path <path-to-local-libraries>
```

## Blast Radius

Affects the scanner binary only. The backend library endpoint is public and requires no authentication

## Additional Notes

Companion PR in default-rules: https://github.com/DataDog/datadog-iac-scanner-default-rules/pull/71 (ports the OPA library tests, fixes `tool.sh test` to read libraries from disk, bumps scanner dependency to v1.7.0).


pkg/engine/source/datadog.go
```diff
@@ -53,7 +42,7 @@ func WithWantedCloudProviders(providers []string) DatadogSourceOption {
 	}
 }
 
-// WithLibraryFallback overrides the QueriesSource used when the backend library endpoint is unreachable.
+// WithLibraryFallback sets a QueriesSource to use for libraries when the user supplies --libraries-path.
 func WithLibraryFallback(source QueriesSource) DatadogSourceOption {
```
ChouraquiBen: suggestion: `libraryFallback` sounds like an error has happened (that was the case), but is now used for local libraries (fallback usage is expected behaviour).
Would `localLibraries` work?

pkg/scan/custom_scan_test.go
```diff
@@ -2,13 +2,54 @@ package scan
...
+		return source.RegoLibraries{
+			LibraryCode:      "package generic.k8s\nimport rego.v1\n",
+			LibraryInputData: "{}",
+		}, nil
+	default:
+		return source.RegoLibraries{}, fmt.Errorf("no stub library for platform %q", platform)
+	}
+}
+
+func testLibSource() source.QueriesSource { return &stubLibSource{} }
```
ChouraquiBen, line 51: Question: Why not use the `&stubLibSource` directly in the functions' arguments

rules_generator/classes/CodeProcessor.py
```diff
@@ -34,40 +33,13 @@ def __write_file(self, path: Path, content: str) -> None:
...
+        raise NotImplementedError("common.json has been removed; library content is now managed in the default-rules repository")
 
     def write_common(self, common: Any) -> None:
-        try:
-            with open("assets/libraries/common.json", "w") as f:
-                json.dump(common, f, indent=2, ensure_ascii=False)
-                print("Succesfully wrote common.json")
-        except Exception as e:
-            print(f"Failed to write common.json: {e}")
+        raise NotImplementedError("common.json has been removed; library content is now managed in the default-rules repository")
```
ChouraquiBen, line 42: nit: Ehhhh, this is a bit of a shame, to leave that broken without a plan. Could you please create a ticket for the clean-up?

pkg/engine/source/datadog.go
```diff
@@ -93,12 +83,12 @@ func (s *DatadogSource) loadLibraries(ctx context.Context) {
 	s.librariesOnce.Do(func() {
 		libs, err := s.client.GetLibraries(ctx)
 		if err != nil {
-			// backend unreachable; embedded fallback will be used
+			s.librariesErr = err
```
ChouraquiBen: Remark: If the fetch fails once, even for a temporary server error, the function won't run again for a retry and propagate the error

pkg/engine/source/datadog.go
ChouraquiBen: Or `libraryOverride` 🤔

ChouraquiBen (dismissed):
A few comments 🙏

rules_generator/classes/CodeProcessor.py
whitemerch: I mean is this file ever gonna be used anymore? It seems like it was what was used to generate the module special cases inside the rules, but seems we are going to move away from it, is it really useful? I created a [ticket](https://datadoghq.atlassian.net/browse/K9VULN-16564) to remove them

pkg/scan/custom_scan_test.go
whitemerch: Because it is called 24 times in the file

rules_generator/classes/CodeProcessor.py
ChouraquiBen: Yeah, I think that removing it is probably what we need here

ChouraquiBen (approved):
LGTM

PR #234, opened 2026-07-03
Reduce redundant repository walks on polyglot scans

whitemerch:
> [!TIP]

## Motivation

On repositories that mix many IaC platforms (Terraform, Kubernetes YAML, Dockerfiles, CI/CD workflows, etc.), the scanner used to walk the filesystem repeatedly: once during platform detection, then again for each parser service. That duplicated I/O and re-read the same files many times during prepare. This PR restructures the prepare phase so a single walk feeds every parser.

## Changes

**Before**

The scan pipeline treated platform detection and source preparation as separate concerns. The analyzer walked the tree to decide which platforms to load, then each parser service walked the same tree independently to find, open, and parse its files. Helm charts could be rendered per service. Platform attribution for parsed documents often re-ran content-based classification even though the analyzer had already classified the same path.

**After**

Platform detection and inventory collection happen in one analyzer pass. That pass now produces everything downstream needs: the file list to scan, Helm chart roots, a path-to-platform map, and cached bytes for content-classified files. The prepare phase consumes that inventory instead of walking again. When all services share the same disk provider, one orchestrated pass renders each chart once, reads each file once, and routes content to the parsers that support that extension. The query engine phase is unchanged.

**Behavioral notes**

1. Scan output should match the previous pipeline; a parity test compares document fingerprints between the shared walk and the legacy per-service walk.
2. Polyglot repos should see less filesystem churn during prepare; the win scales with the number of active parser services.
3. Plain single-document YAML without references skips the resolver rewrite step when it cannot change the document.
4. Server and in-memory scan modes are unaffected: they do not use the shared disk walk and keep the existing per-service prepare path.

## Benchmark

Measured on a large polyglot monorepo (~346k files discovered, 2,135 queries from `datadog-iac-scanner-default-rules` @ `main`). Both runs used `--profiling CPU`, parallel parsing off, and the same local rules/libraries paths. **Before** = `main` branch scanner; **After** = this PR.

| Metric | `main` | This PR | Delta |
|---|---:|---:|---|
| Wall-clock scan time | 8m 30s | 4m 19s | 49% faster |
| `file_type_analyzer` CPU | 302.4s | 146.2s | 52% less |
| `prepare_sources` CPU | 583.5s | 181.9s | 69% less |
| `start_scan` CPU (query eval) | 175.2s | 224.1s | not targeted by this PR |
| `get_queries` CPU | 0.36s | 0.32s | — |
| `generate_report` CPU | 1.41s | 0.15s | — |
| Files scanned | 28,483 | 27,468 | same inventory semantics |
| Violations | 9,995 | 9,995 | identical |

The largest gains are in `file_type_analyzer` and `prepare_sources`, which is where redundant walks and re-reads were removed. Query evaluation (`start_scan`) is intentionally unchanged and can vary between runs, this is handled by this [PR](https://github.com/DataDog/datadog-iac-scanner-default-rules/pull/76)
 
## Author Checklist

## QA Instruction

1. `go test ./pkg/analyzer/... ./pkg/engine/provider/... ./pkg/runner/... ./pkg/scan/... ./pkg/scanner/... ./pkg/resolver/file/...`
2. `make build` then scan a mixed fixture tree, e.g. `./bin/datadog-iac-scanner scan -p <polyglot-path> -o /tmp/out`
3. Compare findings and file counts against `main` on the same path; results should be equivalent.
4. Optional: run with `--profiling CPU` and confirm `prepare_sources` time drops on large polyglot repos.

## Blast Radius

CLI filesystem scans are the primary beneficiary. The HTTP server content-push path and scans that do not share a single filesystem provider continue to use the legacy per-service prepare flow.change.

## Additional Notes


pkg/scanner/scanner.go
```diff
@@ -26,6 +26,16 @@ func PrepareAndScan(
...
 	metrics.Metric.Start("prepare_sources")
+
+	if fsp, ok := runner.SharedWalkProvider(services); ok {
+		err := runner.PrepareSharedWalk(ctx, fsp, services, scanID, openAPIResolveReferences, maxResolverDepth)
+		metrics.Metric.Stop()
+		if err != nil {
+			return err
+		}
+		return StartScan(ctx, scanID, services)
+	}
```
MarshalX: Should we respect `IaCEnableKicsParallelFileParsing` flag here?

pkg/runner/prepare_test.go
```diff
@@ -0,0 +1,166 @@
...
+	require.NoError(t, err)
+	fp := make(map[string]int, len(files))
+	for _, f := range files {
+		key := fmt.Sprintf("%s|%s|%s|%s|%d", f.Platform, f.Kind, f.FilePath, f.HelmID, len(f.OriginalData))
+		fp[key]++
+	}
+	return fp
+}
+
+func TestPrepareSharedWalk_MatchesPerService(t *testing.T) {
```
MarshalX, line 98: Should we add a test case where `SetPrebuiltWalk` will be called? Which is what actually gonna be used in case of CLI run

pkg/runner/service.go
```diff
@@ -254,6 +256,26 @@ func getContent(rc io.Reader, data []byte, maxSizeMB int, filename string) (*Con
...
+		countLines++
+	}
+	return &Content{
+		Content:        &copied,
+		CountLines:     countLines,
+		IsMinified:     minified.IsMinified(filename, copied),
+		CountResources: GetCountTerraformResources(copied),
+	}, nil
+}
+
```
MarshalX, line 285: Looks like this new counting algorithm does not aligns well with legacy `getContent` from `service.go`. The old one gives +1 for every read chunk. Even when file ends in `\n`. 

because of this we do have a line number drift. The avg delta is about -1. But could be worth in case of huge files. Like ~3mb file without new lines is -2 lines in comparison to old `getContent` (because of chunks) 

the problem is that we do have cached and unchached files. which do calculates the same thing differently 

important: the **goal here is parity between**, the `contentFromBytes` looks more right implementation and old `getContent` is buggy one

pkg/scan/utils.go
```diff
@@ -238,24 +240,3 @@ func extractPathType(paths []string) (regular, kuberneter []string) {
...
-				files++
-			}
-
-			return nil
-		}); err != nil {
-			contextLogger.Error().Msgf("failed to walk path %s: %s", path, err)
-		}
-	}
-	return files
-}
```
MarshalX, line 261: That was my most beloved test of the entire project

R.I.P hero, /F

pkg/analyzer/analyzer.go
MarshalX, line 97: _Disclaimer: This is not mistake. I am commenting the right line number (it is out of this PR)_ 

Problem: `terraform.tfvars` and `*.auto.tfvars` and etc. files are **no longer get scanned**

inside `possibleFileTypes` key for `tfvars` is **no-dot** string. 

On the `main` branch this was hidden because the actual scan came from the provider`s own walk

MarshalX (commented):
The real concerns are tfvars files skip and parity of line numbers. Worth addressing. The uncapped cache with no eviction from my tests not a huge. Literally size of json/yaml/etc files. Plus temporal x2 for copy in `contentFromBytes`. 

But, by spending a little of time, I was managed to create a bomb payload to achieve a huge memory consumption. But I do think this is out of scope for this PR (we do have other uncapped caches...) 

Image

MarshalX (approved):

PR #235, opened 2026-07-03
Add --libraries-path to the scan CLI

whitemerch:
## Motivation

`serve` already accepts `--libraries-path`; `scan` did not. After embedded libraries were removed, local rule scans need a way to point at Rego libraries on disk when we are testing

## Changes

Adds `--libraries-path` to `scan` and wires it to `ChangedDefaultLibrariesPath`, matching the existing `serve` behavior.

## QA Instruction

`make build` then run a scan with local queries and libraries, e.g. `./bin/datadog-iac-scanner scan -p <path> -q <rules-dir> --libraries-path <libs-dir>`.

## Blast Radius

CLI `scan` command only.

### DIFF
--- a/cmd/scanner/scan.go
+++ b/cmd/scanner/scan.go
@@ -111,6 +111,10 @@ var scanAction = &cli.Command{
 			Aliases: []string{"q"},
 			Usage:   "a list of query directories paths",
 		},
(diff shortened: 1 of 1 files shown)


barbiergaspard (approved):
LGTM!

PR #239, opened 2026-07-06
Add fetch-bundle + scan --offline-bundle-path for network-isolated scans

NathanGallet-dd:
## Summary

By default, `scan` always fetches the default ruleset and Rego libraries from the Datadog backend (`GetDefaultRuleset`/`GetLibraries`), and unconditionally merges server-side config overrides via `GetRemoteConfig` even when `--queries-path` is set for rules.

This PR adds a path to run `scan` with **zero** network calls:

- **`fetch-bundle`**: a new subcommand that runs on a trusted, network-enabled host. It fetches the default ruleset, Rego libraries, and the repo's merged remote configuration, and writes them to local files (`rules.json`, `libraries.json`, `config.yaml`) in an output directory.
- **`scan --offline-bundle-path <dir>`**: loads that bundle instead of calling the backend — `pkg/datadog.LocalFileClient` (new) serves the pre-fetched ruleset/libraries, and the merged config is read directly via `config.ParseConfig` instead of `config.ReadConfiguration`'s network-backed `WithDatadog` option.

The consuming side reuses the existing, already-production `scan.WithQuerySourceFactory` seam (previously used by the HTTP server for content-push scans) — no changes needed to the core scan pipeline.

### DIFF
--- a/cmd/scanner/fetch_bundle.go
+++ b/cmd/scanner/fetch_bundle.go
@@ -1,0 +1,96 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/DataDog/datadog-iac-scanner/internal/constants"
+	"github.com/DataDog/datadog-iac-scanner/pkg/config"
+	"github.com/DataDog/datadog-iac-scanner/pkg/datadog"
+	cli "github.com/urfave/cli/v3"
+)
+
+const (
+	bundleConfigFileName    = "config.yaml"
+	bundleRulesFileName     = "rules.json"
+	bundleLibrariesFileName = "libraries.json"
+)
+
+var fetchBundleAction = &cli.Command{
+	Name: "fetch-bundle",
+	Usage: "Fetches the default ruleset, Rego libraries, and merged remote " +
+		"configuration for a repository and writes them to local files, for " +
+		"later use with `scan --offline-bundle-path` in a network-isolated environment",
+	Flags: []cli.Flag{
+		&cli.StringFlag{
+			Name:  "repo-path",
+			Usage: "repository root path, used to discover a local configuration file",
+			Value: ".",
+		},
+		&cli.StringFlag{
+			Name:     "repo-url",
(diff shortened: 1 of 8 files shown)


cmd/scanner/scan.go
```diff
@@ -212,13 +218,30 @@ func runScan(ctx context.Context, c *cli.Command) error {
...
-	cfg, _, err := config.ReadConfiguration(ctx, repoDir, config.WithDatadog(datadog.NewDatadogClient(), repoInfo.RepositoryUrl))
-	if err != nil {
-		outErr := fmt.Errorf("error reading the configuration: %w", err)
-		if te := (*config.InvalidLocalConfigError)(nil); errors.As(err, &te) {
-			outErr = errorWithExitCode(outErr, constants.InvalidConfigErrorCode)
+	offlineBundlePath := c.String("offline-bundle-path")
+
+	var cfg *config.IacConfig
+	if offlineBundlePath != "" {
+		cfgBytes, err := os.ReadFile(filepath.Join(offlineBundlePath, bundleConfigFileName)) // nolint:gosec
```
whitemerch: Can we remove this suppression please

pkg/datadog/local_file_client_test.go
whitemerch, line 1: There is no test that calls fetchBundle against stub files, then calls runScan with --offline-bundle-path. The existing e2e suite (test/e2e) would be the right plac

cmd/scanner/fetch_bundle.go
```diff
@@ -0,0 +1,96 @@
...
+		&cli.StringFlag{
+			Name:     "output-dir",
+			Usage:    "directory to write the bundle files to",
+			Required: true,
+		},
+	},
+	Action: fetchBundle,
+}
+
+func fetchBundle(ctx context.Context, c *cli.Command) error {
```
whitemerch, line 58: fetch-bundle writes three flat files with no metadata. I may forget to refresh the bundle before a scan and it will silently run against stale rules

whitemerch (dismissed):

whitemerch (approved):

PR #240, opened 2026-07-06
[K9CODESEC-2945] Add Terraform module resolver package

whitemerch:
## Motivation

Terraform scans today only see files already on disk. When a repository references remote modules, the scanner cannot map those `module` blocks to actual `.tf` content, so findings inside those modules are missed.

This is the first step toward full remote module support: a resolver library that turns a module source string into a local directory path. Later work will wire this into the scan pipeline and add network-based resolution; this change lands the shared foundation without changing scan behavior.

To see the whole flow, refer to https://github.com/DataDog/datadog-iac-scanner/pull/203.

Related Ticket: [K9CODESEC-2945](https://datadoghq.atlassian.net/browse/K9CODESEC-2945?atlOrigin=eyJpIjoiMmQxMWVkNTlkNGQwNGQ5Y2E2ODQ3NjA2OGRhZTgyNDEiLCJwIjoiaiJ9)

## Changes

This PR adds a pluggable resolver package and the minimal module metadata needed to use it.

The resolver layer exposes a `Resolver` interface and a chain that tries implementations in order until one succeeds. In this PR, that chain covers offline sources only: local paths, `.terraform/modules` output from `terraform init`, and a JSON manifest for environments that prefetch modules before scanning. A disk cache deduplicates resolved directories across calls.

`ParsedModule` also gains call-site attribution fields and a typed `UnresolvedError` for callers that need to distinguish "module not found" from other failures. A small `ListModuleEntries` helper is included for upcoming module-discovery tooling.

Nothing in the scan path invokes this yet.

## QA Instruction

1. `go test ./pkg/parser/terraform/modules/...`
2. `go build ./...`

## Impact

Additive only: new package under `pkg/parser/terraform/modules/resolver`, small extensions to `pkg/parser/terraform/modules`. No CLI, scan, or engine behavior changes.

## Additional Notes

[K9CODESEC-2945]: https://datadoghq.atlassian.net/browse/K9CODESEC-2945?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/parser/terraform/modules/modules.go
```diff
@@ -32,13 +32,36 @@ type ParsedModule struct {
 	SourceType     string // local, git, registry, etc.
 	RegistryScope  string // public, private, or "" (non-registry)
 	AttributesData map[string]ModuleAttributesInfo
+	FileName       string
+	DefLine        int
```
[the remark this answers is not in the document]
whitemerch: I am doing this in another PR

pkg/parser/terraform/modules/resolver/gitexec.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+const gitSHALength = 40
+
+// gitProcSem caps concurrent git subprocesses across resolvers.
+var gitProcSem = make(chan struct{}, gitProcConcurrency())
+
+func gitProcConcurrency() int {
+	n := runtime.GOMAXPROCS(0)
+	const floor = 4
```
MikaYuoadas: 💡 **Suggestion:** Maybe you could extract it as a constant at the top of the file with a better name to clearly document the minimum git subprocesses concurrency?

pkg/parser/terraform/modules/resolver/gitexec.go
```diff
@@ -0,0 +1,47 @@
...
+	if len(ref) != gitSHALength {
+		return false
+	}
+	for _, c := range ref {
+		if (c < '0' || c > '9') && (c < 'a' || c > 'f') && (c < 'A' || c > 'F') {
+			return false
+		}
+	}
+	return true
+}
```
MikaYuoadas: ⛔ **Must fix:** Looks like the whole file is dead code, I didn’t find any reference to all the functions defined here.

pkg/parser/terraform/modules/modules.go
```diff
@@ -32,13 +32,36 @@ type ParsedModule struct {
...
 
 type ModuleAttributesInfo struct {
 	Resources []string          `json:"resources"`
 	Inputs    map[string]string `json:"inputs"`
 }
 
+type UnresolvedModule struct {
+	Module ParsedModule
+	Reason string
+}
```
MikaYuoadas: ⛔ **Must fix:** Same it seems to be dead code, there’s no reference to it anywhere. 🤔

pkg/parser/terraform/modules/resolver/resolver.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"context"
+
+	tfmodules "github.com/DataDog/datadog-iac-scanner/pkg/parser/terraform/modules"
+)
+
+// Resolution holds a resolved module directory and optional cleanup.
+type Resolution struct {
+	LocalPath string
+	Cleanup   func() // optional post-scan cleanup
```
MikaYuoadas, line 19: ❓ **Question:** Not used by current code, is it planned for future use?

pkg/parser/terraform/modules/resolver/chain.go
```diff
@@ -0,0 +1,41 @@
...
+		if err == nil {
+			return res, nil
+		}
+		lastErr = err
+	}
+	var ue *tfmodules.UnresolvedError
+	if errors.As(lastErr, &ue) {
+		return Resolution{}, lastErr
+	}
+	return Resolution{}, &tfmodules.UnresolvedError{Reason: lastErr.Error()}
```
MikaYuoadas: 💡 **Suggestion:** Wondering why we only report the last error if no resolver could resolve? Why not use [errors.Join()](https://pkg.go.dev/errors#Join) instead to return them all?

pkg/parser/terraform/modules/resolver/cache.go
```diff
@@ -0,0 +1,121 @@
...
+		if d.IsDir() {
+			return os.MkdirAll(target, cacheDirPerms)
+		}
+		data, err := fs.ReadFile(os.DirFS(src), path)
+		if err != nil {
+			return err
+		}
+		return os.WriteFile(target, data, cacheFilePerms)
+	})
+}
```
MikaYuoadas: 💡 **Suggestion:** Isn’t that just a reimplementation of stdlib’st [CopyFS](https://pkg.go.dev/os#CopyFS)? you could replace any call to `copyDir(dst, src)` with `os.CopyFS(dst, os.DirFS(src))`.

Also it would avoid buffering each file as a whole in memory as it simply use `io.Copy()` streaming internally.

pkg/parser/terraform/modules/list_modules.go
```diff
@@ -0,0 +1,51 @@
...
+
+// ListModuleEntries converts parsed modules to list-modules JSON rows.
+func ListModuleEntries(modules map[string]ParsedModule, includeLocal bool) []ListModuleEntry {
+	entries := make([]ListModuleEntry, 0, len(modules))
+	for src := range modules {
+		mod := modules[src]
+		if mod.IsLocal && !includeLocal {
+			continue
+		}
+		if strings.HasPrefix(mod.Source, "__") || mod.Source == "" {
```
MikaYuoadas, line 29: ❓ **Question:** What does the magic string "__" means here? Maybe add a comment explaining it?

MikaYuoadas (changes requested):

pkg/parser/terraform/modules/resolver/chain.go
whitemerch: I didn't know about it 😄

pkg/parser/terraform/modules/resolver/resolver.go
whitemerch: Yes I was trying to keep everything not related to the core of the implementation in its own PR so I don't end up sending you big PRs haha, but I can definitely put it back to another PR

pkg/parser/terraform/modules/list_modules.go
whitemerch: The __ prefix guards against internal sentinel values like `__UNRESOLVED__` and `__INVALID_TRAVERSAL__` that the HCL evaluator uses when it can't resolve a source expression. I will add a comment about it

MikaYuoadas: Ok thanks for adding that comment 👍

MikaYuoadas (dismissed):
LGTM thanks for addressing every comments 🙏

MikaYuoadas (approved):

PR #242, opened 2026-07-07
[K9CODESEC-2946] Add git-based Terraform module resolvers

whitemerch:
## Motivation

The resolver package from #240 handles offline sources only. Many Terraform repos reference git modules with pinned refs (`git::https://...?ref=v1.2.3`). When the repository is already cloned locally, or when a bare clone is available, we can resolve those modules without a network fetch.

This PR adds git resolution on top of #240. It is stacked on that branch and will merge after #240 lands (which itself waits on #234).

Related Ticket: [K9CODESEC-2946](https://datadoghq.atlassian.net/browse/K9CODESEC-2946?atlOrigin=eyJpIjoiNWZmOGQzNTczYWEzNGIyNTk5MjYwMGM0ODQ2MjQ5MGIiLCJwIjoiaiJ9)

## Changes

**Git resolvers**

Adds two offline git resolvers: `baregit` resolves pinned `git::` sources from a shared bare clone cache, and `localref` resolves refs that already exist in the scan workspace's `.git` directory. Both plug into the existing `Resolver` chain from #240.

**Shared git helpers**

Adds a concurrency-limited git subprocess semaphore and SHA/ref detection utilities used by both resolvers.

Nothing in the scan path invokes these resolvers yet.

## Author Checklist

## QA Instruction

1. `go test ./pkg/parser/terraform/modules/...`
2. `go build ./...`

## Blast Radius

Additive only: new resolver implementations under `pkg/parser/terraform/modules/resolver`. No CLI, scan, or engine behavior changes.

## Additional Notes

[K9CODESEC-2946]: https://datadoghq.atlassian.net/browse/K9CODESEC-2946?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/parser/terraform/modules/resolver/baregit.go
+++ b/pkg/parser/terraform/modules/resolver/baregit.go
@@ -172,9 +172,10 @@ func (repo *bareRepo) ensureClone(ctx context.Context) error {
 		}
 		if info, err := os.Stat(repo.barePath); err == nil && info.IsDir() {
 			// Validate that the directory is a real bare repo, not a partial clone
-			// left by a killed process. gitInDir is not available here since it
-			// uses the gitexec helpers; use exec.Command directly for this one check.
-			if check := gitInDir(ctx, repo.barePath, "rev-parse", "--git-dir"); check.Run() == nil {
+			// left by a killed process. This is a local-only read (no network, no
+			// pack-file I/O) so it intentionally bypasses the acquireGitProc semaphore,
+			// which is reserved for operations that meaningfully consume system resources.
+			if gitInDir(ctx, repo.barePath, "rev-parse", "--git-dir").Run() == nil {
 				repo.cloneOK.Store(true)
 				return nil, nil
 			}
@@ -273,15 +274,15 @@ func (repo *bareRepo) fetchRef(ctx context.Context, ref string) (string, error) 
 
 		// Local fast path: if the bare clone already contains this ref (e.g. a
 		// pre-populated clone or a prior fetch), resolve it without hitting the network.
-		if tryLocal := gitInDir(ctx, repo.barePath, "rev-parse", "--verify", safeRef); tryLocal != nil {
-			if out, localErr := tryLocal.Output(); localErr == nil {
-				if resolved := strings.TrimSpace(string(out)); looksLikeSHA(resolved) {
-					repo.refMu.Lock()
-					repo.refCache[ref] = resolved
-					repo.refMu.Unlock()
-					go repo.saveBareRefCache()
-					return resolved, nil
-				}
+		// This is a local-only read so it intentionally bypasses acquireGitProc;
+		// the semaphore is acquired below only when a network fetch is needed.
+		if out, localErr := gitInDir(ctx, repo.barePath, "rev-parse", "--verify", safeRef).Output(); localErr == nil {
+			if resolved := strings.TrimSpace(string(out)); looksLikeSHA(resolved) {
+				repo.refMu.Lock()
+				repo.refCache[ref] = resolved
+				repo.refMu.Unlock()
+				go repo.saveBareRefCache()
+				return resolved, nil
 			}
 		}
 
--- a/pkg/parser/terraform/modules/resolver/baregit.go
+++ b/pkg/parser/terraform/modules/resolver/baregit.go
@@ -170,96 +170,109 @@ func (repo *bareRepo) ensureClone(ctx context.Context) error {
 		if repo.cloneFailed.Load() {
 			return nil, repo.cachedCloneError()
 		}
-		if info, err := os.Stat(repo.barePath); err == nil && info.IsDir() {
-			// Validate that the directory is a real bare repo, not a partial clone
-			// left by a killed process. This is a local-only read (no network, no
-			// pack-file I/O) so it intentionally bypasses the acquireGitProc semaphore,
-			// which is reserved for operations that meaningfully consume system resources.
-			if gitInDir(ctx, repo.barePath, "rev-parse", "--git-dir").Run() == nil {
-				repo.cloneOK.Store(true)
-				return nil, nil
-			}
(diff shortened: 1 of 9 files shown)


pkg/parser/terraform/modules/resolver/baregit.go
```diff
@@ -0,0 +1,576 @@
...
+	// Strip the transport scheme BEFORE removing the go-getter subdir separator.
+	// Otherwise the "//" in "https://" is mistaken for the subdir marker and the
+	// URL collapses to "https:", so https sources never match SCP-form remotes.
+	for _, scheme := range []string{"https://", "http://", "ssh://", "git://"} {
+		raw = strings.TrimPrefix(raw, scheme)
+	}
+	// SCP-style git@host:org/repo → host/org/repo
+	if at := strings.IndexByte(raw, '@'); at != -1 {
+		after := raw[at+1:]
+		raw = strings.Replace(after, ":", "/", 1)
```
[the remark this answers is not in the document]
whitemerch: The normalizer's purpose is deduplication, it produces a stable key for matching two spellings of the same repo (e.g. git::https://github.com/org/repo.git vs git@github.com:org/repo), not a lossless URL reconstruction. The clone URL comes from `canonicalSSHCloneURL`, which is derived from the original parsed source, not from `normalizeGitRepoURL`.

For the non-default port case: after ssh:// is stripped, a URL like git@example.com:2222/org/repo would not have a user prefix, so the @ branch isn't entered and the port colon is left as-is. For ssh://user@host:2222/path, the @ replacement would indeed mangle the port, but Terraform module sources virtually never use non-default SSH ports; git::ssh:// with a port number isn't a pattern in any Terraform module registry or common provider

pkg/parser/terraform/modules/resolver/baregit.go
```diff
@@ -0,0 +1,619 @@
...
+
+	// Branch/tag ref: fetch and resolve to SHA. Key on "ref:<name>" so it
+	// doesn't conflict with SHA keys.
+	v, err, _ := repo.fetchSF.Do("ref:"+ref, func() (interface{}, error) {
+		// Fast path: return the cached SHA from a previous run, skipping the network entirely.
+		// The archive extraction has its own on-disk cache keyed by SHA, so the result is stable.
+		repo.refMu.RLock()
+		if cachedSHA, ok := repo.refCache[ref]; ok {
+			repo.refMu.RUnlock()
+			return cachedSHA, nil
```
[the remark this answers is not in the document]
whitemerch: This is intentional. The refs.json cache is designed to give deterministic, scan-run-scoped results: each scan run produces a consistent snapshot of modules at a point in time. Aggressively re-fetching mutable branch refs would turn an offline-capable scan into a mandatory network operation, which breaks the main use case, air-gapped or pre-fetched environments

pkg/parser/terraform/modules/resolver/baregit.go
```diff
@@ -0,0 +1,626 @@
...
+			return nil, nil
+		}
+		if repo.cloneFailed.Load() {
+			return nil, repo.cachedCloneError()
+		}
+		if info, err := os.Stat(repo.barePath); err == nil && info.IsDir() {
+			// Validate that the directory is a real bare repo, not a partial clone
+			// left by a killed process. gitInDir is not available here since it
+			// uses the gitexec helpers; use exec.Command directly for this one check.
+			if check := gitInDir(ctx, repo.barePath, "rev-parse", "--git-dir"); check.Run() == nil {
```
MikaYuoadas: ❓ **Question:** This git call isn’t using the `acquireGitProc` sem. Is it on purpose?

Also if not, wouldn’t it be simpler for helpers `gitInDir`, `gitInWorktree` and `gitCloneBare` to acquire and release the lock automatically themselves?

pkg/parser/terraform/modules/resolver/baregit.go
```diff
@@ -0,0 +1,626 @@
...
+		repo.refMu.RUnlock()
+
+		safeRef, refErr := gitSafeArg(ref)
+		if refErr != nil {
+			return "", refErr
+		}
+
+		// Local fast path: if the bare clone already contains this ref (e.g. a
+		// pre-populated clone or a prior fetch), resolve it without hitting the network.
+		if tryLocal := gitInDir(ctx, repo.barePath, "rev-parse", "--verify", safeRef); tryLocal != nil {
```
MikaYuoadas: ❓ **Question:** Same, this git call isn’t using the `acquireGitProc` sem. Is it on purpose?

MikaYuoadas: 🤏 **Nitpick:** I don’t think `gitInDir()` can ever return nil so this guard will always be True and can probably be removed.

MikaYuoadas (dismissed):
LGTM, a couple nits and questions

pkg/parser/terraform/modules/resolver/baregit.go
whitemerch: Yes, intentional. These are local-only rev-parse checks (no network / pack I/O), so they don't need the semaphore, that's reserved for heavier work like clone/fetch/archive. I added a comment to make that explicit. I kept acquisition at the call site rather than inside gitInDir/gitCloneBare because callers already hold the semaphore around network ops

pkg/parser/terraform/modules/resolver/baregit.go
whitemerch: Same as above

pkg/parser/terraform/modules/resolver/baregit.go
MikaYuoadas: Make sense indeed 👍

MikaYuoadas (dismissed):

MikaYuoadas (approved):

PR #249, opened 2026-07-07
Handle missing HCL expression types in Terraform parsing

whitemerch:
## Motivation

Terraform HCL parsing still had gaps for a few expression node types that appear in real modules. When those nodes were hit, splat traversals could lose their trailing path, template `%{for}` blocks could not be rendered, and some synthetic or invalid-expression placeholders fell through to generic unsupported handling.

## Changes

**Splat traversals**

Splat expressions with a trailing attribute or index were truncated because the anonymous splat item in the AST was not handled.

```hcl
# input
security_group_ids = aws_security_group.app[*].id

# before
"aws_security_group.app[*]"

# after
"aws_security_group.app[*].id"
```

**Template for-loops**

`%{for}` / `%{endfor}` template directives produce a `TemplateJoinExpr` in the AST. That shape was unsupported in some parsing paths.

```hcl
# input
user_data = "%{for v in var.subnets}${v.cidr}%{endfor}"

# before
"__UNSUPPORTED_EXPR__"

# after
"[for v in var.subnets : v.cidr]"
```

**Remaining HCL expression types**

`AnonSymbolExpr` (the synthetic item inside `var.list[*].id`) and `ExprSyntaxError` (placeholder for invalid or incomplete expressions) now have explicit handling in the shared HCL expression visitor instead of falling through to the default fallback.

```hcl
# splat with nested traversal
bucket = aws_s3_bucket.logs[*].bucket

# before → "aws_s3_bucket.logs[*]"
# after  → "aws_s3_bucket.logs[*].bucket"
```

## Author Checklist

## QA Instruction

1. `go test ./pkg/hclexpr/... ./pkg/engine ./pkg/builder/engine/... ./pkg/parser/terraform/...`
2. Splat with traversal: parse `var.list[*].id` and confirm the rendered path is `var.list[*].id`, not `var.list[*]`.
3. Template for-loop: parse `%{for v in var.list}${v}%{endfor}` and confirm it renders as `[for v in var.list : v]`.

## Blast Radius

Datadog IaC Scanner Terraform parsing only: HCL expression rendering in the inspector, engine string conversion, module resolver, and shared `hclexpr` dispatch.

## Additional Notes

N/A

### DIFF
--- a/pkg/engine/inspector.go
+++ b/pkg/engine/inspector.go
@@ -1461,13 +1461,11 @@ func (v *inspectorExprVisitor) VisitDefault(e hclsyntax.Expression) (ast.Value, 
 func expressionToASTTemplateExpr(e *hclsyntax.TemplateExpr) ast.Value {
 	result := ""
 	for _, part := range e.Parts {
-		switch p := part.(type) {
-		case *hclsyntax.LiteralValueExpr:
-			if p.Val.Type().Equals(cty.String) {
-				result += p.Val.AsString()
-			}
-		default:
+		v, err := expressionToAST(part)
+		if err != nil || v == nil {
 			result += "${...}"
+		} else {
+			result += astValueToSimpleString(v)
 		}
 	}
 	return ast.String(result)
--- a/pkg/engine/inspector.go
+++ b/pkg/engine/inspector.go
@@ -1462,10 +1462,11 @@ func expressionToASTTemplateExpr(e *hclsyntax.TemplateExpr) ast.Value {
 	result := ""
 	for _, part := range e.Parts {
 		v, err := expressionToAST(part)
-		if err != nil || v == nil {
+		s := astValueToSimpleString(v)
+		if err != nil || v == nil || s == "__UNSUPPORTED_EXPR__" || s == "__UNSUPPORTED_LITERAL__" {
 			result += "${...}"
 		} else {
-			result += astValueToSimpleString(v)
+			result += s
 		}
 	}
 	return ast.String(result)
@@ -1595,6 +1596,10 @@ func expressionToASTForExpr(e *hclsyntax.ForExpr) ast.Value {
 		b.WriteString("}")
 	} else {
 		b.WriteString("[for ")
+		if e.KeyVar != "" {
+			b.WriteString(e.KeyVar)
+			b.WriteString(", ")
+		}
 		b.WriteString(e.ValVar)
 		b.WriteString(" in ")
 		b.WriteString(collStr)
--- a/pkg/engine/inspector.go
+++ b/pkg/engine/inspector.go
@@ -1430,7 +1430,9 @@ func (v *inspectorExprVisitor) VisitObjectCons(e *hclsyntax.ObjectConsExpr) (ast
 	return expressionToASTObjectConsExpr(e), nil
 }
 func (v *inspectorExprVisitor) VisitTemplateJoin(e *hclsyntax.TemplateJoinExpr) (ast.Value, error) {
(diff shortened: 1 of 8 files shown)


robertohuertasm-datadog (commented):
The PR does address its stated goal. For the common case, this is an improvement:

```hcl
"%{for v in var.list}${v}%{endfor}"
```

One edge case to consider: the existing `ForExpr` renderer appears to mishandle list-style for-expressions with two loop variables:

```hcl
[for k, v in var.map : k]
```

`KeyExpr` is `nil` for this form because it is a list/tuple comprehension, not an object comprehension, but `KeyVar` is still populated (k). The current rendering branch only writes `ValVar`, so it can produce:

```hcl
[for v in var.map : k]
```

That bug mostly pre-existed for direct `ForExpr` rendering. This PR makes it more visible because `%{for ...}` template directives now route through the same `ForExpr` renderer, so this template case inherits the issue too:

```hcl
"%{for k, v in var.map}${k}%{endfor}"
```

Check [engine.go:122](https://github.com/DataDog/datadog-iac-scanner/blob/691c870cd12d28d2621c7a7350e8f6f74154487f/pkg/builder/engine/engine.go#L122) and [inspector.go:1598](https://github.com/DataDog/datadog-iac-scanner/blob/691c870cd12d28d2621c7a7350e8f6f74154487f/pkg/engine/inspector.go#L1598).

Could we preserve `KeyVar` in the tuple/list branch when it is set, and add a regression test for both direct `[for k, v in ...]` and template `%{for k, v in ...}` forms?

In theory this should be a small fix:

```go
b.WriteString("[for ")
if e.KeyVar != "" {
	b.WriteString(e.KeyVar)
	b.WriteString(", ")
}
b.WriteString(e.ValVar)
b.WriteString(" in ")
```

pkg/engine/inspector.go
```diff
@@ -1444,20 +1446,26 @@ func (v *inspectorExprVisitor) VisitForExpr(e *hclsyntax.ForExpr) (ast.Value, er
...
 	for _, part := range e.Parts {
-		switch p := part.(type) {
-		case *hclsyntax.LiteralValueExpr:
-			if p.Val.Type().Equals(cty.String) {
-				result += p.Val.AsString()
-			}
-		default:
+		v, err := expressionToAST(part)
+		if err != nil || v == nil {
 			result += "${...}"
```
robertohuertasm-datadog, line 1467: The `${...}` branch only triggers on `err != nil || v == nil`, but the visitor fallbacks (`VisitDefault`, `VisitExprSyntaxError`, and `literalToAst`'s default) all return a non-nil `ast.String` sentinel ("__UNSUPPORTED_EXPR__" / "__UNSUPPORTED_LITERAL__") with a `nil` error. So when a part genuinely can't be rendered, the sentinel gets concatenated verbatim into the template string instead of collapsing to `${...}`. e.g. "`prefix-__UNSUPPORTED_EXPR__`" gets handed to the rules.

In practice this is low impact: for well-formed Terraform every interpolation type is handled after this PR, so you'd mainly hit it on malformed/partial HCL (`ExprSyntaxError`). But since the `${...}` fallback is dead as written, might be worth treating a sentinel result as the fallback case, e.g.:

```go
s := astValueToSimpleString(v)
if err != nil || v == nil || s == "__UNSUPPORTED_EXPR__" || s == "__UNSUPPORTED_LITERAL__" {
    result += "${...}"
} else {
    result += s
}
```

robertohuertasm-datadog (approved):
LGTM! 🚀

PR #250, opened 2026-07-08
[K9CODESEC-2948/2949] Wire Terraform module resolution into engine and module parser

whitemerch:
> [!TIP]

## Motivation

The scanner today silently skips non-local Terraform modules: a `source = "git::..."` or registry reference has no local directory to read from, so the module's variables and resources are never evaluated. This PR lays the internal groundwork for changing that, a resolver interface in the module parser and injection points in the engine, so a follow-up PR can wire in the actual download step without touching default behavior. Nothing here changes what a scan produces today.

## Changes

**Module parser**

The parser now accepts an optional `RemoteResolver` callback, a source-to-local-directory lookup, so the engine can supply downloaded paths without the parser needing to know how fetching works. Resolution failures for individual modules are non-fatal: a warning is logged per module and the scan continues with what is available. A canceled context during enrichment propagates as `context.Canceled` rather than returning partial results. `vfs.FS` threading is preserved throughout so the `serve` endpoint's in-memory filesystem (used by the IDE extension) continues to work for local module sibling resolution.

**Engine injection points**

Two new methods on `Inspector`, `SetRemoteModuleDirectories` and `SetExternalModulePaths`, let the scan pipeline hand off a pre-downloaded source map before the scan runs. Module metadata merging now keys equivalent maps by source and version so multiple instances of the same module resolve correctly. If neither injection method is called the behavior is identical to today.

**Inspector**

The module resolver is connected to the scan lifecycle. Local module instantiation remains gated on the existing `IacEnableLocalModuleEval` feature flag. No scan-pipeline or CLI wiring calls the new injection methods yet.

## Author Checklist

## QA Instruction

CI should pass

## Impact

Additive only. No CLI flags, scan parameters, or default scan behavior changes. The new `Inspector` methods have no external callers yet; the follow-up PR will connect them to scan parameters and an opt-in CLI flag.

### DIFF
--- a/pkg/engine/inspector.go
+++ b/pkg/engine/inspector.go
@@ -12,6 +12,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"maps"
+	"os"
+	"path/filepath"
 	"reflect"
 	"regexp"
 	"runtime/debug"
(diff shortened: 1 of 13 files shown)


MikaYuoadas (approved):
LGTM

PR #251, opened 2026-07-08
[K9CODESEC-2948/2950] Wire Terraform remote modules into scanner

whitemerch:
> [!TIP]
> This PR is stacked on #250 and should be reviewed after that PR.

> [!NOTE]
> I have one final PR after this one to make some performance improvements, cleanup and optimizations.

## Motivation

The resolver and engine layers can now map Terraform module calls to local directories, but the scanner does not yet invoke that work before inspection. This PR adds opt-in scan-pipeline wiring so Terraform remote modules can be resolved, parsed, and evaluated without changing default scan behavior.

## Changes

**Scan pipeline**

When `--x-remote-modules` is enabled, the scanner collects Terraform files from the analyzed file set, parses module calls, resolves them through the existing resolver chain, and injects the resolved directories into the inspector before services run.

**Source inclusion**

Resolved module directories are added to the scan source paths so their Terraform files are parsed like regular scan input. The inspector marks those directories as external module paths so rule-level path filters do not drop findings just because the module content lives outside the repository subtree.

**CLI**

Hidden experimental flags expose the opt-in path: `--x-remote-modules`, `--x-remote-modules-manifest`, and `--x-remote-modules-allowed-host`. The manifest and host allowlist flags require `--x-remote-modules`. Enabling remote modules also enables the existing local-module evaluation gate because the resolved-directory path is consumed by that engine flow.

**Resolution behavior**

The resolver chain first uses local/prepared sources such as a manifest, `.terraform/modules`, and local git refs, then falls back to remote-capable resolvers. `--x-remote-modules-allowed-host` constrains go-getter and registry-style fetches when configured.

## Author Checklist

## QA Instruction

1. `go test ./pkg/scan ./pkg/parser/terraform/modules/... ./pkg/engine ./pkg/engine/source`
2. `go build ./cmd/scanner`
3. Scan a Terraform fixture without `--x-remote-modules` and confirm output matches the base branch.
4. Scan a Terraform fixture with `--x-remote-modules --x-remote-modules-manifest <manifest>` and confirm findings can come from the resolved module directory.

## Impact

Additive and opt-in only. Default CLI scans do not resolve or scan remote modules unless `--x-remote-modules` is passed. With that flag enabled, the scanner may read prepared module directories or fetch module content depending on the source and available resolver.

### DIFF
--- a/pkg/scan/remote_modules.go
+++ b/pkg/scan/remote_modules.go
@@ -78,6 +78,9 @@ func (c *Client) prepareRemoteModules(ctx context.Context, paths []string, inspe
 	remotePaths := mapValues(dirs)
 	inspector.SetExternalModulePaths(remotePaths)
 	c.remoteModulePaths = remotePaths
+	if err := c.addRemoteModuleFilesToInventory(remotePaths); err != nil {
+		return err
+	}
 	contextLogger := logger.FromContext(ctx)
 	contextLogger.Info().Msgf("Resolved %d Terraform remote module directories", len(dirs))
 	return nil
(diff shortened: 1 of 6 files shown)


pkg/scan/remote_modules.go
```diff
@@ -0,0 +1,274 @@
...
+type moduleResolverAdapter struct {
+	resolver tfresolver.Resolver
+}
+
+func (r moduleResolverAdapter) Resolve(ctx context.Context, mod *tfmodules.ParsedModule) (string, error) {
+	res, err := r.resolver.Resolve(ctx, mod)
+	if err != nil {
+		return "", err
+	}
+	return res.LocalPath, nil
```
ChouraquiBen, line 40: nit: res.CleanUp is dropped here and since it is not used, non-cacheable modules are never cleaned up.

ChouraquiBen (commented):
I’m worried the Terraform module ownership is getting split across too many layers. Local module parsing/resolution lives in the Terraform parser package but is invoked from the inspector, while this PR adds remote module discovery/resolution in the scanner before handing state back to the inspector. That means module parsing and lifecycle concerns are now spread across pkg/parser/terraform, pkg/engine, and pkg/scan, which may get difficult to maintain as local/remote behavior evolves.
Could we centralize Terraform module discovery/resolution behind one abstraction that returns the resolved module metadata, extra inventory paths, and cleanup hooks for the scanner/inspector to consume?

whitemerch:
> I’m worried the Terraform module ownership is getting split across too many layers. Local module parsing/resolution lives in the Terraform parser package but is invoked from the inspector, while this PR adds remote module discovery/resolution in the scanner before handing state back to the inspector. That means module parsing and lifecycle concerns are now spread across pkg/parser/terraform, pkg/engine, and pkg/scan, which may get difficult to maintain as local/remote behavior evolves. Could we centralize Terraform module discovery/resolution behind one abstraction that returns the resolved module metadata, extra inventory paths, and cleanup hooks for the scanner/inspector to consume?

@ChouraquiBen I have a PR actually addressing this same complain ahha [Resolve nested Terraform modules with bounded traversal](https://github.com/DataDog/datadog-iac-scanner/pull/253)

ChouraquiBen (approved):
Nice! Thanks for the changes

PR #253, opened 2026-07-13
Resolve nested Terraform modules with bounded traversal

whitemerch:
## Motivation

Opt-in Terraform remote module resolution stopped after direct dependencies, so the scanner could miss findings in remote modules referenced through nested remote or local module calls. This change follows the module graph recursively while bounding network and resource use and preserving source attribution.

## Changes

**Module traversal**

Adds a Terraform-owned graph resolver that follows local and remote module calls up to a configurable maximum depth. It keeps local child directories available for traversal without scheduling them again as root seeds, deduplicates equivalent module resolutions, and bounds concurrent discovery work.

**Resolution safeguards**

Adds configurable graph-depth, per-module fetch-timeout, and total-download-size limits, with defaults of 8 levels, 30 seconds, and 200 MiB. Resolved modules reuse the module cache when available, and temporary downloads are released through one idempotent cleanup lifecycle.

**Scan integration**

Adds materialized Terraform files to the scan inventory even when repository path filters would otherwise exclude their download location. It also carries canonical source mappings and module-call metadata into inspection so findings remain attributed to the original module source.

## QA Instruction

1. `go test ./pkg/parser/terraform/modules/... ./pkg/engine ./pkg/scan -count=1`
2. `go test -race ./pkg/parser/terraform/modules/modulegraph -count=1`
3. `golangci-lint run -c .golangci.yml --timeout 20m`

## Impact

This affects only scans that opt into experimental Terraform remote module resolution. Default scan behavior is unchanged, and the feature is not yet released.

## Additional Notes


pkg/engine/provider/filesystem.go
```diff
@@ -150,6 +151,51 @@ func (s *FileSystemSourceProvider) ExcludePaths(ctx context.Context, paths []str
...
+	for _, scanPath := range s.paths {
+		fileInfo, err := os.Stat(scanPath)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to open path")
+		}
+		if !fileInfo.IsDir() {
+			if shouldSkip, _ := s.checkConditions(ctx, fileInfo, extensions, scanPath, nil); shouldSkip {
+				continue
+			}
+			files = append(files, strings.ReplaceAll(scanPath, "\\", "/"))
```
ChouraquiBen: nit: `filepath.ToSlash`? Or it does not work here because the slash are doubled? Pretty sure that it is the reason, don't bother answering if that is the reason.

pkg/engine/provider/filesystem.go
```diff
@@ -150,6 +151,51 @@ func (s *FileSystemSourceProvider) ExcludePaths(ctx context.Context, paths []str
...
+			continue
+		}
+		noopSink := func(context.Context, string) ([]string, error) {
```
ChouraquiBen: Question: I got the same comment from @NathanGallet-dd previously so passing the ball here: is there a reason why using a function defined directly in the function is more elegant than creating the function outside this scope and reusing it whenever?

pkg/parser/terraform/modules/modulegraph/modulegraph.go
```diff
@@ -0,0 +1,546 @@
...
+}
+
+type resolvedEntry struct {
+	res resolver.Resolution
+	err error
+}
+
+type walker struct {
+	mu             sync.Mutex
+	visited        map[string]bool
```
ChouraquiBen: Remark: That is...
a lot of shared states. I haven't read the rest of the code yet, so I am not yet aware of implementation details but the risks of race conditions are only increased by that many shared states

pkg/parser/terraform/modules/modulegraph/modulegraph.go
```diff
@@ -0,0 +1,546 @@
...
+	}
+	w.visited[localPath] = true
+	return true
+}
+
+func (w *walker) addPaths(paths ...string) {
+	if len(paths) == 0 {
+		return
+	}
+	w.mu.Lock()
```
ChouraquiBen: nit: `defer w.mu.Unlock()` is more idiomatic

ChouraquiBen: Same goes after. This helps prevent to forget to `Unlock`. (You can argue that it may lead to people not unlocking in the correct place, cf up there line 138-143 where it could get really messy). But there is no such issue here

pkg/parser/terraform/modules/modulegraph/modulegraph.go
ChouraquiBen: That would be quite te refactor to make it closer to stateless, I know that, but I wanted at lest to open the discussion

pkg/parser/terraform/modules/modulegraph/modulegraph.go
```diff
@@ -0,0 +1,546 @@
...
+		if inRemoteTree {
+			w.addPaths(flatTerraformFilePaths(localDir)...)
+		}
+		w.traverse(ctx, localDir, childAllowedFiles, repoAllowedDirs, depth+1, inRemoteTree)
+	}
+
+	w.traverseRemoteModules(ctx, mods, repoAllowedDirs, depth)
+}
+
+type remoteModuleGroup struct {
```
ChouraquiBen: nit: aren't usually structs all defined at the top with the vars and stuff?

pkg/parser/terraform/modules/modulegraph/modulegraph.go
```diff
@@ -0,0 +1,546 @@
...
+) {
+	groups := make(map[string]*remoteModuleGroup)
+	for key := range mods {
+		mod := mods[key]
+		if mod.IsLocal {
+			continue
+		}
+		id := remoteResolveIdentity(&mod)
+
+		w.mu.Lock()
```
ChouraquiBen: nit: Okay, this one is really nit-picky a lot more than the others but you could use `RLock` to allow other readers to read and only prevent writes

pkg/parser/terraform/modules/modulegraph/modulegraph.go
```diff
@@ -0,0 +1,546 @@
...
+	}
+	if resolution.LocalPath == "" {
+		if !shared {
+			contextLogger.Warn().
+				Msgf("Resolved remote Terraform module %q without a local path", representative.Source)
+		}
+		return
+	}
+	if !shared {
+		contextLogger.Info().Msgf("Fetched remote Terraform module %q", representative.Source)
```
ChouraquiBen: Remark: that means that all of those logs will appear in the explorer. Won't that be a bit much?

pkg/parser/terraform/modules/modulegraph/modulegraph.go
```diff
@@ -0,0 +1,546 @@
...
+	cleanups       []func()
+
+	sf       singleflight.Group
+	resolver resolver.Resolver
+	maxDepth int
+	fsys     vfs.FS
+
+	parseMu    sync.Mutex
+	parseCache map[string]map[string]tfmodules.ParsedModule
+}
```
ChouraquiBen, line 103: Suggestion: coming back here after some thinking. It is a bit hard to keep notes of all the functions and know which ones should use mutex and which ones hold them. To make sure that there is no deadlock or anything, do you think creating an interface which walker would inherit from where we could annotate the methods using a mutex and the ones who don't would be interesting?

pkg/scan/remote_modules.go
```diff
@@ -9,293 +9,182 @@ import (
...
-		mod := &enriched[i]
-		if mod.IsLocal || mod.AbsSource == "" {
-			continue
-		}
-		root := moduleRoot(mod.FileName, c.ScanParams.RepoPath)
-		dirs[engine.RemoteModuleKey(root, mod.Source, mod.Version)] = mod.AbsSource
-		dirs[engine.RemoteModuleCallKey(root, mod.Source, mod.Version, mod.Name)] = mod.AbsSource
+	maxDepth := c.ScanParams.ModuleMaxDepth
+	if maxDepth == 0 {
+		maxDepth = 8
```
ChouraquiBen: Question: why 8?

pkg/scan/remote_modules.go
```diff
@@ -9,293 +9,182 @@ import (
...
-}
 
-func (f *terraformModuleFiles) addScanPath(scanPath string) error {
-	info, err := os.Stat(scanPath)
-	if err != nil {
-		return err
+	if c.ScanParams.EnableRemoteModules {
+		resolvers = append(resolvers,
+			tfresolver.NewLocalGitRefResolver(c.ScanParams.Path, ""),
+			tfresolver.NewBareGitResolver(""),
```
ChouraquiBen: From Codex: --x-remote-modules-allowed-host is only applied to GoGetterResolver at line 131, but BareGitResolver handles pinned git:: sources first and can clone disallowed hosts before the allowlist is checked. That makes the host allowlist incomplete for git modules.

cmd/scanner/scan.go
```diff
@@ -111,6 +111,24 @@ var scanAction = &cli.Command{
 			Usage:  "(experimental) host allowed for Terraform remote module downloads",
 			Value:  []string{},
 		},
+		&cli.IntFlag{
+			Name:   "x-remote-modules-max-depth",
+			Hidden: true,
+			Usage:  "(experimental) maximum BFS depth for the remote module graph walker (default 8)",
```
ChouraquiBen, line 117: nit: Codex made a comment on my machine saying that this is confusing because if it is 0, it will go back to 8 below, therefore the suggestion:
```suggestion
			Usage:  "(experimental) maximum BFS depth for the remote module graph walker (default 8, will be 8 if put at zero)",
```

pkg/scan/remote_modules.go
```diff
@@ -9,293 +9,182 @@ import (
...
-	return nil
-}
 
-func (f *terraformModuleFiles) addScanPath(scanPath string) error {
-	info, err := os.Stat(scanPath)
-	if err != nil {
-		return err
+	if c.ScanParams.EnableRemoteModules {
+		resolvers = append(resolvers,
+			tfresolver.NewLocalGitRefResolver(c.ScanParams.Path, ""),
```
ChouraquiBen: Codex comment: I think it is worth taking a look
"single-file scans no longer use the local checkout for git::...?ref= modules. LocalGitRefResolver gets c.ScanParams.Path, so -p path/main.tf makes it run git -C path/main.tf, which fails because it is not a directory. Those modules then fall through to network resolution and can be skipped offline/private. Use discovered module roots or filepath.Dir for file scan paths."

ChouraquiBen (commented):
Sorry, I left quite a few comments. I think that at least a few of them should be considered before leaving a final approval. Otherwise, I don't have any issue with the way it is currently working

pkg/scan/remote_modules.go
```diff
@@ -9,293 +9,178 @@ import (
...
-		tfresolver.NewBareGitResolver(""),
+		tfresolver.LocalResolver{},
+		&tfresolver.DotTerraformResolver{RootDirs: dotTerraformRootDirs(moduleDiscoveryPaths)},
 	}
+
 	if c.ScanParams.RemoteModulesManifestPath != "" {
 		manifest, err := tfresolver.LoadManifest(c.ScanParams.RemoteModulesManifestPath)
 		if err != nil {
-			return nil, err
+			contextLogger.Warn().Err(err).
```
ChouraquiBen, line 105: TBH, looked good to me but codex had this comment: saying that this part becoming non fatal can lead to Go getter/git downloads despite typos or invalid manifests and that (this feels more obscure to me), "valid manifest can be ignored when .terraform/modules/modules.json exists". It claims that it can lead to downloading different module code than desired

whitemerch: Yes I noticed that, which I am fixing in the upcoming [PR](https://github.com/DataDog/datadog-iac-scanner/pull/254)

ChouraquiBen (approved):
LGTM

PR #273, opened 2026-07-30
(k9-iac): Validate custom-rule platforms against the same list as scan

valeryjuli:
## Summary

https://datadoghq.atlassian.net/browse/K9CODESEC-3913

To validate custom rules on the UI, the validation API uses the `custom_scan` command. However, the custom scan command supports a different set of platforms than the `scan` command, when in reality should be the same. 

This change is needed for the custom rules UI where today  CICD custom-rule validation was silently failing because `cicd` wasn't handled by the old hardcoded switch.

## Changes
- `custom evaluate`/`custom validate` now validate `--platform` against `platforms.Supported`, the same list `scan` uses, instead of silently falling through to a JSON default for unrecognized platforms.
-  Added tests

## Test plan
- [x] `go build ./...`
- [x] `go test ./cmd/scanner/... ./pkg/scan/...`

### DIFF
--- a/pkg/scan/custom_scan.go
+++ b/pkg/scan/custom_scan.go
@@ -20,6 +20,7 @@ import (
 	"github.com/DataDog/datadog-iac-scanner/pkg/engine/source"
 	"github.com/DataDog/datadog-iac-scanner/pkg/featureflags"
 	"github.com/DataDog/datadog-iac-scanner/pkg/model"
+	"github.com/DataDog/datadog-iac-scanner/pkg/platforms"
 	"github.com/DataDog/datadog-iac-scanner/pkg/printer"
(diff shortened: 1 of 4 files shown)


pkg/scan/custom_scan.go
```diff
@@ -36,26 +37,28 @@ type RegoValidationError struct {
...
-	switch strings.ToLower(platform) {
-	case "terraform":
-		return "scan-target.tf"
-	case "cloudformation", platformARM:
-		return scanTargetJSON
-	case "kubernetes", "ansible":
-		return "scan-target.yaml"
-	case "dockerfile":
-		return "Dockerfile"
-	default:
```
valeryjuli, line 56: CI/CD was missing from this list :(

NathanGallet-dd (approved):

PR #277, opened 2026-07-31
feat(server): accept absolute file paths in content-push mode

robertohuertasm-datadog:
Pairs with [datadog-vscode#2472](https://github.com/DataDog/datadog-vscode/pull/2472)
(K9CODESEC-3287), which is what surfaced this.

## The bug

VS Code lets you open a scannable file that is not inside any open workspace folder — File > Open File
on a lone `Dockerfile`, or a `.tf` from another repo while a different project is open. For those, the
extension cannot shorten the path and pushes an absolute one.

`validateFilePath` rejected absolute paths, and it runs per-file inside `validateAnalyzeRequest`, so a
single such file **failed the entire analyze request** with a 400. The extension caught the error, hit
`if (!response) return []`, and rendered the file with **zero findings** — indistinguishable from
clean. Failed analyses are not cached, so the doomed request was re-issued on every edit and refresh
for as long as the file stayed open.

Low trigger frequency, bad failure direction: a guaranteed failure on a plausible user action, failing
toward "your infrastructure is clean".

## What this PR changes

**Drops the absolute-path branch in `validateFilePath`.** The empty, NUL, and `..` checks stay.
`filepath.Clean` collapses interior `..` for absolute input, so only a *relative* path escaping its
root is still refused.

**Pins `IacEnableLocalModuleEval: false` for server mode.** This is the part that makes the deletion
defensible, and reviewers should weight it more than the deletion itself. See "Why this is safe".

**Extracts the flag overrides into `serverFlagEvaluator(parallelParsing bool)`.** They were inline in
the `scan.Parameters` literal, where a test cannot assert on them. Behaviour for the two pre-existing
flags is unchanged; the extraction exists so the new pin cannot silently regress.

**Fixes `MemFS.ReadDir` synthesizing a directory entry named `""`.** An absolute key under dir `"."`
has an empty first path segment, which named no directory. Latent, not live — see "Kept as-is" for
why it was fixed anyway.

**Tests and doc comments.** Enumerated below.

### The engine has always been absolute-path clean

The CLI does not merely tolerate absolute paths, it *mandates* them: it converts every input with
`filepath.Abs` up front and then runs the same engine. The server was the only caller forbidding what
the rest of the product requires everywhere.

### The check was a proxy for a flag, and now the flag holds it directly

The real hazard behind the absolute-path check was `resolveModuleDocuments` → `tfeval.LoadRootVars`,
which does a genuine `os.ReadDir` on directories derived from pushed paths — bounded to the working
directory's subtree with relative paths, unbounded with absolute ones.

It was unreachable in server mode **only because a remotely-togglable flag happened to default to
`false`**. The `serve` command does not even expose `--x-local-module-eval` (only `scan` does), so the
default was doing all the work. Pinning it moves the guarantee to where it actually lives, and makes
the path-shape check redundant rather than load-bearing.

Note this is a **net security improvement over `main`**: today that flag is one remote toggle away from
enabling an unbounded directory read in server mode. After this PR it is pinned and asserted by a test.

### Why not keep the `..` check as a half-measure

"Drop the absolute check, keep `..`" does not hold together. `..` exists to keep a *relative* path
bounded inside a root; once absolute paths are allowed, nothing needs `..` to reach anywhere. The
posture is close to binary, which is exactly why the guarantee belongs at the flag.

## Pre-existing issue found while reviewing this (please read)

**The YAML/JSON file resolver reads arbitrary files off disk. It is not introduced by this PR, and
this PR does not widen it. It is called out because the code and comments previously claimed an
invariant that does not hold.**

`file.Resolver` substitutes referenced file contents into the parsed document via `os.Open`
(`pkg/resolver/file/file.go:364`), bypassing the request's `vfs.FS` entirely. Two properties make it
wider than the name suggests:

- **Not limited to `$ref`.** `yamlWalk` recurses into *every scalar node* and tries to resolve each as
  a path. `$ref` only controls whether `RefMetadata` is attached afterwards.
- **Ignores `resolveReferences`.** That flag short-circuits only for OpenAPI-looking documents.

Reproduced end-to-end over a real `POST /ide/v1/iac/analyze`, default config, no flags. A pushed
ConfigMap with a plain scalar `x: ../secret.yaml` returned the unpushed file's contents in the response
body.

### Evidence that this PR does not make it worse

The same canary test replayed in a worktree at the merge base (`2349ce54`):

| Pushed path | Reference | This branch | Base `2349ce54` |
| --- | --- | --- | --- |
| `/tmp/ws/main.yaml` | `../secret.yaml` | leaks | **400** — old absolute check |
| `main.yaml` | `../` ×40 + `tmp/secret.yaml` | leaks | **leaks** |

Row 2 settles it: no absolute path involved, and it leaks on `main` today. The reason is that
**overshooting `../` past the filesystem root is clamped to `/`**, so a relative path is already a
fully reliable absolute-anchor primitive, independent of the server's working directory. Absolute
paths are more legible; they reach nothing new.

### Severity bounds

`findFilePath` requires the target's extension to be in the calling parser's list — YAML resolves only
`.yaml`/`.yml`, JSON only `.json`. So `/etc/passwd`, `~/.kube/config`, `~/.aws/credentials`, and
`~/.ssh/id_rsa` are **not** reachable. Still reachable: `~/.docker/config.json`, gcloud's
`application_default_credentials.json`, Kubernetes Secret manifests, CI configs with inline tokens.

Worth knowing that this also fires constantly in *legitimate* operation — a Compose file naming
`./config.yaml`, an Ansible `include_vars`, a serverless `${file(...)}`. The "server mode never
touches disk" mental model is wrong in everyday use, not only under attack.

### Related, weaker evidence

`processCertContent` (`pkg/parser/yaml/parser.go:123`) reads PEM files off disk relative to the pushed
file. Verified at parser level only, **not** end-to-end. Narrower reach: `.pem` is not in the pushed
file set and it needs a rule reading a certificate attribute. Same structural fix applies.

## Trade-offs and decisions taken

**Config path filters silently stop applying to absolute paths.** `ignore-paths`/`only-paths` are
matched by `pathutil.MatchesPath` against relative, workspace-rooted patterns, so an absolute path
matches none of them except a leading-`**` one. Under `only-paths` that means *excluded*, not exempt —
and an unscanned file is indistinguishable from a clean one in the response.

Accepted deliberately, and documented in `validateFilePath` rather than left to be discovered. The
justification is that a file outside every workspace folder is not governed by that workspace's
config, and the extension only pushes absolute paths in exactly that situation. The real fix needs a
filter that stops matching an out-of-workspace file against a config it does not belong to, which is
more than this PR should carry. `TestAnalyze_ConfigIgnorePaths` pins the relative case only.

**Mixed absolute and relative paths in one request are allowed.** Not because we want it, but because
per-path validation makes it the default. The extension cannot produce such a request — a path's shape
is a property of its *directory*, and every file in a request comes from one directory. Rejecting the
mix would mean adding a request-level check with exactly the failure shape this PR exists to delete.
A test pins the behaviour as defined rather than accidental, and covers other API clients (IntelliJ,
future ones).

**No interim guard in the extension.** A guard in `toAnalyzeFile` was proposed, implemented, and
reverted in `20a26781d`. The scanner change ships before the extension ships this feature and the
binary tracks GitHub releases with no minimum-version gating, so there is no window where a
feature-enabled extension talks to a scanner that rejects absolute paths. **This depends on the
release ordering actually holding** — if the scanner change slips past the extension release, revisit.

**Out-of-workspace files get a best-effort analysis, and no extension change.** `missing_files`
escalation no-ops for them, because `resolveExtraPaths` returns `[]` when
`vscode.workspace.getWorkspaceFolder` finds no folder. Findings on the opened file still appear; what
is lost is enrichment of resources configured *through* a `module` block pointing outside the pushed
set. That is strictly better than today's whole-request 400. Widening it means widening a file-read
primitive (the scanner names a path, the extension reads it and sends the content back), which
deserves its own ticket.

## Kept as-is

- **`MemFS.Abs` still normalizes rather than absolutizing.** Relative paths stay relative to the
  request's own workspace, which is what keeps module and missing-file paths correct when one server
  process serves many IDE workspaces. Absolute paths stay absolute and are self-disambiguating. Only
  the comment changed.
- **`MemorySourceProvider.GetBasePaths` still returns `["."]`.** It is dead weight on the server path —
  `scan.Client.Scan` never consults it — so it was documented, not changed.
- **The `..` rejection**, unchanged. `../escape.tf` still 400s.
- **Helm resolver and parallel-parsing flags**, unchanged in behaviour; only relocated into
  `serverFlagEvaluator`.
- **The `MemFS.ReadDir` empty-segment fix is cheap correctness, not a reachability fix.** For an
  all-absolute request `dir == "."` is unreachable, since every directory the Terraform readers derive
  stays absolute; only a mixed request could reach it, and the extension cannot produce one. An entry
  named `""` is wrong however it is reached, and skipping empty segments is one line.

## Follow-ups (not in this PR)

1. **Thread `vfs.FS` through `file.Resolver`**, replacing `os.Stat`/`os.Open`. In server mode the memfs
   holds only pushed files, so an unpushed reference simply fails to resolve — the same degradation as
   a missing module, producing a `missing_files` entry the extension already ignores. CLI behaviour is
   unchanged since its `fsys` is the real disk. Same fix for `processCertContent`.
2. **A regression test asserting no disk reads in server mode** — a pushed document referencing an
   on-disk file must resolve to nothing. This is what keeps the invariant honest going forward.
3. **Revisit the CORS posture.** `Access-Control-Allow-Origin: *` with `Access-Control-Allow-Headers: *`
   on an unauthenticated localhost server that accepts caller-supplied Rego is a broad grant
   independent of the resolver bug. Because `/analyze` takes its ruleset from the request body, a
   caller controls both the input document and the rule that echoes it back. Tightening this blunts
   every variant at once. Entirely pre-existing.
4. **Make config path filters handle out-of-workspace files** rather than silently excluding them.

## Tests

Changed:

| Test | Change |
| --- | --- |
| `TestAnalyze_Validation/absolute_path` | 400 → 200, and the path swapped to a parseable `.tf`. This case goes through the real HTTP handler and runs a full scan, so the outcome for an extensionless path with a stub rule was incidental. |

Added:

- `TestValidateFilePath` — POSIX-absolute, Windows drive-qualified, and the `/c:/ws/main.tf` shape VS
  Code sends on Windows, plus the still-rejected shapes. This is where the new boundary is specified.
- An absolute-path twin of `TestAnalyze_ContentPush_TerraformFinding`, asserting the echoed `fileName`
  equals the pushed path. This is a cross-process contract the extension cannot assert from its side.
  The contract is equality after `Clean`/`ToSlash` normalization, **not** byte identity — a client
  sending a native Windows path (`C:\ws\main.tf`) gets `C:/ws/main.tf` back.
- Absolute-path missing-module escalation, asserting `missing_files` comes back absolute.
- A mixed absolute/relative request.
- `serverFlagEvaluator`'s pins, including `IacEnableLocalModuleEval: false`. **This is the test that
  makes the whole change defensible.**
- `MemFS.ReadDir` over absolute keys, covering the empty-segment case.

Unchanged but worth a reviewer's eye: `TestAnalyze_MissingModuleEscalation` and
`TestAnalyze_MissingFilesCWDIndependent` still pass on relative input, but their assertions now encode
"relative in, relative out" rather than a global invariant.

## Review guidance

The load-bearing change is `serverFlagEvaluator` and its test, not the deleted `if`. If you disagree
that pinning the flag is a sufficient replacement for the path-shape check, that is the thing to push
back on.

The second thing worth a hard look is the `only-paths` trade-off under "Trade-offs and decisions
taken" — it is the one behaviour change here that can make a file silently unscanned.

Verified locally on macOS: `go build ./...`, `go vet`, and `go test ./pkg/...` all pass.
`golangci-lint run -c .golangci.yml --timeout 20m` reports **no issues in any file this PR touches**.

A full-repo lint run does surface 51 pre-existing `goconst`/`govet` issues in files this PR does not
touch. The identical 51, with the same breakdown, appear at the merge base — they are an artifact of a
newer golangci-lint locally (v2.12.2) than the v2.4.0 CI pins, and CI will not see them.

**Not verified locally: Windows.** This change is entirely about path handling, and `filepath.IsAbs`
and `filepath.Clean` are OS-dependent — `TestValidateFilePath` covers drive-qualified shapes precisely
because of that. Everything above ran on macOS only, so CI's `windows-2022` unit-test job is the real
check on that axis and the most likely place for this to fail.

K9CODESEC-2171
K9CODESEC-4923

### DIFF
--- a/pkg/server/analyze.go
+++ b/pkg/server/analyze.go
@@ -241,8 +241,8 @@ func ruleLibraryKey(platform string) (string, error) {
 // Absolute paths are accepted. The IDE pushes one when the file it is analyzing
 // lives outside every open workspace folder (File > Open File on a lone .tf),
 // and rejecting it failed the whole request rather than that one file. Nothing
-// downstream needs the path to be relative: content-push mode serves every read
-// from the in-memory FS, and the engine has always been absolute-path clean
+// downstream needs the path to be relative: content-push mode serves pushed file
+// reads from the in-memory FS, and the engine has always been absolute-path clean
 // (the CLI absolutizes all of its input before running the same pipeline).
 //
 // The disk read this check used to stand in for is Terraform local-module
--- a/pkg/server/analyze.go
+++ b/pkg/server/analyze.go
@@ -250,14 +250,15 @@ func ruleLibraryKey(platform string) (string, error) {
 // this check, is what bounds it. It is not the only read that escapes the
 // in-memory FS: the YAML/JSON file resolver opens paths taken from pushed
 // content. That one predates this change and is reachable with a relative path
-// as well, so path shape does not bound it either; it is tracked separately.
+// as well, so path shape does not bound it either; it is tracked separately in
+// K9CODESEC-4924.
 //
 // Known gap: config path filters (ignore-paths/only-paths) are workspace-rooted
 // patterns, so an absolute path matches none of them except a leading-"**" one.
 // Under only-paths that means excluded, not exempt, and an unscanned file is
 // indistinguishable from a clean one in the response. Accepted for now because
 // it needs a filter that no longer matches an out-of-workspace file against the
-// config of a workspace it does not belong to.
+// config of a workspace it does not belong to. Tracked in K9CODESEC-4940.
 func validateFilePath(p string) error {
 	if p == "" {
 		return errors.New("empty file path")
--- a/pkg/server/analyze.go
+++ b/pkg/server/analyze.go
@@ -280,9 +280,9 @@ func validateFilePath(p string) error {
 // Terraform local-module evaluation is off because it reads directories derived
 // from pushed paths off the real disk: tfeval's LoadRootVars and parseDir call
 // os.ReadDir directly instead of going through the request's in-memory FS.
-// Pushed paths may be absolute, so relying
-// on this flag's default — which is remotely togglable — would leave an
-// arbitrary directory readable by a cross-origin caller. Pinning it here is
+// Pushed paths may be absolute, so do not rely on the flag's default staying
+// safe: pinning it here ensures server mode never enables an arbitrary-directory
+// read via local-module evaluation. Pinning it here is
 // what lets validateFilePath accept absolute paths.
 //
 // Parallel file parsing fans the per-file parse across CPUs; off unless the
--- a/pkg/server/analyze.go
+++ b/pkg/server/analyze.go
@@ -39,8 +39,11 @@ const (
 	commonLibraryID = "common"
 )
 
-// analyzeFile is a single pushed file: a workspace-relative path and its raw
-// (possibly unsaved) content.
+// analyzeFile is a single pushed file: its path and its raw (possibly unsaved)
+// content. The path is workspace-relative for a file inside an IDE workspace
+// folder and absolute for one outside every folder. Both are accepted, and
+// findings and missing_files report paths in the shape they were pushed with,
+// after cleaning and forward-slashing (see vfs.MemFS).
 type analyzeFile struct {
 	Path    string `json:"path"`
 	Content string `json:"content"`
@@ -232,8 +235,29 @@ func ruleLibraryKey(platform string) (string, error) {
 	return normalizePlatform(mapped), nil
 }
 
-// validateFilePath rejects paths that are empty, absolute, contain a NUL byte,
-// or escape the workspace root via "..".
+// validateFilePath rejects paths that are empty, contain a NUL byte, or escape
+// the workspace root via "..".
+//
+// Absolute paths are accepted. The IDE pushes one when the file it is analyzing
+// lives outside every open workspace folder (File > Open File on a lone .tf),
+// and rejecting it failed the whole request rather than that one file. Nothing
+// downstream needs the path to be relative: content-push mode serves every read
+// from the in-memory FS, and the engine has always been absolute-path clean
+// (the CLI absolutizes all of its input before running the same pipeline).
+//
+// The disk read this check used to stand in for is Terraform local-module
+// evaluation, pinned off for server mode in serverFlagEvaluator. That pin, not
+// this check, is what bounds it. It is not the only read that escapes the
+// in-memory FS: the YAML/JSON file resolver opens paths taken from pushed
+// content. That one predates this change and is reachable with a relative path
+// as well, so path shape does not bound it either; it is tracked separately.
+//
+// Known gap: config path filters (ignore-paths/only-paths) are workspace-rooted
+// patterns, so an absolute path matches none of them except a leading-"**" one.
+// Under only-paths that means excluded, not exempt, and an unscanned file is
+// indistinguishable from a clean one in the response. Accepted for now because
+// it needs a filter that no longer matches an out-of-workspace file against the
+// config of a workspace it does not belong to.
 func validateFilePath(p string) error {
 	if p == "" {
 		return errors.New("empty file path")
@@ -241,16 +265,34 @@ func validateFilePath(p string) error {
 	if strings.ContainsRune(p, 0) {
 		return errors.New("file path contains NUL byte")
 	}
-	// filepath.IsAbs is OS-dependent; also reject Unix-style absolute paths on
-	// Windows (e.g. "/etc/passwd" is not absolute per Windows rules).
-	if filepath.IsAbs(p) || strings.HasPrefix(p, "/") {
-		return errors.New("file path must be workspace-relative, got absolute: " + p)
-	}
 	clean := filepath.ToSlash(filepath.Clean(p))
 	if clean == ".." || strings.HasPrefix(clean, "../") {
 		return errors.New("file path escapes the workspace: " + p)
 	}
 	return nil
+}
+
+// serverFlagEvaluator pins the feature flags content-push mode depends on.
+//
+// Helm rendering shells out to a chart on disk, which content-push mode has no
+// way to materialize, so the resolver is off.
+//
+// Terraform local-module evaluation is off because it reads directories derived
+// from pushed paths off the real disk: tfeval's LoadRootVars and parseDir call
+// os.ReadDir directly instead of going through the request's in-memory FS.
+// Pushed paths may be absolute, so relying
+// on this flag's default — which is remotely togglable — would leave an
+// arbitrary directory readable by a cross-origin caller. Pinning it here is
+// what lets validateFilePath accept absolute paths.
+//
+// Parallel file parsing fans the per-file parse across CPUs; off unless the
+// server opts in via --x-parallelparsing.
+func serverFlagEvaluator(parallelParsing bool) featureflags.FlagEvaluator {
+	return featureflags.NewLocalEvaluatorWithOverrides(map[string]bool{
+		featureflags.IacEnableKicsHelmResolver:        false,
+		featureflags.IacEnableLocalModuleEval:         false,
+		featureflags.IaCEnableKicsParallelFileParsing: parallelParsing,
+	})
 }
 
 // analyze runs the content-push scan: build an in-memory FS from the pushed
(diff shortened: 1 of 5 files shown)


pkg/server/analyze.go
```diff
@@ -232,27 +235,66 @@ func ruleLibraryKey(platform string) (string, error) {
...
+// in-memory FS: the YAML/JSON file resolver opens paths taken from pushed
+// content. That one predates this change and is reachable with a relative path
+// as well, so path shape does not bound it either; it is tracked separately.
+//
+// Known gap: config path filters (ignore-paths/only-paths) are workspace-rooted
+// patterns, so an absolute path matches none of them except a leading-"**" one.
+// Under only-paths that means excluded, not exempt, and an unscanned file is
+// indistinguishable from a clean one in the response. Accepted for now because
+// it needs a filter that no longer matches an out-of-workspace file against the
+// config of a workspace it does not belong to.
```
MarshalX: Thanks for writing this gap down rather than leaving it implicit. One thought on the consequence: as described, a file pushed under `only-paths` is silently excluded and the IDE renders it as "no problems found". Is a clean-but-unscanned file distinguishable from a genuinely clean one anywhere in the response today?

Since `missing_files` already exists as a channel for "we couldn't fully analyze this", would it make sense to surface filtered-out files there (or in a `skipped_files` sibling), so the gap is observable while the real fix waits?

pkg/vfs/memfs.go
MarshalX, line 225: Should `underDir` special-case a root `dir`? It builds the prefix as `dir + "/"`, so `dir == "/"` becomes `"//"` and `"C:/"` becomes `"C://"`, neither matches any key. Confirmed with a throwaway test:

  ```go
  m := NewMemFS(map[string][]byte{"/main.tf": []byte("x")})
  m.ReadDir("/")   // → readdir /: no such file or directory
  m.MissingFiles() // → ["/"]
  ```

  Reachable via `source = "../"` from a file pushed at `/ws/main.tf`. The user-visible effect is a bogus `"/"` in `missing_files` for the IDE to try to resolve. Would skipping the separator when `dir` already ends in `/` be enough?

  Possibly worth a sibling case in `TestMemFS_ReadDirAbsoluteKeys` too. It covers `ReadDir(".")` nicely, and `ReadDir("/")` is the mirror image of the same idea.

pkg/server/analyze.go
```diff
@@ -232,27 +235,66 @@ func ruleLibraryKey(platform string) (string, error) {
...
+// downstream needs the path to be relative: content-push mode serves pushed file
+// reads from the in-memory FS, and the engine has always been absolute-path clean
+// (the CLI absolutizes all of its input before running the same pipeline).
+//
+// The disk read this check used to stand in for is Terraform local-module
+// evaluation, pinned off for server mode in serverFlagEvaluator. That pin, not
+// this check, is what bounds it. It is not the only read that escapes the
+// in-memory FS: the YAML/JSON file resolver opens paths taken from pushed
+// content. That one predates this change and is reachable with a relative path
+// as well, so path shape does not bound it either; it is tracked separately.
```
MarshalX: > tracked separately

do you mean another Jira ticket?

robertohuertasm-datadog: Yes — filed as [K9CODESEC-4924](https://datadoghq.atlassian.net/browse/K9CODESEC-4924). Confirmed while writing it up that this is pre-existing and reachable via a relative path today (e.g. `x: ../secret.yaml`), independent of this PR — so it's tracked there rather than scoped into #277. Updated the code comment to reference the ticket directly.

pkg/vfs/memfs.go
robertohuertasm-datadog: Good catch, fixed. `underDir` now skips the extra separator when `dir` already ends in `/` (covers both the root `/` and a Windows drive root like `C:/`), so `ReadDir("/")` lists children instead of reporting a miss. Added the `ReadDir("/")` case to `TestMemFS_ReadDirAbsoluteKeys` as the mirror of the existing `ReadDir(".")` coverage, per your suggestion.

pkg/server/analyze.go
robertohuertasm-datadog: Dug into this a bit more: it's not quite pre-existing. Before this PR, `validateFilePath` rejected absolute top-level paths outright, so an absolute pushed file could never reach the `only-paths`/`ignore-paths` filter in server mode — this exact scenario (silently excluded, indistinguishable from clean) only becomes reachable once absolute paths are accepted here. The underlying limitation in `pathutil.Excluded` (workspace-relative patterns vs. an absolute path) does predate this PR, but only shows up elsewhere today (e.g. external Terraform module paths).

No, `missing_files`/`skipped_files` doesn't fit well here either: this isn't a file the engine reached for and couldn't get (the pushed content is present) — it's a file the config's own filter tells it to skip, same as a clean file, semantically.

Filed [K9CODESEC-4924](https://datadoghq.atlassian.net/browse/K9CODESEC-4924) for the YAML/JSON disk-read gap discussed above; its description calls this `only-paths` gap out explicitly as out of scope for that ticket. I don't have a separate ticket open for it yet, so for now it stays as the "Known gap ... Accepted for now" comment already in the code, and we can file something dedicated if it becomes a real problem in practice. Deferring for this PR.

robertohuertasm-datadog: Update: filed [K9CODESEC-4940](https://datadoghq.atlassian.net/browse/K9CODESEC-4940) for this, and referenced it in the code comment.

robertohuertasm-datadog: @MarshalX this is out of the scope of this PR and it's tracked [here](https://datadoghq.atlassian.net/browse/K9CODESEC-4940).

MarshalX (approved):

PR #286, opened 2026-08-11
[K9CODESEC-4561] Reduce local module evaluation memory

whitemerch:
## Motivation

Enabling local module evaluation on large repositories increased memory use enough to contribute to production OOMs. The scan still produced the same findings, but it kept more intermediate state in memory than necessary while resolving Terraform modules and building the OPA input.

This change reduces that overhead without changing scan results. It is a meaningful improvement on the local-module path, but a full multi-platform scan shows that most memory still comes from broader source preparation, so further investigation is still needed.

## Changes

**Local module resolution**

When local modules are instantiated, the scanner now avoids repeating expensive work that does not change the outcome. Module call discovery reuses already-parsed Terraform documents instead of re-parsing directories just to learn which modules are referenced. Reference data used for deduplicating synthetic module documents is computed once per module call chain rather than rebuilt for every resource. After module resolution finishes, evaluator caches are released so that memory can be reclaimed before Rego evaluation starts. Absolute local module paths are now resolved from the repository root instead of being incorrectly appended to the caller directory.

**Scan payload construction**

While converting scanned documents into OPA input, the scanner now preserves shared object structure instead of deep-copying the same maps and arrays many times. When a scan only needs one platform payload that is equivalent to the full payload, it reuses the same in-memory payload instead of building it twice.

## Benchmark example (full multi-platform scan)

Both runs scanned the same large monorepo with all supported platforms enabled and local module evaluation turned on. Findings were identical across both runs.

| Metric | Before | After | Delta |
| --- | --- | --- | --- |
| Peak RSS | 9.80 GB | 9.63 GB | -170 MB (-1.7%) |
| Live heap at query evaluation | 4.41 GB | 3.95 GB | -460 MB (-10%) |
| Scan duration | 4m 38s | 4m 18s | -20s |
| Scanned files | 29,711 | 29,711 | unchanged |
| Parsed files | 25,391 | 25,391 | unchanged |
| Module resources instantiated | 4,135 | 4,135 | unchanged |
| Findings | 10,462 | 10,462 | unchanged |

This confirms the optimization preserves correctness on a realistic full-repo workload. The smaller RSS improvement on a multi-platform scan also shows that most memory is still spent outside local module evaluation, mainly during source preparation across Helm, Kubernetes, YAML, and other platforms, but still a small win.

## Author Checklist

3. [x] All changed-package tests pass.
4. [ ] I have tested my changes on staging, if applicable.

## QA Instruction

1. Run `go test -mod=mod ./pkg/engine ./pkg/parser/terraform/tfeval ./pkg/parser/terraform/modules -count=1`.
2. Run `golangci-lint run -c .golangci.yml --timeout 20m`.
3. Scan a large repository with all platforms enabled and `--x-local-module-eval`, then confirm findings are unchanged and memory use is lower than before.

## Impact

This affects Terraform local module resolution and OPA payload construction. Other platform parsing behavior and rule evaluation logic are unchanged.

## Additional Notes

This is a safe first step toward re-enabling local module evaluation, but it does not fully solve large-repository memory pressure on its own. Follow-up work should focus on multi-platform source preparation and scan concurrency limits in production.

### DIFF
--- a/pkg/engine/module_resolve.go
+++ b/pkg/engine/module_resolve.go
@@ -213,7 +213,7 @@ func resolveModuleDocuments(
 	// Suppression and stripping are driven by actualCalledDirs (evaluation results).
 	staticCalledDirs := make(map[string]bool)
 	for dir := range dirsWithTf {
-		calledDirs := discoverCalledModuleDirs(evaluator, filesByDir[dir], repoPath, resolver, dir)
+		calledDirs := discoverCalledModuleDirs(ctx, evaluator, filesByDir[dir], repoPath, resolver, dir)
 		for _, called := range calledDirs {
 			staticCalledDirs[called] = true
 		}
@@ -464,6 +464,7 @@ func stripModuleCalls(doc model.Document, filePath, repoPath string, calledDirs 
 }
 
 func discoverCalledModuleDirs(
+	ctx context.Context,
 	evaluator *tfeval.Evaluator,
 	files []*model.FileMetadata,
 	repoPath string,
@@ -474,6 +475,10 @@ func discoverCalledModuleDirs(
 	if ok {
 		return calledDirs
 	}
+	contextLogger := logger.FromContext(ctx)
+	contextLogger.Debug().Str("dir", dir).Msg(
+		"module resolve: falling back to directory parse for called module discovery",
+	)
 	return evaluator.CalledModuleDirs(dir)
 }
 
--- a/pkg/engine/module_resolve.go
+++ b/pkg/engine/module_resolve.go
@@ -213,7 +213,8 @@ func resolveModuleDocuments(
 	// Suppression and stripping are driven by actualCalledDirs (evaluation results).
 	staticCalledDirs := make(map[string]bool)
 	for dir := range dirsWithTf {
-		for _, called := range evaluator.CalledModuleDirs(dir) {
+		calledDirs := discoverCalledModuleDirs(evaluator, filesByDir[dir], repoPath, resolver, dir)
+		for _, called := range calledDirs {
 			staticCalledDirs[called] = true
 		}
 	}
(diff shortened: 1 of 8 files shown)


pkg/parser/terraform/tfeval/evaluator.go
```diff
@@ -113,6 +113,13 @@ func (e *Evaluator) SetRemoteResolver(r RemoteResolver) {
 	e.remoteResolver = r
 }
 
+func (e *Evaluator) ReleaseCaches() {
+	e.parseMu.Lock()
+	e.cache = nil
+	e.dirCache = nil
```
MarshalX: Panic risk. Because of nil. Any subsequent write (at line 107 at line 252) will panic if this evaluator is ever touched again. Should we assign fresh empty maps here instead of nil instead?

pkg/parser/terraform/tfeval/evaluator.go
MarshalX, line 169: I saw that we do use `e.parseMu.Lock()` in brand new `ReleaseCaches`, but this read does not uses this lock

MarshalX, line 255: I saw that we do use `e.parseMu.Lock()` in brand new `ReleaseCaches`, but this write does not uses this lock

pkg/engine/module_resolve.go
```diff
@@ -461,6 +463,64 @@ func stripModuleCalls(doc model.Document, filePath, repoPath string, calledDirs
...
+	files []*model.FileMetadata,
+	repoPath string,
+	resolver tfeval.RemoteResolver,
+	dir string,
+) []string {
+	calledDirs, ok := calledModuleDirsFromDocuments(files, repoPath, resolver)
+	if ok {
+		return calledDirs
+	}
+	return evaluator.CalledModuleDirs(dir)
```
MarshalX, line 482: is it worth putting a log message about using the fallback option?

MarshalX: i am right that we do not cover this case with unit tests? we do now fallback to `evaluator.CalledModuleDirs(dir)`? should we cover if so?

MarshalX (approved):

PR #287, opened 2026-08-11
Add Git CLI fallback for reftable refstorage repositories

whitemerch:
## Motivation

Some Git repositories use reftable reference storage, which the embedded Git library used by the scanner cannot open. Without a fallback, scans fail when collecting repository metadata (branch, commit, remote URL) even though the Git CLI can read those repositories normally.

## Changes

**Repository metadata**

When opening a repository fails because of unsupported reftable refstorage, or when the embedded library cannot read the `origin` remote, the scanner now falls back to the Git CLI to resolve the repository root, remote URL, HEAD commit, and branch. A unit test initializes a reftable repository and verifies metadata collection through this path.

## Author Checklist

4. [ ] I have tested my changes on staging (if applicable).

## QA Instruction

1. Run `go test -mod=mod ./cmd/scanner -run TestGetRepositoryCommitInfoWithRefStorage -count=1`.
2. Scan a local repository that uses reftable refstorage and confirm the scan completes with correct repository metadata in the output.

## Impact

This affects repository metadata discovery during CLI scans only. Scanning logic, rule evaluation, and serve mode are unchanged.

## Additional Notes

This is intentionally scoped separately from local module evaluation memory work so it can be reviewed and rolled out on its own.

### DIFF
--- a/cmd/scanner/scan.go
+++ b/cmd/scanner/scan.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 	"time"
@@ -460,15 +461,18 @@ func getRepositoryCommitInfo(repoPaths []string) (*model.RepositoryCommitInfo, s
 	}
 	repo, repoDir, err := openRepo(commonDir)
 	if err != nil {
+		if isUnsupportedRefStorage(err) {
+			return getRepositoryCommitInfoWithGit(commonDir)
+		}
 		return nil, "", fmt.Errorf("error opening the repository: %w", err)
 	}
 
 	remote, err := repo.Remote("origin")
 	if err != nil {
-		return nil, "", fmt.Errorf("error retrieving remote `origin`: %w", err)
+		return getRepositoryCommitInfoWithGit(commonDir)
 	}
 	if len(remote.Config().URLs) == 0 {
-		return nil, "", errors.New("the repository doesn't have a configured remote")
+		return getRepositoryCommitInfoWithGit(commonDir)
 	}
 	out := &model.RepositoryCommitInfo{}
 	out.RepositoryUrl = remote.Config().URLs[0]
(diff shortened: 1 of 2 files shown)


cmd/scanner/scan.go
```diff
@@ -511,6 +515,93 @@ func getRepositoryCommitInfo(repoPaths []string) (*model.RepositoryCommitInfo, s
...
+		return nil, "", fmt.Errorf("error retrieving remote `origin`: %w", err)
+	}
+	sha, err := runGit(repoDir, "rev-parse", "HEAD")
+	if err != nil {
+		return nil, "", fmt.Errorf("error retrieving HEAD ref: %w", err)
+	}
+
+	branch, err := runGit(repoDir, "symbolic-ref", "--quiet", "--short", "HEAD")
+	if err != nil {
+		branch, err = runGit(repoDir, "for-each-ref", "--format=%(refname:short)", "--points-at", "HEAD", "refs/remotes")
```
ChouraquiBen: remark: `for-each-ref` includes `origin/HEAD` simplified by git as `origin`, so in detached checkout state, instead of choosing the branch the commit belongs to, it may select `origin`. TBH, I'm not sure this is a case that will cause much trouble

cmd/scanner/scan.go
```diff
@@ -511,6 +515,93 @@ func getRepositoryCommitInfo(repoPaths []string) (*model.RepositoryCommitInfo, s
...
+		if err != nil {
+			return nil, "", fmt.Errorf("error retrieving reference list: %w", err)
+		}
+		if index := strings.IndexByte(branch, '\n'); index >= 0 {
```
ChouraquiBen: Remark: If multiple branches point at that commit (see above comment), multiple newlines are to be considered and a `\r` in windows might be kept here

ChouraquiBen (dismissed):
Two comments but no real blockers for me

cmd/scanner/scan.go
whitemerch: Yes it can happen, not so much in practice but still worth fixing

ChouraquiBen (approved):
LGTM

PR #288, opened 2026-08-11
[K9CODESEC-3919] Improve custom Rego validation diagnostics

whitemerch:
## Motivation

Custom Rego rule validation in the rule editor reported incomplete, obscure, or imprecise diagnostics. Missing-import detection only worked for two hardcoded library aliases, some compile errors pointed at the wrong source span, and a missing `package datadog` declaration stopped follow-up checks from surfacing other problems in the same file. It's not an easy task as OPA does provide helpful feedback on the diagnosis in the code.

Related tickets: [K9CODESEC-3919](https://datadoghq.atlassian.net/browse/K9CODESEC-3919), [K9CODESEC-3917](https://datadoghq.atlassian.net/browse/K9CODESEC-3917)

## Changes

**Custom Rego validation**

Parse, static AST checks, and OPA compilation now run independently so one phase no longer hides errors from another. When `package datadog` is missing, the validator still recovers the module and continues through structural and compile checks in the same call.

**Missing-import detection**

Import suggestions are derived from the actual platform and common library sources instead of a hardcoded alias table, so missing-import diagnostics work across all supported platforms and with non-conventional alias names.

**Error enrichment**

Compile errors now preserve OPA's source line, narrow unsafe-variable spans to the offending identifier, translate qualified library paths back to the user's alias, and suppress cascade noise such as dependent undefined-variable errors after a missing import is already reported. Parse-error heuristics report missing commas and unclosed calls together, and final diagnostics are sorted by line and column.

**Tests**

Added multi-platform scenario coverage for realistic rule-authoring mistakes, including mixed syntax errors, simultaneous unrelated problems, and the missing-package recovery path.

## Author Checklist

4. [ ] I have tested my changes on staging (if applicable).

## QA Instruction

**Local playground (temporary, for review)**

For interactive QA without staging or web-ui deployment:

1. `export DD_API_KEY=... DD_APP_KEY=...`
2. `make playground`
3. Open http://localhost:8765

Try removing an import, using the wrong package, or dropping a comma, then confirm the actionable diagnostics and line/column highlighting. Click an error in the list to jump to the line.

This dev-only helper is included to make review easier and will be removed before merge once the PR is reviewed and approved.

## Impact

This affects custom Rego validation and evaluation error reporting used by the rule editor backend. Scan execution, default rule evaluation, and CLI scan behavior are unchanged.

## Additional Notes

[K9CODESEC-3919]: https://datadoghq.atlassian.net/browse/K9CODESEC-3919?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[K9CODESEC-3917]: https://datadoghq.atlassian.net/browse/K9CODESEC-3917?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/scan/custom_scan.go
```diff
@@ -751,8 +1042,44 @@ func findUnclosedParentheses(regoContent string) []RegoValidationError {
 	return errs
 }
 
+func countParensOutsideStrings(text string) (open, closing int) {
```
valeryjuli: ❓- I believe backticks ` strings are something [REGO supports too](https://www.openpolicyagent.org/docs/policy-language#strings), might be worth adding the case here

pkg/scan/custom_scan.go
```diff
@@ -143,18 +152,24 @@ func ValidateCustomRegoQuery(
...
+			return parseErrs, nil
+		}
+		compileContent = datadogPackage + "\n\n" + regoContent
+		recovered, _ := parseRegoModule(compileContent)
+		if recovered == nil {
+			return parseErrs, nil
 		}
-		return parseErrs, nil
+		module = recovered
+		lineOffset = -2
```
valeryjuli: 🤔 - Curious, why is it defaulting to -2 as offset ? I noticed in some cases missing closing parentheses are not pointing to the right line but 2 above

Image

pkg/scan/custom_scan.go
```diff
@@ -235,35 +268,118 @@ func shiftErrorLines(errs []RegoValidationError, delta int) {
...
+type libraryInfo struct {
+	path      string
+	functions map[string]bool
+}
+
+func parseLibrary(libCode string) libraryInfo {
+	info := libraryInfo{functions: map[string]bool{}}
+	module, err := ast.ParseModuleWithOpts("library.rego", libCode, ast.ParserOptions{RegoVersion: ast.RegoV1})
+	if err != nil {
+		return info
```
valeryjuli: 💬 - What kind of errors are expected here ? If the library can't be found or downloaded could we log a more descriptive errors ?

valeryjuli (dismissed):
👍 Validated improvements with the live debugger and left some non-blocking comments

While this is a good start, there are multiple LSP functions that are being re-implemented and we might miss some corner cases. In the next steps, we could look into https://github.com/open-policy-agent/regal to replace all the syntax validation

whitemerch:
> 👍 Validated improvements with the live debugger and left some non-blocking comments
> 
> While this is a good start, there are multiple LSP functions that are being re-implemented and we might miss some corner cases. In the next steps, we could look into https://github.com/open-policy-agent/regal to replace all the syntax validation

@valeryjuli Indeed, I forgot about Regal. I will cleanup and use Regal to be adaptive to our use cases

valeryjuli (approved):

PR #289, opened 2026-08-12
Split each file into lines once per file instead of once per document

whitemerch:
## Summary

- Both `sinkContent` (`sink.go`) and `storeResolvedFiles` (`resolver_sink.go`) called `utils.SplitLines` on the *same file content* once for every document produced from that file. For a file with N documents (e.g. a `---`-separated multi-document YAML manifest), that re-computed and separately retained the full file's line slice N times instead of once.
- This surfaced while investigating heavy memory usage on large repos: a heap profile of a full scan of a large internal monorepo showed `strings.genSplit` alone retaining ~1.45 GB of live heap after the parse phase, roughly 30x more string headers than the repo's actual total line count.
- Fix: compute the line split once per file/rendered file, before the per-document loop, and share the same pointer across every document's `FileMetadata`.

## Results

See Benchmark results comment below.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...` (no new issues)
- [x] `go test -race ./pkg/runner/... ./pkg/detector/... ./pkg/engine/... ./pkg/scan/...` (no races; pre-existing unrelated failures only)
- [x] `golangci-lint run -c .golangci.yml --timeout 20m` on changed packages (0 issues)
- [x] Full scan of a large internal repo before/after: identical findings, lower memory, faster wall time

### DIFF
--- a/cmd/scanner/test_rules.go
+++ b/cmd/scanner/test_rules.go
@@ -413,6 +413,9 @@ func parseFixtureFile(ctx context.Context, tmpRoot, filePath, platform string) (
 			continue
 		}
 		docs, _ := p.Parse(ctx, cleanPath, content, true, false, defaultRuleTestMaxResolverDepth)
+		// Computed once per file and shared across every document's
+		// FileMetadata below; docs.Content is identical for every document.
+		linesOriginalData := scanUtils.SplitLines(docs.Content)
 		for _, doc := range docs.Docs {
 			files = append(files, &model.FileMetadata{
 				ID:                uuid.NewString(),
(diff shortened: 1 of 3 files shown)


cmd/scanner/test_rules.go
```diff
@@ -413,6 +413,9 @@ func parseFixtureFile(ctx context.Context, tmpRoot, filePath, platform string) (
 			continue
 		}
 		docs, _ := p.Parse(ctx, cleanPath, content, true, false, defaultRuleTestMaxResolverDepth)
+		// Computed once per file and shared across every document's
+		// FileMetadata below; docs.Content is identical for every document.
+		linesOriginalData := scanUtils.SplitLines(docs.Content)
```
ChouraquiBen, line 418: hmmm, nice

ChouraquiBen (approved):
LGTM

PR #291, opened 2026-08-12
[K9CODESEC-4561] Lazy-load LineInfoDocument to reduce scan memory

whitemerch:
## Motivation

Large scans retain a second full parsed document tree (`LineInfoDocument`, with `_dd_lines` markers) for every file even though only a small fraction of files ever produce a finding that needs line attribution. On a representative large repository scan this duplicate tree accounted for roughly half of peak heap (~6.5 GiB RSS), contributing to out-of-memory failures when other memory-heavy features (such as local Terraform module evaluation) are enabled.

Related Ticket: [K9CODESEC-4561](https://datadoghq.atlassian.net/browse/K9CODESEC-4561?atlOrigin=eyJpIjoiM2EyNDFmMzVlODllNDJjMmE2ZDgxODFkNjExMjg0OWUiLCJwIjoiaiJ9)

## Changes

**Scan pipeline**

During file ingestion, store a re-parse closure on `FileMetadata` instead of eagerly building `LineInfoDocument`. The tree is materialized on first access via `EnsureLineInfoDocument`, memoized with `sync.Once`, and only for files that actually need line resolution.

**Engine**

Call sites that read or mutate `LineInfoDocument` (vulnerability line building, module-instance cloning, suppression handling) now ensure the document is loaded first. `FileMetadata.Clone()` copies fields safely without duplicating `sync.Once` state.

**Resolver paths**

Helm and OpenAPI resolver ingestion still eagerly populate `LineInfoDocument` because rendered resolver content can differ from `OriginalData`, so on-demand re-parse would produce incorrect line maps.

## QA Instruction

See the benchmark that will be posted by DDCI below.

## Blast Radius

CLI and server scan paths that ingest parsed files through the main runner sink. Finding line numbers, suppressions, and SARIF output should be unchanged. Helm/OpenAPI resolver ingestion behavior is unchanged (still eager).

## Additional Notes

[K9CODESEC-4561]: https://datadoghq.atlassian.net/browse/K9CODESEC-4561?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/engine/module_resolve.go
+++ b/pkg/engine/module_resolve.go
@@ -80,7 +80,11 @@ func (c *Inspector) instantiateLocalModules(
 			// drop only those to avoid double-counting; keep the rest of the body
 			// (variable/output/data/locals) so those rules still fire.
 			delete(f.Document, "resource")
-			delete(f.LineInfoDocument, "resource")
+			if err := f.EnsureLineInfoDocument(ctx); err != nil {
+				contextLogger.Err(err).Msgf("failed to build line-info document for file %s", f.FilePath)
+			} else {
+				delete(f.LineInfoDocument, "resource")
+			}
 		}
 		// Only remove local module call-sites that were instantiated; remote/registry
 		// module blocks must remain so the corresponding Rego branches can still fire.
@@ -411,14 +415,20 @@ func buildRefsMap(self *tfeval.ResolvedResource, allInCall []moduleRefEntry) map
 
 // newInstanceFileMetadata clones fm for a synthetic doc (empty Document so Combine skips it).
 func newInstanceFileMetadata(fm *model.FileMetadata, id, callChain string) *model.FileMetadata {
-	clone := *fm
+	clone := fm.Clone()
 	clone.ID = id
 	clone.Document = model.Document{}
 	clone.ModuleCallChain = callChain
 	if fm.LineInfoDocument != nil {
+		// Already materialized: shallow-clone so later mutations on the
+		// parent (e.g. deleting a suppressed "resource" key) don't alias
+		// into this synthetic instance's copy.
 		clone.LineInfoDocument = maps.Clone(fm.LineInfoDocument)
+		clone.LineInfoLoader = nil
 	}
-	return &clone
+	// Otherwise not yet materialized: the clone keeps fm's LineInfoLoader
+	// (copied by Clone) and lazily reconstructs its own copy on demand.
+	return clone
 }
 
 // callChainKey is repo-relative outer caller + "|" + module address (no line numbers, to keep fingerprints stable).
--- a/pkg/engine/module_resolve.go
+++ b/pkg/engine/module_resolve.go
@@ -415,7 +415,7 @@ func buildRefsMap(self *tfeval.ResolvedResource, allInCall []moduleRefEntry) map
 
 // newInstanceFileMetadata clones fm for a synthetic doc (empty Document so Combine skips it).
 func newInstanceFileMetadata(fm *model.FileMetadata, id, callChain string) *model.FileMetadata {
-	clone := fm.Clone()
+	clone := fm.ShallowCopy()
 	clone.ID = id
 	clone.Document = model.Document{}
 	clone.ModuleCallChain = callChain
@@ -424,10 +424,7 @@ func newInstanceFileMetadata(fm *model.FileMetadata, id, callChain string) *mode
 		// parent (e.g. deleting a suppressed "resource" key) don't alias
 		// into this synthetic instance's copy.
 		clone.LineInfoDocument = maps.Clone(fm.LineInfoDocument)
-		clone.LineInfoLoader = nil
 	}
-	// Otherwise not yet materialized: the clone keeps fm's LineInfoLoader
-	// (copied by Clone) and lazily reconstructs its own copy on demand.
 	return clone
 }
 
--- a/pkg/engine/vulnerability_builder.go
+++ b/pkg/engine/vulnerability_builder.go
@@ -107,6 +107,7 @@ var DefaultVulnerabilityBuilder = func(ctx context.Context, qCtx *QueryContext,
 	// searchLine resolution) may need it, so materialize it once up front.
 	if err := file.EnsureLineInfoDocument(ctx); err != nil {
 		logWithFields.Err(err).Msg("failed to build line-info document for file")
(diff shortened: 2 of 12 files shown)


pkg/model/model.go
```diff
@@ -136,10 +137,20 @@ type CommentsCommands map[string]string
...
+	Document         Document
+	LineInfoDocument map[string]interface{}
+	// LineInfoLoader, when set, lazily reconstructs LineInfoDocument (a
+	// second full copy of the parsed document tree, kept intact with
+	// _dd_lines markers purely to resolve a confirmed finding's source line)
+	// on first access via EnsureLineInfoDocument, instead of every file
+	// paying for it eagerly. Only a small fraction of scanned files ever
+	// have a finding, so most never need it. nil means LineInfoDocument is
+	// already populated (or intentionally left unset).
+	LineInfoLoader    func(ctx context.Context, f *FileMetadata) (map[string]interface{}, error)
```
MarshalX: does this break JSON serialization for callers that marshal `Client.Scan()` results (e.g. `Results.Files`)? Might need `json:"-"` here.

pkg/runner/service.go
```diff
@@ -297,6 +297,38 @@ func (s *Service) saveToFile(ctx context.Context, file *model.FileMetadata) {
...
+// FileMetadata, for multi-document files (e.g. "---"-separated YAML).
+func (s *Service) newLineInfoLoader(
+	filename string,
+	docIdx int,
+	openAPIResolveReferences bool,
+	isMinified bool,
+	maxResolverDepth int,
+) func(ctx context.Context, f *model.FileMetadata) (map[string]interface{}, error) {
+	return func(ctx context.Context, f *model.FileMetadata) (map[string]interface{}, error) {
+		reparsed, err := s.Parser.Parse(
```
MarshalX: doesn't the JSON parser's `json.Indent` step make this reparse produce different line numbers than the original eager parse did? Worth double-checking against a minified `.tfplan.json` fixture

pkg/engine/module_resolve.go
```diff
@@ -80,7 +80,11 @@ func (c *Inspector) instantiateLocalModules(
 			// drop only those to avoid double-counting; keep the rest of the body
 			// (variable/output/data/locals) so those rules still fire.
 			delete(f.Document, "resource")
-			delete(f.LineInfoDocument, "resource")
+			if err := f.EnsureLineInfoDocument(ctx); err != nil {
```
MarshalX, line 83: inside `EnsureLineInfoDocument` we delete `"resource"` from `f.Document` but not from `f.LineInfoDocument` (still nil). doesn't that reintroduce the double-counting this code is supposed to prevent?

pkg/runner/service.go
```diff
@@ -297,6 +297,38 @@ func (s *Service) saveToFile(ctx context.Context, file *model.FileMetadata) {
...
+// file's line-info document (the raw parse output, with intact _dd_lines
+// markers, used only to resolve a confirmed finding's source line) by
+// re-parsing the file's own already-retained OriginalData on demand. This
+// avoids every scanned file having to hold a second full copy of its parsed
+// document tree for the whole scan just in case one of its findings needs a
+// line number - most files never have any finding at all.
+//
+// docIdx selects which of the re-parse's documents corresponds to this
+// FileMetadata, for multi-document files (e.g. "---"-separated YAML).
+func (s *Service) newLineInfoLoader(
```
MarshalX: We do capture the whole `*Service`, so anything holding a `FileMetadata` also pins the Inspector/compiled queries/source provider. Should this just capture `s.Parser` instead, to actually get the memory win?

pkg/model/model.go
```diff
@@ -316,6 +373,10 @@ func (m FileMetadatas) Combine(ctx context.Context, lineInfo bool) Documents {
 			continue
 		}
 		if lineInfo {
+			if err := f.EnsureLineInfoDocument(ctx); err != nil {
+				contextLogger.Err(err).Msgf("failed to build line-info document for file %s", f.FilePath)
+				continue
```
MarshalX, line 398: should we fall back to the document without line info instead of `continue`? otherwise it drops the file's document entirely from the line-info payload with only a log line

pkg/engine/vulnerability_builder.go
```diff
@@ -102,6 +102,13 @@ var DefaultVulnerabilityBuilder = func(ctx context.Context, qCtx *QueryContext,
 		Str("queryName", qCtx.Query.Metadata.Query).
 		Logger()
 
+	// LineInfoDocument may be populated lazily (see EnsureLineInfoDocument);
+	// both branches below (searchKey resolution and calculeSearchLine's
+	// searchLine resolution) may need it, so materialize it once up front.
+	if err := file.EnsureLineInfoDocument(ctx); err != nil {
+		logWithFields.Err(err).Msg("failed to build line-info document for file")
```
MarshalX, line 109: should we bump tracker/metric here in additional to err? to not rely on logs only to be able to understand

MarshalX (commented):
rule-test harness never exercises the new lazy path at all. lets add at least one test asserting the reparsed doc matches the original eager doc

pkg/model/model.go
```diff
@@ -158,6 +169,52 @@ type FileMetadata struct {
 	Platform string
 }
 
+// Clone returns a shallow copy of f. FileMetadata carries a sync.Once (for
+// EnsureLineInfoDocument's memoization), so a plain `x := *f` struct-literal
+// copy would duplicate its current state (and trips go vet's copylocks
+// check); Clone instead copies field-by-field and gives the result its own,
+// independent, not-yet-fired memoization state.
+func (f *FileMetadata) Clone() *FileMetadata {
```
MarshalX: Note: i am not quit sure what is the best practice of doing this in golang. brand new added field could be missed here ezly

pkg/engine/module_resolve.go
whitemerch: Double-counting is prevented by `delete(f.Document, "resource")` before OPA runs on Document; `LineInfoDocument` is not fed to queries. `EnsureLineInfoDocument` only runs afterward to mirror that delete in the line-info tree. If `Ensure` fails, `LineInfoDocument` stays nil, that drops line attribution, not finding counts.

MarshalX (approved):

PR #293, opened 2026-08-13
Speed up file discovery for large repositories

whitemerch:
## Motivation

Profiling a large monorepo scan showed most wall time in file discovery and source preparation, not rule evaluation. The analyzer walk, inventory build, YAML path resolver, and Terraform parsing each repeated expensive work (gitignore matching before extension filtering, duplicate `stat` calls, per-call regex compilation, and duplicate HCL parsing for variables/locals).

## Changes

**File analyzer**

The walk filters by scannable extension before running gitignore matching. Helm chart roots are recorded when the walk visits a `Chart.yaml` file instead of calling `Stat` on every directory. The max-file-size check moved from the serial walk into the parallel classification workers.

**Inventory walk**

`checkConditions` now returns the extension it already resolved, so the walk does not call `GetExtension` a second time per file. Shared-walk chart dispatch keeps slash-normalized chart roots so prebuilt inventory renders Helm charts the same way as the per-service walk (including on Windows).

**YAML path resolver**

The serverless `${file(...)}` reference pattern is compiled once at package init. `findFilePath` rejects unsupported extensions before `os.Stat`, avoiding a syscall on every scalar value that is not a file reference.

**Terraform parsing**

Variables and locals are extracted from each `.tf` file in a single parse. Data-source collection skips rewrite-and-parse for files that contain no `aws_iam_policy_document` block. Bare `data.*` references inside quoted strings and `${...}` interpolations are no longer rewritten, so policy ARNs that embed partition/region references still parse correctly.

## QA Instruction

1. `go test ./pkg/analyzer/... ./pkg/engine/provider/... ./pkg/resolver/file/... ./pkg/parser/terraform/... ./pkg/runner/... ./pkg/scan/...`
2. `make build && ./bin/datadog-iac-scanner scan -p <path> -o /tmp/out`
3. Confirm CI passes and the regression benchmark comment on this PR reports no finding-count change, and reports improvements in the wall time.

## Impact

CLI scan startup, file discovery, and Terraform source preparation only. No change to rule logic or SARIF shape for correctly classified files.

### DIFF
--- a/pkg/analyzer/analyzer.go
+++ b/pkg/analyzer/analyzer.go
@@ -182,6 +182,7 @@ type regexSlice struct {
 type analyzerInfo struct {
 	typesFlag       []string
 	filePath        string
+	maxFileSize     int
 	filePlatformMap *sync.Map
 	contentCache    *sync.Map
 	helmCache       *sync.Map
@@ -393,13 +394,14 @@ func Analyze(ctx context.Context, a *Analyzer) (model.AnalyzedPaths, error) {
 				if provider.IsTerraformCacheDir(path) {
 					return filepath.SkipDir
 				}
-				if _, statErr := os.Stat(filepath.Join(path, "Chart.yaml")); statErr == nil {
-					chartRoots = append(chartRoots, filepath.ToSlash(path))
-				}
 				return nil
 			}
 
 			totalFiles++
+
+			if filepath.Base(path) == "Chart.yaml" {
+				chartRoots = append(chartRoots, filepath.ToSlash(filepath.Dir(path)))
+			}
 
 			if d.Type()&fs.ModeSymlink != 0 {
 				if _, statErr := os.Stat(path); statErr != nil {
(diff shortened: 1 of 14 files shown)


NathanGallet-dd (changes requested):
Two correctness issues can suppress findings:

1. pkg/analyzer/analyzer.go: ChartRoots is updated before a Chart.yaml symlink is validated. A dangling Chart.yaml symlink can make prebuilt inventory filter every file below that directory. Add the root only after the file resolves, with a main.tf plus dangling-symlink regression test.

2. pkg/parser/terraform/data_source.go: the raw aws_iam_policy_document byte check is a semantic gate. HCL supports escaped labels such as aws_iam_pol\u0069cy_document, which parses to the same data source but bypasses the gate. Use a parser or lexer-aware precheck, or remove the gate, and add escaped-label coverage.

NathanGallet-dd (approved):

PR #295, opened 2026-08-14
Surface resource_changes/configuration in tfplan documents and resolve pattern filters

ChouraquiBen:
## Summary
- Terraform-plan JSON documents previously exposed only the flattened `planned_values.root_module` resources, dropping `resource_changes[].change.after_unknown` and `configuration` entirely. Rego rules had no way to distinguish a cross-resource reference from genuinely missing configuration — the dominant root cause of false positives/negatives audited in the default-rules repo. Both are now passed through verbatim from the raw parsed document (not the typed `terraform-json` structs, which drop empty slices and add spurious defaulted fields on round-trip).
- Wired `TFPLAN`'s `pattern` attribute through `resolveJSONFilter` (matching existing HCL/`.tf` behavior), so CloudWatch metric-filter patterns in plan JSON get parsed into the `_dd_filter_expr` tree that `common.rego`'s `filter_pattern_clause_is_mandatory` already knows how to consume. Scoped strictly to the top-level `resource` subtree so `resource_changes`/`configuration` remain byte-identical to the raw plan.

## Test plan
- [x] `go build ./...`
- [x] `go test ./pkg/parser/json/... ./pkg/runner/...` (99 passed)
- [x] `go test ./...` (1957 passed, 8 pre-existing/unrelated E2E failures, same baseline as `main`)
- [x] New unit tests: `resource_changes`/`configuration` pass-through (including empty-array edge case), end-to-end `Parser.Parse` → `PrepareScanDocument` pipeline test proving `_dd_lines` doesn't leak into the new fields, and a scoping test proving `pattern` under `resource` resolves to `_dd_filter_expr` while the same key under `configuration` stays raw text
- [x] Standalone `opa eval` checks confirming a real Rego rule can read `after_unknown`/`configuration.root_module.resources[].expressions` and that `common.rego`'s `filter_pattern_clause_is_mandatory` correctly evaluates the real `_dd_filter_expr` tree emitted by `resolveJSONFilter`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/parser/json/tfplan.go
+++ b/pkg/parser/json/tfplan.go
@@ -8,6 +8,7 @@ package json
 import (
 	"encoding/json"
 	"fmt"
+	"regexp"
 
 	"github.com/DataDog/datadog-iac-scanner/pkg/model"
 	hcl_plan "github.com/hashicorp/terraform-json"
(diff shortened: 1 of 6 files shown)


ChouraquiBen:
Closing to open a fresh PR from the same branch for a clean Regression Detector baseline comparison — same code, no changes.

PR #298, opened 2026-08-17
Speed up file discovery by fixing gitignore matching and redundant line counting

whitemerch:
## Motivation

Profiling a large monorepo scan showed most wall time in file discovery and source preparation. Per-file `.gitignore` matching dominated the serial directory walk, scanned files were opened twice (classify + line count), and YAML platform classification unmarshaled full documents for files regex detection had already ruled out.

Two correctness gaps showed up while tightening these paths: root-anchored `.gitignore` patterns were applied in subdirectories, and Helm-rendered virtual template paths failed extension validation because the parser stat'd paths that do not exist on disk.

An intermediate revision reintroduced a large regression by checking every path component against the full pattern list on every file. This branch replaces that with directory pruning during the walk.

## Changes

**File discovery**

`.gitignore` patterns use `go-git`'s matcher with git's last-match-wins precedence. Ignored directories are matched once on the way down and pruned with `SkipDir`, which is both faster than per-file matching and correct: git cannot re-include a file whose parent directory is excluded. File-targeting patterns (e.g. `*.log`) are still tested on each file.

**Line counting**

Lines are counted from content already read for classification, only once a file is known to be scanned.

**Parser extension validation**

`isValidExtension` resolves the extension from the path first and only stats extensionless paths. Helm-rendered YAML on virtual template paths is no longer dropped with a spurious "file not found" error.

**YAML platform classification**

A lexical root-key prefilter skips most YAML files; survivors are parsed as `yaml.Node`. Handles sequence-root Ansible playbooks, merge-key/alias inventory, quoted root keys (decoded and compared, not assumed), vault-encrypted `group_vars`/`host_vars` (stay unclassified), and indentation checked before trimming.

**Classification I/O**

Content-classified extensions use one open for the max-size check and read; the buffer is pre-sized from the stat result instead of growing through `io.ReadAll`.

**Helm chart detection cache**

Every directory visited while walking up to `Chart.yaml` is memoized, not just the file's parent directory.

**Dependencies**

`github.com/sabhiram/go-gitignore` is removed from `go.mod` and `LICENSE-3rdparty.csv`.

## QA Instruction

CI should pass.

Sixteen additional Kubernetes findings are expected: one Helm subchart whose rendered template paths were previously skipped. Those reflect real misconfigurations in rendered manifests, not a classification regression. This is visible on Regression Detector.

## Impact

File discovery, platform classification, Helm rendering, and the line-of-code metric. No change to rule logic or SARIF shape. Finding counts can increase for Helm dependency subcharts whose rendered paths do not exist on disk. Files under a root-anchored `.gitignore` pattern in a subdirectory are now scanned, matching git behavior.

### DIFF
--- a/pkg/analyzer/analyzer.go
+++ b/pkg/analyzer/analyzer.go
@@ -871,32 +871,49 @@ func checkReturnType(ctx context.Context, path, returnType, ext string, content 
 // hc is the scan-scoped cache; when nil the lookup always hits the filesystem.
 func checkHelm(ctx context.Context, path string, hc *sync.Map) bool {
 	contextLogger := logger.FromContext(ctx)
-	dir := filepath.Dir(path)
+	startDir := filepath.Dir(path)
+	if hc != nil {
+		if v, ok := hc.Load(startDir); ok {
+			return v.(bool)
+		}
+	}
+
+	// Every directory walked before the answer is known shares that answer, so
+	// all of them are memoized. This keeps the total number of Chart.yaml probes
+	// proportional to the number of directories in the repository rather than to
+	// directories times their depth, and lets sibling subtrees reuse the walk.
+	walked := []string{startDir}
+	dir := startDir
 	for {
-		if hc != nil {
+		if hc != nil && dir != startDir {
 			if v, ok := hc.Load(dir); ok {
-				return v.(bool)
+				return storeHelmResults(hc, walked, v.(bool))
 			}
 		}
 		_, err := os.Stat(filepath.Join(dir, "Chart.yaml"))
 		if err == nil {
-			if hc != nil {
-				hc.Store(dir, true)
-			}
-			return true
+			return storeHelmResults(hc, walked, true)
 		}
 		if !errors.Is(err, os.ErrNotExist) {
 			contextLogger.Error().Msgf("failed to check helm: %s", err)
 		}
 		parent := filepath.Dir(dir)
 		if parent == dir {
-			if hc != nil {
-				hc.Store(dir, false)
-			}
-			return false
+			return storeHelmResults(hc, walked, false)
 		}
 		dir = parent
+		walked = append(walked, dir)
 	}
+}
+
+func storeHelmResults(hc *sync.Map, dirs []string, result bool) bool {
+	if hc == nil {
+		return result
+	}
+	for _, dir := range dirs {
+		hc.Store(dir, result)
+	}
+	return result
 }
 
 func checkYamlPlatform(ctx context.Context, content []byte, path string) string {
--- a/pkg/analyzer/analyzer.go
+++ b/pkg/analyzer/analyzer.go
@@ -556,7 +556,7 @@ func (a *analyzerInfo) worker(ctx context.Context, results, unwanted chan<- stri
 		return
 	}
 
-	if a.maxFileSize >= 0 {
+	if a.maxFileSize >= 0 && !isContentClassifiedExt(ext) {
 		if info, err := os.Stat(a.filePath); err == nil {
 			if float64(info.Size())/float64(sizeMb) > float64(a.maxFileSize) {
 				contextLogger := logger.FromContext(ctx)
@@ -568,7 +568,11 @@ func (a *analyzerInfo) worker(ctx context.Context, results, unwanted chan<- stri
 		}
 	}
 
-	content, ok := a.readClassifyContent(ctx, ext)
+	content, ok, tooLarge := a.readClassifyContent(ctx, ext)
+	if tooLarge {
+		unwanted <- a.filePath
+		return
+	}
 	if !ok {
 		return
 	}
@@ -607,17 +611,47 @@ func isContentClassifiedExt(ext string) bool {
 	return ext == yaml || ext == yml || ext == json || ext == sh
 }
 
-func (a *analyzerInfo) readClassifyContent(ctx context.Context, ext string) ([]byte, bool) {
+func (a *analyzerInfo) readClassifyContent(ctx context.Context, ext string) (content []byte, ok, tooLarge bool) {
 	if !isContentClassifiedExt(ext) {
-		return nil, true
+		return nil, true, false
 	}
-	content, err := os.ReadFile(a.filePath)
+	file, err := os.Open(filepath.Clean(a.filePath))
 	if err != nil {
 		contextLogger := logger.FromContext(ctx)
 		contextLogger.Error().Msgf("failed to analyze file: %s", err)
-		return nil, false
+		return nil, false, false
 	}
-	return content, true
+	defer func() {
+		if closeErr := file.Close(); closeErr != nil {
+			contextLogger := logger.FromContext(ctx)
+			contextLogger.Err(closeErr).Msgf("failed to close '%s'", a.filePath)
+		}
+	}()
+
+	size := int64(-1)
+	if info, statErr := file.Stat(); statErr == nil {
+		size = info.Size()
+		if a.maxFileSize >= 0 && float64(size)/float64(sizeMb) > float64(a.maxFileSize) {
+			contextLogger := logger.FromContext(ctx)
+			contextLogger.Warn().Msgf(
+				"file %s exceeds maximum file size of %d Mb", a.filePath, a.maxFileSize)
+			return nil, false, true
+		}
+	}
+
+	// Sizing the buffer from the stat already taken above reads the file in one
+	// pass. io.ReadAll would start at 512 bytes and repeatedly double, so every
+	// classified file would cost a handful of allocations and copies instead.
+	var buf bytes.Buffer
+	if size > 0 && int64(int(size)) == size {
+		buf.Grow(int(size) + bytes.MinRead)
+	}
+	if _, err := buf.ReadFrom(file); err != nil {
+		contextLogger := logger.FromContext(ctx)
+		contextLogger.Error().Msgf("failed to analyze file: %s", err)
+		return nil, false, false
+	}
+	return buf.Bytes(), true, false
 }
 
 func (a *analyzerInfo) persistWorkerState(content []byte, platform string) {
--- a/pkg/analyzer/analyzer.go
+++ b/pkg/analyzer/analyzer.go
@@ -931,11 +931,9 @@ func checkYamlPlatform(ctx context.Context, content []byte, path string) string 
 	if utils.IsAnsibleVaultEncrypted(content) {
 		return ""
 	}
-	if checkForAnsibleByPaths(path) {
-		return ansible
-	}
 
-	if !yamlRootHasAnyKey(content, yamlPlatformRootKeys...) {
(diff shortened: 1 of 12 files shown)


pkg/analyzer/gitignore.go
```diff
@@ -0,0 +1,71 @@
...
+
+func (m *gitIgnoreMatcher) matches(relPath string, isDir bool) bool {
+	if m == nil {
+		return false
+	}
+	normalized := strings.Trim(filepath.ToSlash(relPath), "/")
+	if normalized == "" || normalized == "." {
+		return false
+	}
+	return m.matcher.Match(strings.Split(normalized, "/"), isDir)
```
NathanGallet-dd: [P2] Preserve re-inclusions below trailing /** patterns. The go-git matcher reports abc/** as matching the directory abc itself, whereas git check-ignore does not: that pattern only ignores its contents. Because Analyze prunes any directory for which MatchesDir returns true, patterns abc/**, !abc/def/, and !abc/def/keep.tf will prune abc and never scan keep.tf even though Git re-includes it. Please avoid pruning the parent for trailing /** (or otherwise align directory matching with Git) and add this case to the matcher tests.

pkg/analyzer/yaml_classify.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+// yamlRootHasAnyKey reports whether content appears to declare any of keys at
+// the document root, without parsing the full YAML tree.
+func yamlRootHasAnyKey(content []byte, keys ...string) bool {
+	if len(content) == 0 {
+		return false
+	}
+	content = bytes.TrimPrefix(content, []byte{0xEF, 0xBB, 0xBF})
+	for _, line := range bytes.Split(content, []byte("\n")) {
```
NathanGallet-dd, line 26: [P2] Keep the prefilter conservative for flow-style roots. This line-by-line scan only recognizes block-style root keys or dash-prefixed block sequences, so valid YAML such as {resources: []} and [{hosts: all}] returns false here. checkYamlPlatform then skips parsing, and GDM or Ansible files that were previously detected become unclassified. Root-level merges whose matching key exists only inside an anchor have the same problem. Please cover flow and merge syntax before returning early, with regression tests.

pkg/analyzer/gitignore.go
```diff
@@ -0,0 +1,110 @@
...
+
+	var filePatterns, dirPatterns []gitignore.Pattern
+	for _, line := range strings.Split(string(content), "\n") {
+		line = strings.TrimSuffix(line, "\r")
+		if strings.HasPrefix(line, "#") || strings.TrimSpace(line) == "" {
+			continue
+		}
+		// A nil domain anchors the patterns at the repository root, which is
+		// where the parsed file lives.
+		if !isDirectoryOnlyPattern(line) {
```
NathanGallet-dd: [P2] Preserve descendants of re-included wildcard directories

Removing directory-only patterns from `filePatterns` breaks Git semantics when a wildcard ignores a directory and a later directory-only negation reopens it. For example, with `foo/*` followed by `!foo/bar/`, Git does not ignore `foo/bar/main.tf`, but this matcher enters `foo/bar` and then `MatchesPath` still applies `foo/*` to `foo/bar/main.tf` (the go-git glob matcher accepts the matching prefix), so `Analyze` excludes the file. Please preserve the directory re-inclusion for descendant file checks and add this case to the matcher tests.

pkg/analyzer/yaml_classify.go
```diff
@@ -0,0 +1,230 @@
...
+		return true
+	}
+	if quote := trimmed[0]; quote == '"' || quote == '\'' {
+		return quotedRootKeyIsAny(trimmed, quote, keys...)
+	}
+	return plainRootKeyIsAny(trimmed, keys...) || plainRootKeyIsAny(trimmed, "<<")
+}
+
+func yamlRootLineContent(line []byte) []byte {
+	if len(line) > 0 && (line[0] == ' ' || line[0] == '\t') {
```
NathanGallet-dd: [P2] Keep valid root-indented YAML eligible for parsing

YAML accepted by `gopkg.in/yaml.v3` can begin its root node at non-zero indentation; for example, `  resources: []` is a root mapping and the previous parse-first path classified it as Google Deployment Manager. Returning `nil` for every indented line makes the lexical prefilter return false, so this file is now unclassified (the same applies to an indented `playbooks:` root). Please track the root indentation or conservatively fall back to parsing, and add a regression test.

pkg/analyzer/analyzer.go
```diff
@@ -849,52 +926,32 @@ func checkYamlPlatform(ctx context.Context, content []byte, path string) string
...
-		contextLogger.Warn().Msgf("failed to parse yaml file (%s): %s", path, err)
-		return ""
-	}
-
-	// Get the yaml content in node
-	contentNode := &node
-	if node.Kind == yamlParser.DocumentNode && len(node.Content) > 0 {
-		contentNode = node.Content[0]
+	if checkForAnsibleByPaths(path) {
+		return ansible
```
NathanGallet-dd: [P2] Preserve the empty/non-mapping guard for vars files

Returning `ansible` solely from the path now classifies empty or comment-only `group_vars` and `host_vars` files that the previous parse-first path rejected as scalar/empty roots. Those files are added to the inventory and then the YAML parser returns `no documents found in yaml file`, producing an avoidable parse error (arbitrary malformed content under these directories is similarly accepted here). Please retain validation of a non-empty mapping root for path-classified vars files while keeping encrypted vaults excluded, and cover an empty/comment-only vars file.

MarshalX:
Why violation count drift? Is it expected?

whitemerch:
> Why violation count drift? Is it expected?

@MarshalX It is expected, and I have put it in the PR description. The more verbose explanation is that Helm renders subchart templates in memory. After render, those manifests get paths like `charts/<dependency>/templates/deployment.yaml`. Those paths are logical, the content exists in memory, but there is no file on disk at that location.

Before this PR, `isValidExtension` always called stat on the path before parsing. For those virtual Helm paths, stat failed with “file not found”, extension validation returned false, and the parser bailed out with `ErrNotSupportedFile`. The rendered YAML was dropped even though we already had the content.

That’s why those ~16 Kubernetes findings were missing: we never scanned manifests from dependency subcharts whose rendered paths don’t exist on disk

pkg/analyzer/analyzer.go
```diff
@@ -390,10 +389,26 @@ func Analyze(ctx context.Context, a *Analyzer) (model.AnalyzedPaths, error) {
...
 			if d.IsDir() {
 				if provider.IsTerraformCacheDir(path) {
 					return filepath.SkipDir
 				}
+				// Pruning the subtree is cheaper than testing every file below
+				// it, and it is what makes an ignored directory final: git
+				// cannot re-include a file whose parent is excluded, so the
+				// files under it must never be considered at all. The directory
+				// itself is recorded as excluded so downstream walks skip it too.
+				if insideRepo && hasGitIgnoreFile && gitIgnore.MatchesDir(trimmedPath) {
```
NathanGallet-dd, line 420: [P2] Check ignored ancestors when a directory is the walk root

When `Paths` starts below an ignored directory (for example, `RepoPath=/repo`, `Paths=[/repo/build/nested]`, and `.gitignore` contains `build/`), `WalkDir` never visits `build`. The root callback only asks `MatchesDir("build/nested")`; for this slashless directory pattern that checks the basename `nested` and returns false. Because `singleFilePath` is also false, the files below are scanned even though Git ignores them. I reproduced this through `Analyze` with `build/nested/main.tf`. Please check parent directories for a directory walk root as well (before descending), and add that case to the tests.

pkg/analyzer/yaml_classify.go
```diff
@@ -0,0 +1,238 @@
...
+				return true
+			}
+		}
+		return false
+	}
+	if root.Kind != yamlParser.MappingNode {
+		return false
+	}
+	playbooks := yamlMapKeyNode(root, playBooks)
+	if playbooks != nil && playbooks.Kind == yamlParser.SequenceNode {
```
NathanGallet-dd, line 227: [P2] Dereference alias-backed playbook sequences

A valid Ansible wrapper such as `plays: &plays` followed by `playbooks: *plays` now leaves `playbooks` as an `AliasNode`, so this `SequenceNode` check fails and the file becomes unclassified. The previous `model.Document` path resolved this alias into the `playbooks` slice and detected the play via its `hosts` key. I reproduced the current result as an empty platform. Please dereference aliases before the kind check (and likewise before iterating the sequence) and add a regression test for an alias-backed `playbooks` value.

pkg/analyzer/gitignore.go
```diff
@@ -0,0 +1,145 @@
...
+		}
+	}
+
+	// A trailing /** requires at least one path component below its parent.
+	if strings.HasSuffix(body, "/**") {
+		body = strings.TrimSuffix(body, "/**") + "/*/**"
+	}
+	const pathEnd = "\x00"
+	return gitIgnorePattern{
+		pattern:       gitignore.ParsePattern(inclusion+body+"/"+pathEnd, nil),
```
NathanGallet-dd, line 74: [P2] Preserve recursive ignores after reopening a directory

Appending the sentinel makes go-git fail to backtrack for patterns such as `foo/**/*`. With `foo/**/*`, `!foo/bar/`, and `!foo/bar/keep.tf`, Git still ignores `foo/bar/a.tf` (only `keep.tf` is re-included), but this matcher reports `foo/bar/a.tf` as included: the `**/*` portion consumes `bar` and then the sentinel fails against `a.tf`. `Analyze` therefore scans a file Git excludes. I reproduced this by differential-testing `ignoredDuringWalk` against `git check-ignore`. Please preserve matching for recursive wildcard patterns when a later negation reopens the directory, and add this combination to the tests.

pkg/analyzer/yaml_classify.go
```diff
@@ -0,0 +1,246 @@
...
+	playbooks := yamlResolveAlias(yamlMapKeyNode(root, playBooks))
+	if playbooks != nil && playbooks.Kind == yamlParser.SequenceNode {
+		for _, item := range playbooks.Content {
+			if ansiblePlayKeywordsFromNode(item) {
+				return true
+			}
+		}
+	}
+	for _, hostKey := range ansibleHost {
+		hosts := yamlMapKeyNode(root, hostKey)
```
NathanGallet-dd: [P2] Resolve aliases for inventory host groups

The playbook path now dereferences aliases, but the inventory path does not. For valid YAML such as `inventory: &inventory` with a `hosts:` mapping followed by `all: *inventory`, `yamlMapKeyNode` returns an `AliasNode`; this kind check skips it and `checkYamlPlatform` returns an empty platform. The previous document conversion resolved the alias and classified the file as Ansible. Please call `yamlResolveAlias` on the host-group value and add an alias-backed inventory regression test.

NathanGallet-dd (approved):

PR #309, opened 2026-08-24
[K9CODESEC-5290] Harden HTTP and registry module fetches against SSRF

whitemerch:
## Motivation

Remote module fetching used default HTTP clients with no destination policy: private, loopback, link-local, and metadata addresses were reachable, redirects and `X-Terraform-Get` hops were not revalidated, and DNS rebinding was possible because hostname checks were not bound to the connection. This change adds a shared HTTP destination policy for registry and go-getter HTTP paths so module acquisition cannot be steered to internal endpoints.

Related Ticket: [K9CODESEC-5290](https://datadoghq.atlassian.net/browse/K9CODESEC-5290)

## Changes

**HTTP destination policy**

Introduce a shared policy that rejects non-public unicast destinations, validates every redirect hop, resolves DNS once per hop, dials only validated addresses, deduplicates DNS answers, and bounds concurrent dial attempts. Proxy bypass is disabled on the secured transport.

**Registry client**

Wire the policy into Terraform registry discovery, version lookup, and download requests. The registry cache is built after resolver options are finalized so the host allowlist applies consistently.

**go-getter HTTP**

Route HTTP/HTTPS module fetches through the policy-controlled client. HTTP-origin `X-Terraform-Get` indirection stays within HTTP(S) so nested hops cannot switch to git, file, or cloud getters. Git, SSH, S3, and GCS transports remain unchanged for a follow-up.

## Author Checklist

Tested on repository A, and the scan time/memory/finding count is the same as before.

## QA Instruction

1. `go test ./pkg/parser/terraform/modules/resolver/... ./pkg/scan/...`
2. `go test -race ./pkg/parser/terraform/modules/resolver`
3. `make build && ./bin/datadog-iac-scanner scan -p <terraform-fixture> -t Terraform -o /tmp/out --x-remote-modules --x-remote-modules-allowed-host registry.terraform.io`
4. Confirm resolver tests cover direct private-IP rejection, redirect validation, `X-Terraform-Get` private destinations, and allowlist enforcement.

## Blast Radius

This affects only the Datadog IaC Scanner remote-module resolver when network fetching is enabled (`--x-remote-modules`). It does not change rule assets, SARIF shape, or subprocess git/SSH/S3/GCS fetch behavior.

## Additional Notes

[K9CODESEC-5290]: https://datadoghq.atlassian.net/browse/K9CODESEC-5290?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[K9CODESEC-2662]: https://datadoghq.atlassian.net/browse/K9CODESEC-2662?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/parser/terraform/modules/resolver/archive_extract_test.go
+++ b/pkg/parser/terraform/modules/resolver/archive_extract_test.go
@@ -8,7 +8,9 @@ package resolver
 import (
 	"archive/tar"
 	"bytes"
+	"io"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"testing"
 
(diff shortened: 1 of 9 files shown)


NathanGallet-dd (commented):
P1: Calling archive.Output buffers the full tar stream before extractRegularFilesWithBudget checks maxArchiveExtractBytes. A module repository can therefore make the scanner allocate an arbitrarily large archive in memory, well beyond the stated 200 MiB cap, and be OOM-killed before the extraction limit runs. Stream stdout directly into the bounded extractor, with a reader that enforces the cap while reading, rather than materializing the archive in memory.

NathanGallet-dd (commented):
P1: The special-use denylist omits 64:ff9b:1::/48, the RFC 8215 local-use IPv4/IPv6 translation prefix. It passes validatePublicAddress because Go treats it as global unicast and it is neither private nor link-local. An attacker can therefore use a literal or DNS answer in that prefix to have the scanner reach an internal endpoint through the environment's local NAT64 translator, bypassing this SSRF boundary. IANA marks the prefix as not globally reachable; add it to the blocked prefixes and cover it with a regression test.

NathanGallet-dd (commented):
P1: The denylist is still incomplete for IANA IPv6 ranges that Go classifies as global unicast but IANA marks as not globally reachable. For example, 2001:2::/48 is the benchmarking prefix and currently passes validatePublicAddress, so a module URL can target a service routed inside the scanner's administrative domain through that prefix. The same gap remains for 100:0:0:1::/64, 3fff::/20, and 5f00::/16. Block the non-globally-reachable entries from the IANA special-purpose registry rather than maintaining a partial list.

NathanGallet-dd (approved):

PR #310, opened 2026-08-24
[K9CODESEC-5291] Enforce destination policy on git and cloud module transports

whitemerch:
## Motivation

Remote module fetching still had network paths that could bypass the [ticket-2 HTTP destination policy](https://github.com/DataDog/datadog-iac-scanner/pull/309): subprocess git cloned over SSH or raw HTTPS, and go-getter exposed S3, GCS, Mercurium, and network git transports without equivalent validation.

Related ticket: [K9CODESEC-5291](https://datadoghq.atlassian.net/browse/K9CODESEC-5291)

## Changes

**HTTPS git transport**

BareGit now keeps `git::https://` sources on HTTPS and routes clone/fetch subprocesses through a local CONNECT proxy backed by the existing destination policy and IP pinning. Git environment overrides are stripped, cached bare repos with unsafe local config are rejected, and embedded credentials in module URLs are refused.

**Disabled unsafe transports**

SSH git, go-getter network git, S3, GCS, and Mercurial transports are rejected with explicit resolution errors because they cannot enforce the same connection-time policy today.

**Tests**

Added SSRF, DNS pinning, proxy lifecycle, transport rejection, and cached-config regression coverage for the new paths.

## Author Checklist

## QA Instruction

1. `go test -race ./pkg/parser/terraform/modules/resolver ./pkg/scan -count=1`
2. Confirm `git::https://...?ref=...` modules still resolve when the host is public and allowed.
3. Confirm `git::ssh://...`, `s3::...`, and `gcs::...` sources fail with explicit transport-disabled errors.

## Blast Radius

This PR only affects remote Terraform module fetching when `--x-remote-modules` is enabled. Public HTTPS git modules continue to work through the policy proxy; SSH, S3, GCS, and unpinned network git transports are now blocked instead of silently bypassing the security boundary.

## Additional Notes

[K9CODESEC-5291]: https://datadoghq.atlassian.net/browse/K9CODESEC-5291?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/parser/terraform/modules/resolver/gogetter.go
+++ b/pkg/parser/terraform/modules/resolver/gogetter.go
@@ -79,7 +79,10 @@ type GoGetterConfig struct {
 	HostAllowlist []string
 	Cache         *moduleCache
 	RegistryCache *registryCache
-	httpClient    *http.Client
+	// Git resolves pinnable git:: download URLs (registry X-Terraform-Get and
+	// default-branch sources) through BareGit instead of an unpinned git subprocess.
+	Git        Resolver
+	httpClient *http.Client
 
 	TmpDir string
 
@@ -126,7 +129,7 @@ func (r *GoGetterResolver) Resolve(ctx context.Context, mod *tfmodules.ParsedMod
 	if mod.IsLocal {
 		return Resolution{}, &tfmodules.UnresolvedError{Reason: "local modules are handled by LocalResolver"}
 	}
-	// BareGit owns git:: sources with ref=. Falling through after a BareGit failure
+	// BareGit owns pinnable git:: sources. Falling through after a BareGit failure
 	// would trigger a full working-tree clone of the entire repository per module.
 	if bareGitOwnsSource(mod.Source) {
 		return Resolution{}, &tfmodules.UnresolvedError{
@@ -185,6 +188,17 @@ func (r *GoGetterResolver) fetchAndCommit(
 	}
 	if selectedSubdir == "" && sourceType == sourceTypeRegistry {
 		selectedSubdir = registrySubdir(mod.Source)
+	}
+	if gitMod, ok := pinnableGitModuleFromGetterSource(packageSource, selectedSubdir); ok {
+		if r.cfg.Git == nil {
+			return Resolution{}, &tfmodules.UnresolvedError{
+				Reason: "pinned git download must be resolved by BareGitResolver",
+			}
+		}
+		if err := r.checkAllowlist(packageSource); err != nil {
+			return Resolution{}, err
+		}
+		return r.cfg.Git.Resolve(ctx, gitMod)
 	}
 	if err := validateGetterSourceTransport(packageSource); err != nil {
 		return Resolution{}, err
--- a/pkg/parser/terraform/modules/resolver/gogetter.go
+++ b/pkg/parser/terraform/modules/resolver/gogetter.go
@@ -35,6 +35,7 @@ const (
 	mib                   = 1024 * 1024
 
 	sourceTypeRegistry = "registry"
+	sourceTypeGit      = "git"
 
 	defaultFetchConcurrency = 32 // network-bound; override via IAC_MODULE_FETCH_CONCURRENCY
 )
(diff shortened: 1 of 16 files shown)


pkg/parser/terraform/modules/resolver/gogetter.go
```diff
@@ -185,6 +186,9 @@ func (r *GoGetterResolver) fetchAndCommit(
 	if selectedSubdir == "" && sourceType == sourceTypeRegistry {
 		selectedSubdir = registrySubdir(mod.Source)
 	}
+	if err := validateGetterSourceTransport(packageSource); err != nil {
```
NathanGallet-dd, line 203: P1: This rejects standard Terraform Registry modules whose download endpoint returns `X-Terraform-Get: git::https://…?ref=<sha>`. For example, the public `terraform-aws-modules/vpc/aws` endpoint returns exactly that form. Because the original module source is a registry address, it never reaches `BareGitResolver`; `getterSourceFor` obtains the git URL here and `validateGetterSourceTransport` then rejects it before fetch. As a result, commonly used pinned Registry modules fail whenever remote modules are enabled. Please either route registry-returned pinned HTTPS git URLs through the policy-enforcing BareGit path, or preserve a policy-enforced getter for this case.

NathanGallet-dd, line 203: P1: This also turns every direct HTTPS git source without `ref=` into an unresolved module. Those sources are deliberately excluded from `BareGitResolver` (`bareGitOwnsSource` requires a ref), then reach this check as `git::https://…` and are rejected. A default-branch HTTPS git source is valid Terraform configuration and was supported before this PR; the stated blast radius says public HTTPS git modules continue to work. It needs a policy-enforced path for the no-ref case instead of rejecting it solely because go-getter would invoke git.

pkg/parser/terraform/modules/resolver/gogetter.go
```diff
@@ -185,6 +189,20 @@ func (r *GoGetterResolver) fetchAndCommit(
...
+	if gitMod, ok := pinnableGitModuleFromGetterSource(packageSource, selectedSubdir); ok {
+		if r.cfg.Git == nil {
+			return Resolution{}, &tfmodules.UnresolvedError{
+				Reason: "pinned git download must be resolved by BareGitResolver",
+			}
+		}
+		if err := r.checkAllowlist(packageSource); err != nil {
+			return Resolution{}, err
+		}
+		return r.cfg.Git.Resolve(ctx, gitMod)
```
NathanGallet-dd, line 201: P1: Delegating here skips all of GoGetter’s byte accounting (`reserveDirBytes` / `MaxTotalBytes`). `--x-remote-modules-max-bytes` is documented as the total download-size limit across *all* remote modules, but a scan can now resolve arbitrarily many Registry modules whose `X-Terraform-Get` is `git::https://…` and each is materialized by `BareGit` outside that shared budget (only its independent 200 MiB-per-archive limit applies). That reopens an unbounded disk/network-consumption path for the most common Registry module transport. Please give `BareGit` the same shared budget or account the resolved package before returning it.

pkg/parser/terraform/modules/resolver/gogetter.go
```diff
@@ -171,6 +175,27 @@ func (r *GoGetterResolver) Resolve(ctx context.Context, mod *tfmodules.ParsedMod
...
+		}
+	}
+	if err := r.checkAllowlist(packageSource); err != nil {
+		return Resolution{}, err
+	}
+	res, err := r.cfg.Git.Resolve(ctx, gitMod)
+	if err != nil {
+		return Resolution{}, err
+	}
+	if err := r.reserveDirBytes(res.PackageRoot); err != nil {
```
NathanGallet-dd: P1: This only accounts Git downloads that reach `GoGetterResolver` (the Registry `X-Terraform-Get` handoff). A direct no-ref source such as `git::https://github.com/org/repo.git` is claimed by `BareGitResolver` earlier in the resolver chain (`buildModuleResolverChain` adds it before GoGetter), so this helper is never called. Those newly supported default-branch sources therefore still bypass `--x-remote-modules-max-bytes` entirely. The shared budget needs to be enforced by/wrapped around `BareGitResolver`, rather than only after this delegation.

whitemerch: Indeed, but I have a dedicated [ticket](https://datadoghq.atlassian.net/browse/K9CODESEC-5288?atlOrigin=eyJpIjoiZThmMzIxYWY4M2RmNGNiZDlhNTdhY2E0Y2JmMDhjYTMiLCJwIjoiaiJ9) for it

NathanGallet-dd (approved):

PR #316, opened 2026-08-25
[K9CODESEC-5292] Make resolver concurrency and caches deterministic

whitemerch:
## Motivation

Shared go-getter instances could steal another fetch's deadline, canceled parses could be cached as empty, and module caches grew without a size bound while a single DNS blip banned a registry for 24 hours.

Related ticket: [K9CODESEC-5292](https://datadoghq.atlassian.net/browse/K9CODESEC-5292)

## Changes

**Resolver concurrency**

Each fetch now builds its own go-getter map instead of mutating package globals, HTTP getters honor the package byte ceiling in flight, and a zero or negative per-host concurrency setting falls back to the default cap of 6.

**Parse and resolution caches**

Canceled module parses and canceled remote resolutions are not stored, and the parse semaphore is scoped to one scan so a later retry can still discover modules.

**On-disk and registry caches**

Equivalent registry and git source spellings share a cache key while different versions do not. New package stores record size metadata and only reconcile or evict when the running total exceeds the limit, so cache hits do not walk the tree. Registry discovery failures stay in-process with a short backoff and are no longer written to `dead-registry-hosts.json`.

**CLI**

`--x-remote-modules-cache-dir` and `--x-remote-modules-cache-max-bytes` (default 2GiB) select the cache root and bound its size.

## Author Checklist

## QA Instruction

1. `go test -race ./pkg/parser/terraform/modules/resolver ./pkg/parser/terraform/modules/modulegraph -count=1`
2. `go test ./pkg/scan ./cmd/scanner -count=1`
3. Confirm `TestGoGetterResolveRaceWith32ConcurrentFetches` and `TestModuleCacheHitsDoNotEvictSiblings` pass.

## Blast Radius

This PR only changes Terraform remote-module fetching, caching, and experimental CLI flags. Scan results for repositories that do not enable `--x-remote-modules` are unchanged.

[K9CODESEC-5292]: https://datadoghq.atlassian.net/browse/K9CODESEC-5292?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cmd/scanner/scan.go
+++ b/cmd/scanner/scan.go
@@ -153,6 +153,17 @@ var scanAction = &cli.Command{
 			Hidden: true,
 			Usage:  "(experimental) target bytes admitted for repository and remote module parsing (default disabled)",
 		},
+		&cli.StringFlag{
+			Name:   "x-remote-modules-cache-dir",
+			Hidden: true,
+			Usage:  "(experimental) directory for fetched Terraform module caches",
+		},
+		&cli.Int64Flag{
+			Name:   "x-remote-modules-cache-max-bytes",
+			Hidden: true,
+			Usage:  "(experimental) maximum on-disk size of the fetched Terraform module cache (default 2GiB)",
+			Value:  scan.DefaultRemoteModuleMaxCacheBytes,
+		},
 		&cli.BoolFlag{
 			Name:   "x-terraform-plan",
 			Hidden: true,
@@ -348,7 +359,8 @@ func runScan(ctx context.Context, c *cli.Command) error {
 	}
 	if !c.Bool("x-remote-modules") &&
 		(c.String("x-remote-modules-manifest") != "" ||
-			len(c.StringSlice("x-remote-modules-allowed-host")) > 0) {
+			len(c.StringSlice("x-remote-modules-allowed-host")) > 0 ||
+			c.String("x-remote-modules-cache-dir") != "") {
 		return errorWithExitCode(
 			errors.New("remote module resolver options require --x-remote-modules"),
 			constants.InvalidConfigErrorCode,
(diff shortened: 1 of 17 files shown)


NathanGallet-dd:
P1: the aggregate cache budget still does not enforce its cap for an oversized git entry. `ModuleCacheBudget.AdmitIfFits` simply returns when a `BareGitResolver`/`LocalGitRefResolver` entry exceeds `MaxBytes`; it neither rejects the resolution nor removes the just-published entry. In a scan that resolves only that module, no later `Admit` necessarily runs, so the oversized `git-bare`/`git-local` directory remains on disk above `--x-remote-modules-cache-max-bytes` indefinitely. (The new test confirms it remains until an explicit later `Admit("")`.) Please reject or remove an entry that cannot fit before returning a successful resolution.

NathanGallet-dd:
P1: the short-lived pins allow cache eviction to delete modules that the active scan still needs. The pins in `lookupCache`/`commitFetchedDir` and the git resolvers end as soon as `ConfineResolution` returns, but the module graph keeps the returned paths and only parses/scans them later (after it resolves the rest of the graph). A later `Budget.Admit` can therefore evict an earlier, now-unpinned package while it is still in `Result.ScanPaths`/`Result.Modules`; the later parse or engine scan then reads a removed directory and silently loses that module. The publish-to-pin gap is an additional race, but even making that handoff atomic would not fix the lifetime problem. Keep entries pinned until the graph/scan cleanup runs (for example, via `Resolution.Cleanup`), or do not evict entries belonging to the active scan.

NathanGallet-dd:
Follow-up on the current `18ec9c61` implementation — the two P1s remain:

1. **Oversized git entries remain on disk.** `AdmitIfFits` returns when `cacheEntrySize(path) > MaxBytes`; it does not reject the resolution or remove the entry. If the scan resolves only that oversized git module, no subsequent admission is required, so the cache stays above `--x-remote-modules-cache-max-bytes`.

2. **Pins do not last through scan consumption.** `lookupCache` and `commitFetchedDir` defer the lease release before returning the `Resolution`; the git resolvers likewise unpin on return. The module graph retains those paths and scans them later, so a later admission can evict an already-resolved module before it is parsed/scanned.

The cache needs either scan-lifetime pins released by cleanup, or eviction must exclude active-scan entries; oversized git entries need rejection or removal before a successful resolution is returned.

NathanGallet-dd (approved):

PR #321, opened 2026-08-26
Fix exponential memory growth in local Terraform module resolution

whitemerch:
## Motivation

Local module resolution could OOM scan pods when a repository used deeply nested modules with identical inputs reached by many call paths. The evaluation cache keyed on the module call chain, so the same module was evaluated and retained once per path (branch^depth) instead of once per distinct directory and inputs. A single pathological repository could exhaust a 40 GiB pod while typical scans stayed under 2 GiB.

## Changes

**Module evaluation (`tfeval`)**

Memoize completed module evaluations on `(directory, package root, resolved inputs)` and re-stamp module address and call chain onto cached results when another path reaches the same module. Do not cache evaluations that skipped descendants (depth, cycle, or budget), so a later valid call site cannot reuse a truncated subtree. Enforce the instantiation budget inside recursive evaluation, reserve quota before allocating expanded resources, and keep speculative sibling prepass on a separate counter so it does not consume the final-instance quota. Return `ErrModuleNotEvaluated` when depth, cycle, or budget stops evaluation, and reset the budget counter before each root so one large stack cannot exhaust the quota for the rest of a monorepo.

**Scan engine**

Release the evaluation cache after each root module finishes so peak memory scales with the largest single root rather than the whole repository. When a module directory has any skipped call site, keep its body and call-site blocks in the scan so partially resolved directories are not suppressed with no synthetic replacement.

**Builder engine**

Guard null and unknown cty literals before calling `AsString()` during expression conversion so a `null` literal in HCL cannot panic the scan.

**Tests**

Regression tests cover input-based memoization, per-path attribution on cache reuse, eval-cache retention across roots, nested fan-out affordability, depth-truncated cache rejection, budget accounting, and a skipped-by-default OOM reproduction probe (`IAC_PROBE_OOM=1`).

## QA Instruction

1. `go test ./pkg/parser/terraform/tfeval/... ./pkg/engine/... ./pkg/builder/engine/... -count=1`
2. `golangci-lint run -c .golangci.yml --timeout 20m pkg/parser/terraform/tfeval/... pkg/engine/... pkg/builder/engine/...`
3. Optional: `IAC_PROBE_OOM=1 IAC_PROBE_DEPTH=12 IAC_PROBE_BRANCH=3 go test ./pkg/parser/terraform/tfeval -run TestProbe_OOMReproBeforeAfter -v -count=1` and confirm peak heap stays bounded and resource count matches `branch^(depth-1)`.

## Impact

Affects the IaC scanner path that evaluates local Terraform modules when local module evaluation is enabled. For typical repositories, resolution semantics and findings coverage are unchanged: the same modules are instantiated with the same per-path attribution. Pathological nested fan-out shapes that previously OOMed should complete with the same instantiated resource counts at much lower peak memory. Edge cases where depth, cycle, or budget stopped evaluation now keep module bodies in the scan instead of silently dropping coverage. The `module_resources_instantiated` metric can still be high on large fan-out repos; this change bounds memory, not that count.

### DIFF
--- a/pkg/parser/terraform/tfeval/oom_repro_probe_test.go
+++ b/pkg/parser/terraform/tfeval/oom_repro_probe_test.go
@@ -78,9 +78,9 @@ func TestProbe_OOMReproBeforeAfter(t *testing.T) {
 	}
 
 	stack := writeNestedFanOutForProbe(t, depth, branch)
-	theoretical := powInt(branch, depth)
+	theoretical := powInt(branch, depth-1)
 
-	t.Logf("fixture: depth=%d branch=%d files=%d theoretical_instances=branch^depth=%d",
+	t.Logf("fixture: depth=%d branch=%d files=%d theoretical_resources=branch^(depth-1)=%d",
 		depth, branch, depth+1, theoretical)
 
 	run := func(label string) (resources, cacheEntries int, peakMiB float64, elapsed time.Duration, oom bool) {
@@ -125,6 +125,10 @@ func TestProbe_OOMReproBeforeAfter(t *testing.T) {
 
 	if cache > depth+2 {
 		t.Errorf("WITH FIX: cache has %d entries, expected ~%d", cache, depth+1)
+	}
+	if theoretical <= defaultMaxInstantiated && res != theoretical {
+		t.Errorf("resolved %d resources, want %d below the instantiation budget",
+			res, theoretical)
 	}
 }
 
--- a/pkg/parser/terraform/tfeval/oom_repro_probe_test.go
+++ b/pkg/parser/terraform/tfeval/oom_repro_probe_test.go
@@ -1,0 +1,137 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache-2.0 License.
+ *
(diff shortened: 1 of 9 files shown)


pkg/parser/terraform/tfeval/oom_repro_probe_test.go
```diff
@@ -0,0 +1,141 @@
...
+	stack := filepath.Join(root, "stack")
+	if err := os.MkdirAll(stack, 0o755); err != nil {
+		t.Fatalf("mkdir: %v", err)
+	}
+	if err := os.WriteFile(filepath.Join(stack, "main.tf"),
+		[]byte("module \"m\" {\n  source = \"../lvl00\"\n  in = \"seed\"\n}\n"), 0o644); err != nil {
+		t.Fatalf("write: %v", err)
+	}
+	return stack
+}
```
MikaYuoadas, line 50: **💡 Suggestion:** This looks like an exact copy of `writeNestedFanOut()` from `pkg/parser/terraform/tfeval/eval_cache_key_test.go`. Couldn’t we just have it as a shared test utils somewhere?

MikaYuoadas (approved):
LGTM, just some small nitpick in the tests.

PR #325, opened 2026-08-27
Bound the Terraform module instantiation budget over the whole scan

whitemerch:
## Summary

Production OOM kills happened because the instantiation budget was handed back at every root boundary. Each concurrent scan on a pod gets its own evaluator, and with `MaxConcurrentAnalyze` at four by default, four scans each holding a fresh 200k-resource allowance could exceed the 40 GiB container limit.

This change keeps the budget for the whole scan: only the speculative sibling prepass counter resets per root, since that work is discarded with the root it belongs to. The cap moves from 200,000 to 150,000 resources per scan.

## Test plan

- [x] `TestInstantiationBudget_SpansRootsRatherThanResettingPerRoot`
- [x] `TestInstantiationBudget_BoundsTotalAcrossManyRoots` — 40 roots against a budget of 500; fails without the fix (1,280 instantiated)
- [x] `go test ./pkg/parser/terraform/tfeval ./pkg/engine -count=1`
- [x] `golangci-lint run -c .golangci.yml` clean on changed packages

## Blast radius

Terraform local module evaluation only, gated by `k9-iac-enable-local-module-eval`.

### DIFF
--- a/pkg/engine/module_resolve.go
+++ b/pkg/engine/module_resolve.go
@@ -21,6 +21,7 @@ import (
 	tfmodules "github.com/DataDog/datadog-iac-scanner/pkg/parser/terraform/modules"
 	"github.com/DataDog/datadog-iac-scanner/pkg/parser/terraform/tfeval"
 	"github.com/cespare/xxhash/v2"
+	"github.com/rs/zerolog"
 )
 
 // extraCallerInfo records a deduplicated module caller so its findings can be
@@ -42,6 +43,8 @@ type moduleResolutionResult struct {
 	unresolvedModuleDirs map[string]bool
 	extras               map[string][]extraCallerInfo
 	resourceCount        int
+	rootCount            int
+	budgetExceeded       bool
 	ok                   bool
 }
 
@@ -76,6 +79,7 @@ func (c *Inspector) instantiateLocalModules(
 			len(res.docs),
 			deduplicatedCallerCount(&res),
 		)
+	logModuleInstantiationSummary(&contextLogger, &res)
 	if !res.ok {
 		return nil, nil, nil
 	}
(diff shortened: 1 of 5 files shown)


whitemerch:
## Correction: which of these two faults caused the production kills

The description attributes today's OOM kills to fault 1. That attribution is wrong, and the evidence points squarely at fault 2. Recording it here because it changes what this PR is for.

**Fault 1 never fired in production.** Production runs at `warn`, and `warn` ships — 1.58M `tfeval` warnings today. Both `max module depth ... exceeded` and `instantiation budget exhausted` are `warn`, and both return zero results across the kill window and the two hours before it.

**It could not have, on Aug 26.** There was an earlier kill spike on Aug 26 at 11:45 UTC, nine pods in four minutes. Fault 1 was introduced at 09:51 CEST on **Aug 27**. The same failure predates the commit by a day, so a single pathological repository is not the explanation either — Aug 26 hit Thriveworks, Alliants and Airtasker; Aug 27 hit Depop and Benchling.

**What the two incidents share** is aggregate load. Each killed pod was running several scans at once, bounded by `MaxConcurrentAnalyze` (4 by default). Each scan builds its own evaluator, and the instantiation budget was handed back at every root boundary — so four concurrent scans each held an allowance that bounded nothing at all. Enabling the flag for more orgs did not introduce a bad repository; it filled those concurrent slots with module evaluation instead of flat parsing.

That is fault 2, and bounding it per scan is what closes this. At 150,000 resources and a measured 33 KiB each, one scan is worth ~4.8 GiB and four are ~19 GiB, inside a 40 GiB container with the flat-parse baseline on top.

**One thing this PR now also fixes:** the instantiation count was only logged at `info`, which production drops. A scan on its way to exhausting the container looked identical to an idle one, which is why this took two incidents to characterise. The count is now repeated at `warn` once it passes a quarter of the budget, so an approaching limit is visible before a process dies rather than inferred afterwards.

**Not done here, deliberately.** A process-wide budget shared across concurrent scans would bound the host more tightly, but it would make a scan's findings depend on what else was running beside it, and a security scan has to be reproducible. The invariant to hold instead is `MaxConcurrentAnalyze × cap` staying inside the container, now recorded on the constant. If scan concurrency is ever raised, that number needs revisiting — and gating module evaluation behind its own smaller semaphore would be the way to decouple the two.

NathanGallet-dd (approved):

PR #327, opened 2026-08-28
Add phase-attributed peak RSS telemetry for scan OOM investigation

whitemerch:
## Motivation

Large Terraform module instantiation scans can push the scanner subprocess past pod memory limits. We need production telemetry that ties peak resident memory to scan phases before rolling out local module evaluation more broadly.

## Changes

**memwatch**

Add a context-scoped helper that samples `getrusage` Maxrss when scan phases complete and logs one info summary with `peak_rss_bytes` and `peak_rss_phase`.

**Scan pipeline**

Record completed phase samples throughout the CLI scan lifecycle and write `PeakRSSBytes` and `PeakRSSPhase` into scan metadata so consumers do not need to parse scanner logs.

## QA Instruction

1. `go test -race ./internal/memwatch/... ./pkg/engine/... ./pkg/scan/... ./pkg/scanner/...`
2. `make build`
3. `./bin/datadog-iac-scanner scan -p test/fixtures/test_terraform -o /tmp/iac-out -t Terraform --metadata-path /tmp/iac-meta.json --log-level info`
4. Confirm the info log contains `memwatch: peak resident memory summary` and `/tmp/iac-meta.json` includes `PeakRSSBytes` and `PeakRSSPhase`.

## Blast Radius

CLI scan subprocess logging and metadata output only. The watcher is scoped to the scan context, and server scans without a watcher remain unaffected. No rule or finding behavior changes.

## Additional Notes

### DIFF
--- a/internal/memwatch/memwatch.go
+++ b/internal/memwatch/memwatch.go
@@ -1,0 +1,95 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache-2.0 License.
+ *
+ * This product includes software developed at Datadog (https://www.datadoghq.com)  Copyright 2024 Datadog, Inc.
+ */
+
+// Package memwatch records the process peak resident set size together with the
+// scan phase it was reached in.
+package memwatch
+
+import (
+	"context"
+	"sync"
+
+	"github.com/rs/zerolog"
+)
+
+const (
+	PhaseStartup        = "startup"
+	PhaseAnalyzePaths   = "analyze_paths"
+	PhaseModuleResolve  = "remote_module_resolve"
+	PhaseGetQueries     = "get_queries"
+	PhasePrepareSources = "prepare_sources"
+	PhaseModuleEval     = "local_module_eval"
+	PhaseStartScan      = "start_scan"
+	PhaseGenerateReport = "generate_report"
+)
+
+type watcherKey struct{}
+
+type Watcher struct {
+	mu        sync.Mutex
+	logger    zerolog.Logger
+	sampleRSS func() (uint64, bool)
+	peakBytes uint64
+	peakPhase string
+	stopped   bool
+}
+
+func Start(ctx context.Context, log *zerolog.Logger) (context.Context, *Watcher) {
+	w := &Watcher{
+		logger:    *log,
+		sampleRSS: peakRSSBytes,
+	}
+	w.sample(PhaseStartup)
+	return context.WithValue(ctx, watcherKey{}, w), w
+}
+
+func Sample(ctx context.Context, phase string) {
+	if w, ok := ctx.Value(watcherKey{}).(*Watcher); ok {
+		w.sample(phase)
+	}
+}
+
+func (w *Watcher) Peak() (bytes uint64, phase string) {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+	return w.peakBytes, w.peakPhase
+}
+
+func (w *Watcher) Stop() {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+
+	if w.stopped {
+		return
+	}
+	w.stopped = true
+
+	if w.peakBytes == 0 {
+		return
+	}
+
+	w.logger.Info().
+		Uint64("peak_rss_bytes", w.peakBytes).
+		Str("peak_rss_phase", w.peakPhase).
+		Msg("memwatch: peak resident memory summary")
+}
+
+func (w *Watcher) sample(phase string) {
+	current, ok := w.sampleRSS()
+	if !ok {
+		return
+	}
+
+	w.mu.Lock()
+	defer w.mu.Unlock()
+	if w.stopped {
+		return
+	}
+	if current > w.peakBytes {
+		w.peakBytes = current
+		w.peakPhase = phase
+	}
+}
--- a/internal/memwatch/memwatch_test.go
+++ b/internal/memwatch/memwatch_test.go
@@ -1,0 +1,111 @@
+package memwatch
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"strings"
+	"testing"
+
+	"github.com/rs/zerolog"
+)
+
+func requirePeakRSSSampling(t *testing.T) {
+	t.Helper()
+	if _, ok := peakRSSBytes(); !ok {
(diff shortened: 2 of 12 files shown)


pkg/engine/inspector.go
```diff
@@ -486,8 +487,10 @@ func (c *Inspector) Inspect(
 	if shouldInstantiateLocalModules(platforms, files) &&
 		c.flagEvaluator != nil &&
 		c.flagEvaluator.EvaluateWithOrg(featureflags.IacEnableLocalModuleEval) {
+		memwatch.SetPhase(memwatch.PhaseModuleEval)
```
MarshalX: concurrent phase mutation; logical corruption

internal/memwatch/memwatch.go
```diff
@@ -0,0 +1,109 @@
...
+)
+
+var (
+	mu           sync.Mutex
+	active       bool
+	logger       zerolog.Logger
+	currentPhase string
+	peakBytes    uint64
+	peakPhase    string
+)
```
MarshalX: package level vars must not used anymore since we have concurrent web server inside single process; it will not work properly while debugging the http server

internal/memwatch/memwatch.go
```diff
@@ -0,0 +1,109 @@
...
+
+// Start begins peak-RSS tracking for the current scan.
+func Start(log *zerolog.Logger) *Watcher {
+	mu.Lock()
+	defer mu.Unlock()
+
+	logger = *log
+	active = true
+	currentPhase = PhaseStartup
+	peakBytes = 0
```
MarshalX: maxrss is process-lifetime mark. it can not decrease. resetting to 0 is not good

internal/memwatch/memwatch.go
```diff
@@ -0,0 +1,109 @@
...
+		return
+	}
+
+	sampleLocked()
+	active = false
+	if peakBytes == 0 {
+		return
+	}
+
+	logger.Warn().
```
MarshalX: we had problem with many warns in the past. let's think twice do we really need a warn here. or info is enough

pkg/scan/metadata.go
```diff
@@ -17,6 +17,10 @@ type ScanMetadata struct {
 	Stats ScanStats
 	// RuleStats contains statistics about rules.
 	RuleStats RuleStats
```
MarshalX, line 19: since we added json serialization here for new fields, should we add it for the existed fields?

internal/memwatch/memwatch_test.go
```diff
@@ -0,0 +1,83 @@
...
+	})
+
+	nop := zerolog.Nop()
+	Start(&nop)
+	SetPhase(PhaseAnalyzePaths)
+	peakPhaseBeforeTransition := peakPhase
+
+	SetPhase(PhaseModuleResolve)
+	if peakPhase != peakPhaseBeforeTransition && peakPhase == PhaseModuleResolve {
+		t.Fatalf("peak attributed to new phase %q before a higher Maxrss was observed", peakPhase)
```
MarshalX: is it dead code? how we can reach this peakPhase state

internal/memwatch/memwatch_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+	nop := zerolog.Nop()
+	w := Start(&nop)
+	SetPhase(PhaseGenerateReport)
+	beforeStop, _ := Peak()
+	w.Stop()
+	afterStop, _ := Peak()
+
+	if afterStop < beforeStop {
+		t.Fatalf("Peak() before Stop = %d, after Stop = %d", beforeStop, afterStop)
```
MarshalX: do we test here that peakBytes are monotonic?

internal/memwatch/memwatch_test.go
```diff
@@ -0,0 +1,83 @@
...
+	out := logs.String()
+	if !strings.Contains(out, "memwatch: peak resident memory summary") {
+		t.Fatalf("expected summary log, got %q", out)
+	}
+
+	var entry map[string]any
+	if err := json.Unmarshal([]byte(strings.TrimSpace(out)), &entry); err != nil {
+		t.Fatalf("unmarshal summary log: %v", err)
+	}
+	if entry["peak_rss_phase"] == "" {
```
MarshalX: is missing key nil? not empty str

internal/memwatch/memwatch_test.go
whitemerch: yes, removed

pkg/scan/metadata.go
whitemerch: I didn't want to modify the existing structure as it's not the goal of the PR. Trying to keep it scoped to what it intends to do

internal/memwatch/memwatch.go
whitemerch: Not necessarily, it was just to investigate and make dashboards. Switched the summary log to Info

internal/memwatch/memwatch_test.go
whitemerch: TestSampleKeepsLargestPeakAndItsPhase feeds 100 → 300 → 200 and asserts we keep 300 and its phase

internal/memwatch/memwatch_test.go
whitemerch: !ok covers a missing key, and the empty-string check covers the other case. This test only runs when sampling actually works, so i expect a real phase in the log

pkg/scan/post_scan.go
```diff
@@ -137,11 +139,13 @@ func (c *Client) postScan(ctx context.Context, scanResults *Results) (ScanMetada
 
 	handler, exitCode := consoleHelpers.ResultsExitCode(&summary)
 	if handler.ShowError("results") && exitCode != 0 {
+		memwatch.Sample(ctx, memwatch.PhaseGenerateReport)
```
MarshalX: NIT: right before os.Exit. Nothing will read it because it skips Stop() deferrer?

whitemerch: Yeah I will drop it, that's what happens after these many fixes

MarshalX (approved):
There is still no HTTP Server mode support for these metrics, but thanks to redesign of global vars it should be possible to add in the future

PR #339, opened 2026-09-02
Fix plan-JSON parsing: dropped identifiers, missing data sources, and bracket-boundary bugs

ChouraquiBen:
Three independent bug fixes to the Terraform plan-JSON parsing/detection path, found while investigating a false positive in `datadog-iac-scanner-default-rules` PR 201.

## Changes

**1. Fix jsonfilter parser dropping unquoted multi-segment identifiers**
`VisitQualifiedidentifier` delegated to the default `VisitChildren`, which discards child return values, so unquoted values like `ec2.amazonaws.com` resolved to `nil` instead of the concatenated string. Now returns `ctx.GetText()` directly. This was the root cause of the PR 201 false positive: a Rego guard comparing against `ec2.amazonaws.com` was silently comparing against `nil`.

**2. Surface resolved plan-JSON data sources under `doc.data.<type>.<name>`**
`prior_state.values.root_module` holds resolved data blocks but was never read, so queries written against `.tf` data sources (e.g. `doc.data.aws_kms_key.<label>.key_id`) never matched plan JSON even when the data source resolved successfully. Now walks `prior_state` the same way `planned_values` is already walked, filters to `mode == "data"`, and merges into `data.<type>.<name>` to match the HCL parser's shape. Absent `prior_state` (offline/unresolved data sources) is a no-op.

**3. Fix bracket-boundary detection for quoted for_each keys in plan searchKeys**
`terraformPlanPath` used `strings.LastIndex` to find a bracket group's close, and `splitTopLevelDots` counted `[`/`]` depth naively — both assumed brackets in a for_each/count key can't contain a literal `]`, which is legal in a quoted HCL key (e.g. `module.secrets["prod]eu"]`). That produced a garbage gjson path, so `GetLineBySearchLine` failed and the line silently defaulted to 1. Replaced both with a quote-aware scan (`findMatchingBracket`, and the same awareness in `splitTopLevelDots`) that still respects nested, unquoted brackets like a count index `this[0]`.

## Author Checklist

## QA Instruction

Run `go test ./pkg/...`. Notably:
- `pkg/parser/jsonfilter/json_filter_test.go` — multi-segment identifier regression case.
- `pkg/parser/json/tfplan_test.go` — resolved data source merged under `data.<type>.<name>`.
- `pkg/detector/default_detect_test.go` — for_each key containing a literal `]`.

## Blast Radius

`datadog-iac-scanner` only: the Terraform plan-JSON parser and the plan searchKey-to-gjson-path detector. No changes to HCL parsing, other platforms, or the Datadog backend.

Note: fix #2 (as the number and not the PR linked automatically by GitHub) makes previously-dead `doc.data.<type>` Rego code paths in some default rules fire for the first time (e.g. `terraform-gcp-service-account-with-improper-privileges`, `terraform-aws-iam-policies-with-full-privileges`), which now correctly flags over-permissive data sources already present in those rules' test fixtures. The backend's expected-results for those fixtures predate this fix and don't yet include these findings — that needs a follow-up in `datadog-iac-scanner-default-rules`, tracked separately.

## Additional Notes

### DIFF
--- a/pkg/parser/json/tfplan.go
+++ b/pkg/parser/json/tfplan.go
@@ -342,11 +342,14 @@ func parseTFPlan(doc model.Document) (model.Document, error) {
 }
 
 // extractResourceHeaderLines returns each resource's "values" attribute
-// line, keyed by resource address.
+// line, keyed by resource address. Addresses are unique across the whole
+// plan, so planned_values and prior_state (resolved data sources) share one map.
 func extractResourceHeaderLines(rawPlan []byte) map[string]int {
 	lines := make(map[string]int)
-	root_module := gjson.GetBytes(rawPlan, "planned_values.root_module")
-	walkModule(&root_module, lines)
+	plannedRootModule := gjson.GetBytes(rawPlan, "planned_values.root_module")
+	walkModule(&plannedRootModule, lines)
+	priorStateRootModule := gjson.GetBytes(rawPlan, "prior_state.values.root_module")
+	walkModule(&priorStateRootModule, lines)
 	return lines
 }
 
@@ -383,7 +386,7 @@ func readPlan(
 	kp.readModule(plan.PlannedValues.RootModule, "", resourceLines, correlation)
 
 	if plan.PriorState != nil && plan.PriorState.Values != nil {
-		kp.readPriorStateData(plan.PriorState.Values.RootModule, "", deletedAddresses(plan.ResourceChanges))
+		kp.readPriorStateData(plan.PriorState.Values.RootModule, "", resourceLines, deletedAddresses(plan.ResourceChanges))
 	}
 
 	doc := model.Document{}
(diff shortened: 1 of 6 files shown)


ChouraquiBen:
As written in the note: some rules are failing because some rules paths are now triggered by the fixes. This is expected and goes to show the fixes work. The affected rules are the following:
* terraform-aws-iam-policies-with-full-privileges
* terraform-gcp-service-account-with-improper-privileges

pkg/parser/json/tfplan.go
```diff
@@ -379,6 +382,10 @@ func readPlan(
 
 	kp.readModule(plan.PlannedValues.RootModule, "", resourceLines, correlation)
 
+	if plan.PriorState != nil && plan.PriorState.Values != nil {
+		kp.readPriorStateData(plan.PriorState.Values.RootModule, "")
```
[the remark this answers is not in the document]
whitemerch: tru

pkg/parser/json/tfplan.go
```diff
@@ -443,6 +450,44 @@ func (kp *TFPlan) readModule(
...
+		// collide, mirroring readModule's resourceKey construction.
+		dataKey := resource.Name
+		if moduleAddress != "" {
+			dataKey = moduleAddress + "." + resource.Name
+		}
+		if resource.Index != nil {
+			dataKey = formatResourceKeyWithIndex(dataKey, resource.Index)
+		}
+
+		kp.Data[resource.Type][dataKey] = resource.AttributeValues
```
whitemerch: readModule injects `_dd_lines` for resources, but `readPriorStateData` doesn’t. should we do it here?

ChouraquiBen: You're right, that will only make detection easier later on 🙏

whitemerch (approved):

PR #341, opened 2026-09-02
Fix Terraform scanning and SARIF remediation edge cases

whitemerch:
## Motivation

Production `iac-scanning` logs showed recurring scanner-side failures that reduce coverage or remediation quality without failing whole scans: `jsonencode` suffix expressions were mis-parsed and logged as errors, null Terraform values could panic module evaluation or spam data-source resolution, unknown Terraform values could make converted documents fail JSON marshaling and skip entire files, literal file paths with bracket characters were treated as malformed globs and logged at error level, and SARIF replacement fixes failed when the vulnerable line pointed at a block header instead of the attribute inside the block.

## Changes

**Query payload preprocessing**

Parse the full HCL expression before expanding `jsonencode(...)` and only transform standalone calls, leaving expressions like `${jsonencode(x)}.exists(...)` unchanged without emitting error logs.

**Terraform evaluation**

Skip `ElementIterator` on null or unknown module output objects during local module evaluation, treat null tuple elements in data-source resolution as expected instead of logging `ExpToString` errors, and preserve unknown conditionals, function results, and template interpolations as source expressions so converted documents remain serializable.

**Path filtering**

When exclude or include path expressions contain glob metacharacters but are not valid glob syntax, fall back to treating the expression as a literal path without emitting scanner errors.

**SARIF remediation**

When replacement fix generation cannot parse the reported vulnerable line, search inside the resource or module block for the matching attribute before giving up.

## Author Checklist

1. [ ] I have reviewed my own PR.
4. [ ] I have tested my changes on staging, if applicable.
5. [ ] I have updated relevant documentation, if applicable.

## QA Instruction

1. `go test ./pkg/engine -run TestTransformJsonencodeInPayloadLeavesSuffixExpressionUnchanged -count=1`
2. `go test ./pkg/parser/terraform/tfeval -run TestEvaluateLocalModuleBlocksIgnoresNullPreliminaryOutputs -count=1`
3. `go test ./pkg/parser/terraform -run TestResolveTuplePreservesNullValues -count=1`
4. `go test ./pkg/parser/terraform/converter -run TestUnknownValuesRemainSerializable -count=1`
5. `go test ./pkg/engine/provider -run TestGetExcludePathsTreatsMalformedGlobAsLiteral -count=1`
6. `go test ./pkg/report/remediations -run TestTransformToSarifFix_ReplacementFindsAttributeInsideBlock -count=1`
7. `go test ./pkg/... -count=1`
8. `golangci-lint run -c .golangci.yml --timeout 20m ./pkg/...`

## Blast Radius

This affects the Datadog IaC Scanner only: Terraform parsing and evaluation, OPA payload preprocessing, scan path filtering, and SARIF fix generation. Findings are still emitted when expressions cannot be fully evaluated or automatic fixes cannot be attached; these changes reduce false-negative risk and improve remediation attachment for edge-case Terraform configs.

## Additional Notes

The changes are split into five focused commits. Custom-rule `deletion` remediation types remain intentionally unsupported because built-in rules use `removal`, `addition`, and `replacement`.

### DIFF
--- a/pkg/engine/provider/filesystem.go
+++ b/pkg/engine/provider/filesystem.go
@@ -115,12 +115,10 @@ func (s *FileSystemSourceProvider) addExcluded(ctx context.Context, excludePaths
 }
 
 // GetExcludePaths gets all the files that should be excluded
-func GetExcludePaths(ctx context.Context, pathExpressions string) ([]string, error) {
-	contextLogger := logger.FromContext(ctx)
+func GetExcludePaths(_ context.Context, pathExpressions string) ([]string, error) {
 	if strings.ContainsAny(pathExpressions, "*?[") {
 		info, err := filepathx.Glob(pathExpressions)
 		if err != nil {
-			contextLogger.Error().Msgf("failed to get exclude path %s: %s", pathExpressions, err)
 			return []string{pathExpressions}, nil
 		}
 		return info, nil
--- a/pkg/engine/provider/filesystem.go
+++ b/pkg/engine/provider/filesystem.go
@@ -59,7 +59,7 @@ func NewFileSystemSourceProvider(ctx context.Context, paths, excludes, onlyPaths
 	}
 
 	for _, exclude := range excludes {
-		excludePaths, err := GetExcludePaths(ctx, exclude)
+		excludePaths, err := GetExcludePaths(exclude)
 		if err != nil {
 			return nil, err
 		}
@@ -77,7 +77,7 @@ func NewFileSystemSourceProvider(ctx context.Context, paths, excludes, onlyPaths
 	if len(onlyPaths) > 0 {
 		fs.onlyPaths = make([]string, 0)
 		for _, only := range onlyPaths {
-			expanded, err := GetExcludePaths(ctx, only)
+			expanded, err := GetExcludePaths(only)
 			if err != nil {
 				return nil, err
 			}
@@ -115,7 +115,7 @@ func (s *FileSystemSourceProvider) addExcluded(ctx context.Context, excludePaths
 }
 
 // GetExcludePaths gets all the files that should be excluded
-func GetExcludePaths(_ context.Context, pathExpressions string) ([]string, error) {
+func GetExcludePaths(pathExpressions string) ([]string, error) {
 	if strings.ContainsAny(pathExpressions, "*?[") {
 		info, err := filepathx.Glob(pathExpressions)
 		if err != nil {
--- a/pkg/engine/provider/filesystem_test.go
+++ b/pkg/engine/provider/filesystem_test.go
@@ -6,6 +6,7 @@
 package provider
 
 import (
+	"bytes"
 	"context"
 	"io"
 	"os"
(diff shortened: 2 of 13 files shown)


pkg/engine/provider/filesystem.go
```diff
@@ -115,12 +115,10 @@ func (s *FileSystemSourceProvider) addExcluded(ctx context.Context, excludePaths
 }
 
 // GetExcludePaths gets all the files that should be excluded
-func GetExcludePaths(ctx context.Context, pathExpressions string) ([]string, error) {
-	contextLogger := logger.FromContext(ctx)
+func GetExcludePaths(_ context.Context, pathExpressions string) ([]string, error) {
```
MikaYuoadas: ❓ **Question:** What’s the point of keeping the context param if we’re not gonna use it?
We can just remove it and update the call sites (I found only 4 call sites, so that’s really not too bad).

```suggestion
func GetExcludePaths(pathExpressions string) ([]string, error) {
```

pkg/engine/provider/filesystem_test.go
```diff
@@ -652,6 +654,18 @@ func TestProvider_getExcludePaths(t *testing.T) {
...
+	ctx := zerolog.New(&logs).WithContext(context.Background())
+	path := filepath.Join("terraform", "module[production.tf")
+
+	got, err := GetExcludePaths(ctx, path)
+
+	require.NoError(t, err)
+	require.Equal(t, []string{path}, got)
+	require.NotContains(t, logs.String(), `"level":"error"`)
+}
+
```
MikaYuoadas: ❓ **Question:** Not sure why you’re adding an extra test as you didn’t change anything about this function’s code. You just removed a log line. 🤔

pkg/report/remediations/remediations_helper.go
```diff
@@ -164,6 +173,40 @@ func buildReplacementFix(ctx context.Context, vuln model.VulnerableFile,
...
+		return "", 0, false
+	}
+
+	start := max(vuln.BlockLocation.Start.Line, 1)
+	end := min(vuln.BlockLocation.End.Line, len(vuln.FileSource))
+	alternatives := parseAlternatives(before)
+	for lineNumber := start; lineNumber <= end; lineNumber++ {
+		line := vuln.FileSource[lineNumber-1]
+		trimmed := strings.TrimSpace(line)
+		if trimmed == "" || strings.HasPrefix(trimmed, "#") || strings.HasPrefix(trimmed, "//") {
```
MikaYuoadas, line 193: ❓ **Question:** What about multiline comments (e.g. "/*   */")? Do we want to support them here too?

MikaYuoadas (commented):
Overall LGTM, just left a few questions.

pkg/engine/provider/filesystem_test.go
whitemerch: I should have cleaned this up, it's because i initially changed it haha

pkg/report/remediations/remediations_helper.go
whitemerch: Im matching existing remediation comment handling, focused on the Codex case it flagged. It would be a nice thing but need a bit more code. I can follow up on this

pkg/engine/provider/filesystem.go
whitemerch: No reason to not to

MikaYuoadas (approved):

PR #344, opened 2026-09-04
Add retired CLI flag registry

whitemerch:
## Motivation

Hidden `--x-*` flags are internal experimental toggles passed by callers such as iac-scanning. When we remove a flag from the scanner, callers must currently land coordinated changes or the CLI exits with a usage error. This change introduces an explicit retired-flag registry so removed flags can stay registered as hidden no-ops until callers catch up, instead of guessing argv or stripping unknown `x-*` names.

## Changes

**CLI**

Retired flags live in `retired_flags.go`, isolated from active flag definitions. Each entry keeps its original urfave type, is hidden from help, and logs a warn when used. Shared retired flags apply to both `scan` and `serve`; command-specific ones stay scoped (for example scan-only).

As a first example, `--timeout` is moved out of `scan` into the registry. It no longer affects scanning; callers that still pass it get a warn and the command continues.

## QA Instruction

1. `go test ./cmd/scanner -run TestRetired -v -count=1`
2. Build and run: `./bin/datadog-iac-scanner scan -p test/fixtures/test_terraform --timeout 60 --log-level warn` — should succeed with a warn log for `timeout`.
3. Confirm a real typo still fails: `./bin/datadog-iac-scanner scan -p test/fixtures/test_terraform --pathh /tmp`
4. Confirm `--timeout` is not accepted on `serve`: `./bin/datadog-iac-scanner serve --timeout 60` — should fail with a usage error.

## Impact

CLI only. No change to scan results, rules, or the HTTP server.

## Author Checklist

### DIFF
--- a/cmd/scanner/retired_flags.go
+++ b/cmd/scanner/retired_flags.go
@@ -1,0 +1,46 @@
+package main
+
+import (
+	"context"
+
+	"github.com/rs/zerolog/log"
+	cli "github.com/urfave/cli/v3"
+)
+
+// Shared retired flags must be removed from both scan and serve in the same change.
+func sharedRetiredFlags() []cli.Flag {
+	return nil
+}
+
+func retiredScanOnlyFlags() []cli.Flag {
+	return []cli.Flag{
+		&cli.IntFlag{
+			Name:   "timeout",
+			Hidden: true,
+			Action: retiredFlagAction[int]("timeout"),
+		},
+	}
+}
+
+func retiredServeOnlyFlags() []cli.Flag {
+	return nil
+}
+
+func retiredScanFlags() []cli.Flag {
+	return joinRetiredFlags(retiredScanOnlyFlags())
+}
+
+func retiredServeFlags() []cli.Flag {
+	return joinRetiredFlags(retiredServeOnlyFlags())
+}
+
+func joinRetiredFlags(specific []cli.Flag) []cli.Flag {
+	return append(append([]cli.Flag{}, sharedRetiredFlags()...), specific...)
+}
+
+func retiredFlagAction[T any](name string) func(context.Context, *cli.Command, T) error {
+	return func(ctx context.Context, _ *cli.Command, _ T) error {
+		log.Ctx(ctx).Warn().Str("flag", name).Msg("ignoring retired flag")
+		return nil
+	}
+}
--- a/cmd/scanner/retired_flags.go
+++ b/cmd/scanner/retired_flags.go
@@ -7,6 +7,10 @@ import (
 	cli "github.com/urfave/cli/v3"
 )
 
+const retiredTimeoutUsage = "(DEPRECATED) no longer has any effect; " +
+	"queries run to completion and slow rules are logged instead. " +
+	"This flag will be removed."
+
 // Shared retired flags must be removed from both scan and serve in the same change.
 func sharedRetiredFlags() []cli.Flag {
 	return nil
@@ -14,14 +18,15 @@ func sharedRetiredFlags() []cli.Flag {
 
 func retiredScanOnlyFlags() []cli.Flag {
 	return []cli.Flag{
-		retiredInt("timeout"),
+		retiredInt("timeout", retiredTimeoutUsage),
 	}
 }
 
-func retiredInt(name string) *cli.IntFlag {
+func retiredInt(name, usage string) *cli.IntFlag {
 	return &cli.IntFlag{
 		Name:   name,
 		Hidden: true,
+		Usage:  usage,
 		Action: retiredFlagAction[int](name),
 	}
 }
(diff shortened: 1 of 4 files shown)


whitemerch:
@codex review

cmd/scanner/experimental_flags.go
```diff
@@ -0,0 +1,97 @@
...
+	flagName := strings.TrimPrefix(msg, unknownFlagErrPrefix)
+	flagName = strings.TrimPrefix(flagName, "-")
+	if !strings.HasPrefix(flagName, "x-") {
+		return "", false
+	}
+
+	return flagName, true
+}
+
+func removeFlagArg(args []string, flagName string) []string {
```
MarshalX: the current implementation does corrupt argv

Image

Image

cmd/scanner/experimental_flags.go
```diff
@@ -0,0 +1,97 @@
...
+	cli "github.com/urfave/cli/v3"
+)
+
+const unknownFlagErrPrefix = "flag provided but not defined: "
+
+// runCLI executes the root command, ignoring unknown hidden experimental flags
+// (names starting with "x-") so callers can drop deprecated toggles without
+// coordinated releases. Typos on real flags still fail fast.
+func runCLI(ctx context.Context, cmd *cli.Command, args []string) error {
+	for {
```
MarshalX: Unbounded retry loop. High change of hanging forever here

cmd/scanner/experimental_flags.go
```diff
@@ -0,0 +1,97 @@
...
+		if !ok {
+			return err
+		}
+
+		log.Warn().Str("flag", flagName).Msg("ignoring unknown experimental flag")
+		args = removeFlagArg(args, flagName)
+	}
+}
+
+func unknownExperimentalFlagName(err error) (string, bool) {
```
MarshalX: Should we flip the logic of entire detection here? we do know existed flags. it is allow list. everything beyond it must be rejected

cmd/scanner/experimental_flags.go
```diff
@@ -0,0 +1,97 @@
...
+		if err == nil {
+			return nil
+		}
+
+		flagName, ok := unknownExperimentalFlagName(err)
+		if !ok {
+			return err
+		}
+
+		log.Warn().Str("flag", flagName).Msg("ignoring unknown experimental flag")
```
MarshalX: `log.Warn` happens before applying log options. so `--log-format` and `--log-level` are ignored :(

MarshalX (commented):
To be honest the current approach "run, fail, guess, re-run" does not seem to be a way of solution. I would suggest to give a shot to allow list based approach

cmd/scanner/experimental_flags.go
whitemerch: shoot

cmd/scanner/experimental_flags.go
whitemerch: Indeed, I will move it

cmd/scanner/experimental_flags.go
whitemerch: I will replace the retry loop and usage-error handlers with a single argv filter driven by the command tree’s flags

MarshalX (commented):
The implementation still relies on team convention which everyone must remember in the head. For example the use of `--x-` prefix is allowed only for experimental **bools**, and not any experimental arguments. Because, for example, `... scan --x-removed-manifest /tmp/example.json -p /tmp/example` will not work and the path will not be truncated. One more convention will be to call CLI using `=` like this: `... scan --x-removed-manifest=/tmp/example.json -p /tmp/example`

Don't we had before `--x-remote-modules-cache-dir` and other `--x-remote...`? Which are not bool?

cmd/scanner/retired_flags.go
```diff
@@ -0,0 +1,46 @@
...
+	return nil
+}
+
+func retiredScanOnlyFlags() []cli.Flag {
+	return []cli.Flag{
+		&cli.IntFlag{
+			Name:   "timeout",
+			Hidden: true,
+			Action: retiredFlagAction[int]("timeout"),
+		},
```
MarshalX: NIT: create and call `retiredInt("timeout")`

MarshalX (commented):
Stale local checkout

~⚠️  this is not complete review, i stop right after facing this issue:~

<details><summary>old screenshot</summary>
<p>

Image

</p>
</details>

cmd/scanner/scan.go
```diff
@@ -66,12 +66,6 @@ var scanAction = &cli.Command{
 			Usage: "maximum depth that the scanner will resolve files in",
 			Value: 15,
 		},
-		&cli.IntFlag{
-			Name:   "timeout",
-			Usage:  "(DEPRECATED) no longer has any effect; queries run to completion and slow rules are logged instead. This flag will be removed.",
```
MarshalX, line 71: NIT: should we move back this usage string to retied_flags? to keep context for what is was used for

MarshalX (dismissed):
LGTM! sorry for double review of the same commit XD

MarshalX (approved):

PR #345, opened 2026-09-04
Fix line locations for Helm template wrappers

whitemerch:
## Motivation

Scanner findings are silently omitted from SARIF when line detection returns `-1`: the vulnerability builder coerces th line field to 1 but leaves the resource location at zero. Production telemetry also shows a large Kubernetes cohort where many rules fail together on the same Helm files. Those files are include-only wrappers whose rendered YAML is defined in a named template, so the detector cannot find rendered YAML keys in the attributed wrapper source.

## Changes

- Anchor unmatched, markerless Helm findings to the source line invoking `include`, `template`, or `tpl`.
- Set a complete line 1, column 1 resource location whenever the vulnerability builder applies its line 1 fallback.
- Add regression coverage for include-only Helm wrappers and complete fallback resource locations.

## Author Checklist

- [ ] All new and existing tests pass. Relevant packages pass; the full module run is blocked by existing path-name assumptions and a backend library API returning 502 in this sandbox.
- [ ] I have tested my changes on staging (not applicable).
- [ ] I have updated any relevant documentation (not applicable).

## QA Instruction

Run:

```bash
go test ./pkg/detector/... ./pkg/engine ./pkg/report/model ./pkg/runner ./pkg/resolver/helm -count=1
golangci-lint run -c .golangci.yml --timeout 20m
```

Verify that an include-only Helm template resolves a finding to the include invocation and that an otherwise undetected finding retains a valid line 1, column 1 SARIF resource location.

## Blast Radius

Datadog IaC Scanner finding source locations and SARIF emission across all platforms. The Helm-specific behavior only affects markerless templates that invoke named output templates.

## Additional Notes

Production evidence was checked against a representative include-only Helm wrapper that generated the same location error across many unrelated Kubernetes rules. No detection rule logic changes are included.

---

PR by Bits - [View session in Datadog](https://app.datadoghq.com/code/d1bbcf96-ee04-4bd6-a8e2-2901c9bff009)

Comment @datadog to request changes

### DIFF
--- a/pkg/detector/helm/helm_detect.go
+++ b/pkg/detector/helm/helm_detect.go
@@ -114,6 +114,26 @@ func (d DetectKindLine) DetectLine(ctx context.Context, file *model.FileMetadata
 		}
 	}
 
+	// Helm attributes output from a named template to the file that invoked it.
+	// Include-only wrapper files therefore contain no rendered YAML keys for the
+	// search path to match. Anchor those findings to the invocation that emitted
+	// the resource instead of returning an undetected line.
+	if file.HelmID == "" {
+		if line, col, ok := findHelmTemplateInvocation(lines); ok {
+			adjustedLine := line + 1
+			return model.VulnerabilityLines{
+				Line:                  adjustedLine,
+				VulnLines:             detector.GetAdjacentVulnLines(line, outputLines, lines),
+				LineWithVulnerability: lines[line],
+				ResolvedFile:          file.FilePath,
+				VulnerablilityLocation: model.ResourceLocation{
+					Start: model.ResourceLine{Line: adjustedLine, Col: col},
+					End:   model.ResourceLine{Line: adjustedLine, Col: len(lines[line])},
+				},
+			}
+		}
+	}
+
 	var filePathSplit = strings.Split(file.FilePath, "/")
 	contextLogger.Warn().Msgf("Failed to detect line associated with identified result in file %s", filePathSplit[len(filePathSplit)-1])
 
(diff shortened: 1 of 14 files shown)


pkg/detector/helm/helm_detect.go
```diff
@@ -124,6 +144,39 @@ func (d DetectKindLine) DetectLine(ctx context.Context, file *model.FileMetadata
...
+			} else {
+				actionEnd += actionStart
+				offset = actionEnd + len("}}")
+			}
+
+			action := strings.TrimLeft(sourceLine[actionStart:actionEnd], "- \t")
+			if strings.HasPrefix(action, "include ") ||
+				strings.HasPrefix(action, "template ") ||
+				strings.HasPrefix(action, "tpl ") {
+				if invocationIndex == documentIndex {
```
ChouraquiBen: Edge-case here: take

```
{{- if .Values.deployA }}
{{- include "resource.a" . }}
{{- end }}

{{- include "resource.b" . }}
```

If `deployA` is `false`, `resource.a` is not rendered. Therefore, despite its invocation index being 0, it's document index is not because it does not hold any

ChouraquiBen (commented):
One comment 🙏

pkg/detector/helm/helm_detect.go
whitemerch: indeed

pkg/resolver/helm/helm.go
```diff
@@ -427,6 +428,60 @@ func setID(chartReq *chart.Chart) *chart.Chart {
 	return chartReq
 }
 
+// addHelmInvocationMarkers instruments action-only wrapper templates so the
+// rendered output retains the source position of the invocation that actually
+// ran. The marker action is inserted immediately before include/template/tpl,
+// which keeps it under the same Helm control flow as the invocation.
+func addHelmInvocationMarkers(file *chart.File) *chart.File {
```
ChouraquiBen, line 435: Even more edge case: when an included resource is actually empty,
```
{{- include "empty.helper" . }}
{{- include "resource" . }}
```

Helm will not put a document separator in between the resolved parts, thus the markers will be written as:

```
# KICS_HELM_INVOCATION_1_0:
# KICS_HELM_INVOCATION_2_0:
Some K8s code
```

Instead of

```
# KICS_HELM_INVOCATION_1_0:
Some K8s code 1
---
# KICS_HELM_INVOCATION_2_0:
Some K8s code 2
```

If there is an issue with `"resource"`, line 1 will be returned.

TBH, this edge case is not something that we should actually meet except if customers really despise their code base and leave empty files to be rendered, so I am not considering this as blocking

ChouraquiBen (approved):
One nit, otherwise LGTM

PR #356, opened 2026-09-07
Fix production IaC scanning error signatures

whitemerch:
## Motivation

Production was emitting recurring errors across the service: null literal evaluation noise, query panics during line attribution, YAML line-info rebuild failures on multi-document CRDs, Terraform block lookup failures on header/comment lines, and SARIF replacement remediation failures on resource headers.

This PR hardens the scanner so those paths degrade safely instead of logging errors or panicking.

## Changes

**Terraform evaluation**

Introduce `ErrNullLiteral` and suppress error logging for expected nulls in nested tuple expressions such as `var.enabled ? "enabled" : null`.

**Line detection and remediation location**

Clamp Terraform remediation insertion when function wrappers point past the end of a file, which caused `variable_without_description` panics. When a matched line sits outside any HCL block (for example license comments above a resource), attach the nearest block or fall back to a line-only location instead of failing with `failed to locate block for line`.

**Detector helpers**

Guard `ExtractLineFragment` when the expected substring is missing from the physical line, which caused `communication_over_http` panics on multiline YAML matches. Restrict Docker placeholder parsing to numeric `{{N}}` tokens so native CI syntax like `${{ github.ref }}` no longer triggers `failed to extract curly brackets substring`.

**YAML line-info rebuild**

Skip empty trailing `---` documents and continue past individually unparseable documents so large CRD bundles such as Strimzi can be reparsed for line attribution without `failed to build line-info document`.

**SARIF replacement remediations**

When a replacement fix lands on a Terraform resource header and the target attribute is not on that line, search inside the owning block for the attribute and fall back to an addition remediation there; value-only patches without a key are rejected rather than emitting invalid HCL.

## Author Checklist

## QA Instruction

1. Run `go test ./pkg/detector/... ./pkg/parser/terraform/... ./pkg/parser/yaml/... ./pkg/builder/engine/... ./pkg/report/remediations/... -count=1`.
2. Run `golangci-lint run -c .golangci.yml --timeout 20m ./pkg/detector/... ./pkg/parser/terraform/... ./pkg/parser/yaml/... ./pkg/builder/engine/... ./pkg/report/remediations/...`.
3. Confirm Strimzi-style multi-document YAML with a trailing `---` parses successfully and produces documents for line-info rebuild.
4. Scan Terraform with leading comment lines and replacement remediations on resource headers; verify scans complete without the error signatures above and SARIF fixes are still emitted where possible.

## Blast Radius

This affects the Datadog IaC Scanner CLI and any service that runs it. Changes are limited to parsing, line attribution, and SARIF remediation generation; detection rules and query logic are unchanged.

## Additional Notes

Commit `e41ac593` on `main` already guards null Terraform module outputs during evaluation; this PR complements that work by reducing remaining null-related log noise and the other error signatures observed in the last 24 hours.

**Regression detector (repo B):** the +0.6% finding-count delta is expected coverage recovery, not a false-positive regression. On `main`, multi-document Kubernetes YAML/CRD bundles with a trailing `---` fail YAML parse entirely and produce zero scannable documents; this PR parses them successfully (10 CRDs in that file). dd-source contains many such files, which explains the small finding increase.

### DIFF
--- a/pkg/report/remediations/remediations_helper.go
+++ b/pkg/report/remediations/remediations_helper.go
@@ -47,8 +47,13 @@ func buildReplacementFix(ctx context.Context, vuln model.VulnerableFile,
 	keyValRegex := regexp.MustCompile(`(?m)["']?(\w+)["']?\s*[:=]\s*(\[.*?\]|".*?"|[^#]+)`)
 	matches := keyValRegex.FindStringSubmatch(vuln.LineWithVulnerability)
 	if isBlockHeader(vuln.LineWithVulnerability) || len(matches) < 3 {
-		line, lineNumber, found := findReplacementTarget(&vuln, before, keyValRegex)
+		line, lineNumber, found, keyPresent := findReplacementTarget(&vuln, before, keyValRegex)
 		if !found {
+			if keyPresent {
+				err := fmt.Errorf("attribute already present with a different value in %s", vuln.FileName)
+				contextLogger.Warn().Msg(err.Error())
+				return model.SarifFix{}, err
+			}
 			return fallbackReplacementToAddition(&vuln, before, after, startLocation)
 		}
 		vuln.LineWithVulnerability = line
@@ -175,11 +180,11 @@ func findReplacementTarget(
 	vuln *model.VulnerableFile,
 	before string,
 	keyValRegex *regexp.Regexp,
-) (line string, lineNumber int, found bool) {
+) (line string, lineNumber int, found, keyPresent bool) {
 	firstAlternative := strings.Split(before, " or ")[0]
 	expectedKey, _ := splitKeyValue(firstAlternative)
 	if expectedKey == "" {
-		return "", 0, false
+		return "", 0, false, false
 	}
 
 	start := max(vuln.BlockLocation.Start.Line, 1)
@@ -192,7 +197,7 @@ func scanReplacementRange(
 	start, end int,
 	expectedKey, before string,
 	keyValRegex *regexp.Regexp,
-) (line string, lineNumber int, found bool) {
+) (line string, lineNumber int, found, keyPresent bool) {
 	if start < 1 {
 		start = 1
 	}
@@ -210,14 +215,15 @@ func scanReplacementRange(
 		if len(matches) < 3 || normalize(matches[1]) != expectedKey {
 			continue
 		}
+		keyPresent = true
 		value := normalize(matches[2])
 		for _, alternative := range alternatives {
 			if alternative != "" && strings.Contains(value, alternative) {
-				return line, lineNumber, true
+				return line, lineNumber, true, true
 			}
 		}
 	}
-	return "", 0, false
+	return "", 0, false, keyPresent
 }
 
 func isBlockHeader(line string) bool {
@@ -239,16 +245,27 @@ func fallbackReplacementToAddition(
 	if addition == "" {
 		return model.SarifFix{}, fmt.Errorf("could not parse key-value from line: %s", vuln.LineWithVulnerability)
 	}
-	if !strings.Contains(addition, "=") {
+	if !isTopLevelAttribute(addition) {
 		key, _ := splitKeyValue(before)
 		if key == "" {
 			return model.SarifFix{}, fmt.Errorf("could not parse key-value from line: %s", vuln.LineWithVulnerability)
 		}
 		addition = key + " = " + addition
 	}
+	if !isTopLevelAttribute(addition) {
+		return model.SarifFix{}, fmt.Errorf("remediation is not a top-level attribute: %s", addition)
+	}
 	updated := *vuln
 	updated.Remediation = addition
-	return buildAdditionFix(updated, startLocation)
+	return buildAdditionFix(updated, additionInsertLocation(vuln, startLocation))
+}
+
+func additionInsertLocation(vuln *model.VulnerableFile, startLocation model.SarifResourceLocation) model.SarifResourceLocation {
+	end := vuln.BlockLocation.End.Line
+	if end < 1 || end > len(vuln.FileSource) {
+		return startLocation
+	}
+	return model.SarifResourceLocation{Line: end, Col: 1}
 }
 
 // nolint:gocritic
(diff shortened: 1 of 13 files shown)


pkg/report/remediations/remediations_helper.go
```diff
@@ -199,14 +212,45 @@ func findReplacementTarget(
...
+	if !strings.Contains(addition, "=") {
+		key, _ := splitKeyValue(before)
+		if key == "" {
+			return model.SarifFix{}, fmt.Errorf("could not parse key-value from line: %s", vuln.LineWithVulnerability)
+		}
+		addition = key + " = " + addition
+	}
+	updated := *vuln
+	updated.Remediation = addition
+	return buildAdditionFix(updated, startLocation)
```
ChouraquiBen: When the block is empty, the addition is added outside of the block, next to the header.

pkg/report/remediations/remediations_helper.go
```diff
@@ -46,12 +46,10 @@ func buildReplacementFix(ctx context.Context, vuln model.VulnerableFile,
...
 	keyValRegex := regexp.MustCompile(`(?m)["']?(\w+)["']?\s*[:=]\s*(\[.*?\]|".*?"|[^#]+)`)
 	matches := keyValRegex.FindStringSubmatch(vuln.LineWithVulnerability)
-	if len(matches) < 3 {
+	if isBlockHeader(vuln.LineWithVulnerability) || len(matches) < 3 {
 		line, lineNumber, found := findReplacementTarget(&vuln, before, keyValRegex)
 		if !found {
-			err := fmt.Errorf("could not parse key-value from line: %s", vuln.LineWithVulnerability)
-			contextLogger.Error().Msg(err.Error())
-			return model.SarifFix{}, err
+			return fallbackReplacementToAddition(&vuln, before, after, startLocation)
```
ChouraquiBen, line 57: `findReplacementTarget` returns `false` both when the key is absent and when it exists with a different value. For example, an existing `enabled = var.enabled` causes the fallback to add `enabled = false`, producing `Attribute redefined`

pkg/report/remediations/remediations_helper.go
```diff
@@ -199,14 +212,45 @@ func findReplacementTarget(
...
+func fallbackReplacementToAddition(
+	vuln *model.VulnerableFile,
+	before, after string,
+	startLocation model.SarifResourceLocation,
+) (model.SarifFix, error) {
+	addition := strings.TrimSpace(after)
+	if addition == "" {
+		return model.SarifFix{}, fmt.Errorf("could not parse key-value from line: %s", vuln.LineWithVulnerability)
+	}
+	if !strings.Contains(addition, "=") {
```
ChouraquiBen: An issue may occur when the remediation is not a full assignment. Consider a missing tag, the remediation for which would be:
```
{
  "before": "{}",
  "after": "{ Environment = \"prod\" }"
}
```

The code needs to verify that `after` is structurally a top-level HCL attribute

pkg/report/remediations/utilities.go
```diff
@@ -1,11 +1,14 @@
 package model
 
 import (
+	"regexp"
 	"strings"
 
 	"github.com/DataDog/datadog-iac-scanner/pkg/model"
 )
 
+var topLevelAttributeRegex = regexp.MustCompile(`(?s)^["']?[A-Za-z_][\w-]*["']?\s*[:=]\s*\S`)
```
ChouraquiBen: expressions with `:` and `==` such as `description: "new"` or `enabled == true` are still considered top level attributes.

I'm also wondering if we should have some enforcement in the default rules repo as to what a remediation should look like but at the same time, this really is no priority.

pkg/report/remediations/remediations_helper.go
```diff
@@ -197,21 +215,68 @@ func findReplacementTarget(
...
+	}
+	if !isTopLevelAttribute(addition) {
+		return model.SarifFix{}, fmt.Errorf("remediation is not a top-level attribute: %s", addition)
+	}
+	updated := *vuln
+	updated.Remediation = addition
+	return buildAdditionFix(updated, additionInsertLocation(vuln, startLocation))
+}
+
+func additionInsertLocation(vuln *model.VulnerableFile, startLocation model.SarifResourceLocation) model.SarifResourceLocation {
```
ChouraquiBen, line 282: Since `Col` is always 1, this may be an issue for attributes written such as:

```
resource name {
    # Remediation will go before the block
    flawed_attribute {}
}
```

ChouraquiBen (commented):
Two comments 🙏

pkg/report/remediations/utilities.go
whitemerch: it makes sense. reminder to myself to follow-up on this

pkg/report/remediations/remediations_helper.go
```diff
@@ -44,7 +44,7 @@ func buildReplacementFix(ctx context.Context, vuln model.VulnerableFile,
 	after := strings.TrimSpace(patch["after"])
 
 	// Regex for key and value (list/quoted/unquoted)
-	keyValRegex := regexp.MustCompile(`(?m)["']?(\w+)["']?\s*[:=]\s*(\[.*?\]|".*?"|[^#]+)`)
+	keyValRegex := regexp.MustCompile(`(?m)["']?(\w+)["']?\s*=\s*(\[.*?\]|".*?"|[^#=][^#]*)`)
```
ChouraquiBen: Changing `keyValRegex` from `[:=]` to only `=` breaks the shared SARIF replacement path for colon-delimited files. For:
```
enabled: true
```
with `before: "enabled = true"` and `after: "false"`, the previous code produced:
```
enabled: false
```
The latest code fails to recognize the existing YAML attribute and falls through to an HCL-style addition:
```
enabled = false
enabled: true
```

pkg/report/remediations/remediations_helper_test.go
```diff
@@ -86,6 +87,379 @@ func TestTransformToSarifFixE2E(t *testing.T) { //nolint
...
+		Remediation:           `{"before": "enabled = true", "after": "false"}`,
+		LineWithVulnerability: `resource "aws_s3_bucket" "b" {`,
+		Line:                  1,
+		FileSource: []string{
+			`resource "aws_s3_bucket" "b" {`,
+			`    # comment`,
+			`    logging {}`,
+			`}`,
+		},
+		BlockLocation: model.ResourceLocation{
```
ChouraquiBen, line 376: The test is not realistic because such `BlockLocation` is never returned by [vulnerabilityLinesFromBlock](https://github.com/DataDog/datadog-iac-scanner/blob/3e0b7ac187345dc4f3709632534b3ac3cc00ceca/pkg/detector/terraform/terraform_detect.go#L208). The latter sends the outer resource block.

This test makes it seem that the `remediation` will be at the correct place but it won't once executed.

pkg/report/remediations/remediations_helper.go
```diff
@@ -197,21 +215,89 @@ func findReplacementTarget(
...
+}
+
+func additionInsertLocation(vuln *model.VulnerableFile, startLocation model.SarifResourceLocation) model.SarifResourceLocation {
+	start := max(vuln.BlockLocation.Start.Line, 1)
+	end := min(vuln.BlockLocation.End.Line, len(vuln.FileSource))
+	if end < start {
+		return startLocation
+	}
+	for line := end; line >= start; line-- {
+		if col := strings.LastIndex(vuln.FileSource[line-1], "}"); col >= 0 {
```
ChouraquiBen: Using `LastIndex` here means that
```
resource "google_project_iam_binding" "project3" { // I add this just to break the code }
```
may break the code

ChouraquiBen (commented):
Three comments 🙏

ChouraquiBen (approved):

PR #357, opened 2026-09-08
Register common Terraform evaluation functions

whitemerch:
## Motivation

Terraform expression evaluation only knew a subset of built-in functions, so common calls such as `length`, `lookup`, `try`, and `replace` stayed unresolved interpolations and rules missed the resulting values.

## Changes

**Terraform evaluation**

Register `try`, `can`, type conversions, YAML encode/decode, and Terraform-compatible `length`, `coalesce`, `lookup`, `index`, and `replace`. Stdlib implementations that differ from Terraform (string `length`, empty-string `coalesce`, optional `lookup` default, find-by-value `index`, `/regex/` `replace`) use thin wrappers so known arguments evaluate instead of wrapping as `${...}`.

Unknown or failing calls still wrap, so impure or incomplete expressions stay scan-safe.

## Author Checklist

## QA Instruction

1. `go test ./pkg/parser/terraform/functions/ ./pkg/parser/terraform/converter/ ./pkg/parser/terraform/tfeval/ -count=1`
2. Confirm `length("hello")`, `try(var.missing, "fallback")`, and `lookup({a = "x"}, "b", "d")` evaluate in converter tests rather than remaining interpolations.

## Blast Radius

This PR only changes Terraform expression evaluation in the scanner. Findings can increase where attributes previously stayed as unresolved function calls and now resolve to concrete values.

## Additional Notes

Filesystem functions (`file`, `templatefile`) and impure functions (`timestamp`, `uuid`) are intentionally still unregistered.

### DIFF
--- a/.copyright-overrides.yml
+++ b/.copyright-overrides.yml
@@ -302,6 +302,9 @@ github.com/kr/text: Copyright 2012 Keith Rarick
 github.com/anchore/go-struct-converter: "" # At date, this repository have no copyright
 github.com/santhosh-tekuri/jsonschema/*: Copyright Santhosh Tekuri
 github.com/DataDog/jsonapi: Copyright 2022-Present Datadog, Inc
+github.com/zclconf/go-cty-yaml:
+  - Copyright 2011-2016 Canonical Ltd.
+  - Copyright 2019 Martin Atkins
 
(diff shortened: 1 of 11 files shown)


barbiergaspard (approved):
LGTM

PR #358, opened 2026-09-08
Redact DB URL credentials from parse error logs

whitemerch:
## Motivation

The prod service emits roughly **2,800 logs/day** tagged as sensitive data. An investigation traced every one of them back to this repository.

When the scanner registers a `.ini` file for the Ansible INI hosts parser, `aini.Parse()` fails on lines that are not valid inventory syntax, and its error quotes the offending line verbatim:

```
failed to parse file content: /sandbox/repository/test/etc/test.ini | error.message: bad key=value pair supplied:postgres:xxx
```

That error was logged with `.Err(err)` without sanitization, so any credential embedded in a scanned **customer** file was copied straight into our logs. The downstream `dd-source` relay already recognises these as expected parse errors and downgrades them ERROR → WARN, but it passes the message through unsanitised.

`model.RedactURLCredentials` already existed for exactly this purpose, but its regex only matched `ssh://`, `http://` and `https://` — so database connection strings, the schemes that actually appear in these files, were never covered.

JIRA Ticket: N/A — found via log investigation of `service:iac-scanning sensitive_data_category:credentials`.

## Changes

- `pkg/model/summary.go`: broadened `urlAuthRedactRegex` to cover the database and message-broker schemes that show up in scanned files — `postgres`, `postgresql`, `mysql`, `mongodb`, `mongodb+srv`, `redis`, `rediss`, `amqp`, `amqps` — alongside the existing `ssh`/`http`/`https`. The scheme list is now a documented named constant, and matching is case-insensitive (URL schemes are case-insensitive per RFC 3986). Behaviour is unchanged for URLs with no embedded userinfo.
- `pkg/runner/sink.go`: added `redactErrorForLog`, which renders an error for logging with URL credentials stripped, and used it at the parse-failure log sites (both the ERROR path and the Helm-template DEBUG path). The redacted message is logged under `zerolog.ErrorFieldName`, i.e. the same `error` JSON field `.Err()` used, so the log shape stays byte-compatible for the `dd-source` relay.
- `pkg/runner/resolver_sink.go`: applied the same redaction to the resolved-file parse-failure log and to both render-failure logs in `logResolverResolveError` — Helm render errors quote the offending template/values fragment, so they are the same class of leak.

The error text is still logged (with scheme, host, port and path intact) so parse failures remain triageable; only the `user[:password]@` userinfo is removed.

## Testing

- `pkg/model/summary_test.go`: extended `TestRemoveURLCredentials` from 5 to 20 cases — one per newly covered scheme, plus the exact production error string, case-insensitive schemes, multiple URLs in one message, and negative cases proving credential-free URLs and unrelated `user:pass@host` text are left untouched. Converted the loop to subtests so failures name the offending case.
- `pkg/runner/sink_test.go`: added `TestSink_ParseFailureRedactsCredentials`, an end-to-end test that drives `sink()` with the **real** Ansible INI parser over `[main]\ndogdata_url = postgres://user:sup3rs3cret@localhost:5432/dogdata`, reproducing the production `bad key=value pair supplied: …` error, and asserts the captured log line still reports the failure and the host but no longer contains the password. Also added `TestRedactErrorForLog` for the nil/redacted/untouched cases.
- `pkg/runner/resolver_sink_test.go`: added `TestLogResolverResolveError_RedactsCredentials`, covering both the ERROR and DEBUG render-failure branches.
- `go test ./pkg/runner/ ./pkg/model/ ./pkg/engine/ ./pkg/parser/terraform/... -count=1` — all pass. `pkg/engine/provider` fails identically on `main` in this environment (pre-existing, unrelated).
- `golangci-lint run -c .golangci.yml --timeout 20m` with the CI-pinned **v2.4.0** — 0 issues repo-wide.

## Author Checklist

## QA Instruction

Scan a repository containing an `.ini` file whose value is a database connection string with embedded credentials, e.g.:

```ini
[main]
dogdata_url = postgres:xxx
```

The scanner should still log the parse failure at ERROR level, but the message must read `postgres://localhost:5432/dogdata` — the `user:sup3rs3cret@` portion must be gone. Confirm the log record still carries the message under the `error` JSON field so the `dd-source` relay keeps classifying it as an expected parse error. Afterwards, `service:iac-scanning sensitive_data_category:credentials` should trend toward zero for parse-failure logs.

## Blast Radius

DataDog IaC Scanner only. Two things change observably:

1. Log content — `user[:password]@` is stripped from URLs in parse/render error messages. Log levels, field names, field order and message text are unchanged, so the `dd-source` relay in `iac-scanning` needs no change.
2. `model.RedactURLCredentials` is shared with Terraform module resolution (`Source` / `Reason` fields in module attribution and SARIF output). Broadening it means a module source using one of the newly listed schemes would also get redacted — strictly more redaction, never less, and Terraform module sources do not use database schemes in practice.

No behaviour change to detection rules, scan results, or exit codes.

## Additional Notes

---

PR by Bits - [View session in Datadog](https://app.datadoghq.com/code/ba53d301-5ce9-4fc7-9f6c-0c4be0efde70)

Comment @datadog to request changes

### DIFF
--- a/pkg/model/summary.go
+++ b/pkg/model/summary.go
@@ -122,17 +122,32 @@ type PathParameters struct {
 	PathExtractionMap map[string]ExtractedPathObject
 }
 
-// redactableURLSchemes lists the URL schemes whose embedded userinfo is
-// stripped by RedactURLCredentials. Besides the VCS/transport schemes used by
-// module sources (ssh, http, https), it covers the database and message-broker
-// schemes that commonly appear in scanned files (connection strings in .ini,
-// .env, values.yaml, …) and can therefore end up inside third-party parser
-// errors that we log.
-const redactableURLSchemes = `ssh|https?|postgres(?:ql)?|mysql|mongodb(?:\+srv)?|rediss?|amqps?`
-
 var (
-	queryRegex         = regexp.MustCompile(`\?([\w-]+(=[\w-]*)?(&[\w-]+(=[\w-]*)?)*)?`)
-	urlAuthRedactRegex = regexp.MustCompile(`(?i)((?:` + redactableURLSchemes + `)://)(?:\S+(?::\S*)?@)`)
+	queryRegex = regexp.MustCompile(`\?([\w-]+(=[\w-]*)?(&[\w-]+(=[\w-]*)?)*)?`)
+
+	// urlAuthRedactRegex captures the scheme of any URL carrying embedded
+	// userinfo, so RedactURLCredentials can drop the userinfo and keep the
+	// rest. Both halves are deliberately shaped after RFC 3986:
+	//
+	//   - The scheme is matched generically (`ALPHA *( ALPHA / DIGIT / "+" /
+	//     "-" / "." )`) rather than from a fixed list of schemes. A list has
+	//     to be exhaustive to be safe, and it never is: it misses
+	//     driver-qualified schemes (postgresql+psycopg2, mysql+pymysql) and
+	//     every database or broker not yet enumerated (clickhouse, sqlserver,
+	//     cassandra, kafka, …). `user:password@` in an authority is a
+	//     credential whatever the scheme, and there is no scheme for which we
+	//     would want to keep it, so matching all of them is both simpler and
+	//     strictly safer.
+	//
+	//   - Userinfo is `[^\s/?#]+`, i.e. confined to the authority. Anything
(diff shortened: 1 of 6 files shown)


MikaYuoadas (approved):
LGTM.
Comment are a bit too verbose and LLM-y.
Could be made more concise but it’s really mostly a nitpick, it’s fine to merge as is.

PR #359, opened 2026-09-08
Add OpenTofu (.tofu) file support

whitemerch:
> [!TIP]

## Motivation

OpenTofu projects use `.tofu` and `.tofu.json` next to the usual `.tf` files. The scanner ignored those extensions, so OpenTofu-only repos and mixed repos were scanned incompletely.

When a directory has both twins (e.g. `main.tf` and `main.tofu`), OpenTofu treats the `.tofu` file as the one that counts for merged config. We still want to scan each file on its own when it shows up in the inventory, but we need the right precedence when we merge a directory for variables, module eval, and enrichment.

## Changes

**Discovery and inventory**

The scanner now picks up `.tofu` and `.tofu.json` the same way it does Terraform configs. If both twins are present, both stay in the inventory, we do not drop the `.tf` sibling at listing time.

**Directory merge**

When combining config from a directory (vars, data sources, module graph, `list-modules`), `.tofu` wins over `.tf` and `.tofu.json` over `.tf.json`, but only among files in the current scan set. Scanning `main.tf` directly still uses that file's own context, not the shadowed `.tofu` twin.

**Shared path helpers**

Path checks and twin shadowing live in `pkg/tfpath`. Module listing and symlink confinement go through `dir` helpers and a cached `pathutil` layer so remote module resolution stays bounded.

## Author Checklist

## QA Instruction

1. `go test ./pkg/tfpath/... ./pkg/parser/terraform/... ./pkg/engine/... ./pkg/scan/... ./cmd/scanner/... -count=1`
2. `go test ./test/e2e/... -count=1`
3. Scan a directory with `main.tf` + `main.tofu` and confirm both files produce findings when scanned separately, while module/variable merge follows OpenTofu precedence.

## Blast Radius

CLI scans, server-side Terraform analysis, and `list-modules` for repos that use OpenTofu file extensions. No rule asset changes.

## Additional Notes


whitemerch:
@codex review

pkg/tfpath/tfpath.go
```diff
@@ -0,0 +1,247 @@
...
+	ExtTofu     = ".tofu"
+	ExtTFJSON   = ".tf.json"
+	ExtTofuJSON = ".tofu.json"
+	ExtTFVars   = ".tfvars"
+)
+
+// IsJSONConfig reports whether path uses Terraform/OpenTofu JSON configuration syntax.
+func IsJSONConfig(path string) bool {
+	lower := strings.ToLower(path)
+	return strings.HasSuffix(lower, ExtTFJSON) || strings.HasSuffix(lower, ExtTofuJSON)
```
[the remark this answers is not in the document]
whitemerch: Neither do .tf.json, as supporting these two would widen the scope of this PR. I will implement it together with the .tf.json after it

pkg/tfpath/tfpath.go
```diff
@@ -0,0 +1,247 @@
...
+// SelectKeeping is Select, except keep stays in the result when it matched pred.
+// If keep is a .tf twin of a selected .tofu file, that .tofu twin is dropped so a
+// scanned Terraform file is not evaluated with its OpenTofu replacement.
+func SelectKeeping(paths []string, pred func(string) bool, keep string) []string {
+	return SelectKeepingWithTofuPrecedence(paths, pred, keep, nil)
+}
+
+// SelectKeepingWithTofuPrecedence is SelectWithTofuPrecedence, except keep
+// stays in the result and its twin is dropped.
+func SelectKeepingWithTofuPrecedence(
```
ChouraquiBen, line 77: When `keep = "main.tf"`, the function ends up mismatching `main.tf` with `variable.tofu`. Since those do not belong to the same edition of the book, it would not create a valid result.

pkg/tfpath/tfpath.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+	if len(shadowed) == 0 {
+		return items, nil
+	}
+	return kept, shadowed
+}
+
+func shadowKey(path, jsonExt, hclExt string) (string, bool) {
+	dir, base := splitDirBase(path)
+	lowerBase := strings.ToLower(base)
```
ChouraquiBen, line 222: Lowering this strings unwillingly makes twin matching case insensitive. [OpenTofu is case sensitive](https://github.com/opentofu/opentofu/blob/ae4c17990964b1ca0819ffaec03366cf969be5ed/internal/configs/parser_config_dir.go#L271-L305) in its parsing, so it should also be the case here

ChouraquiBen (commented):
Two comments!

pkg/tfpath/tfpath.go
whitemerch: Yeah, fair question, but main.tf and variable.tofu aren’t twins here

We only treat files as twins when they share the same basename (main.tf ↔ main.tofu, foo.tf.json ↔ foo.tofu.json). So with keep = "main.tf", we keep main.tf and still load variable.tofu, same as OpenTofu would

ChouraquiBen (approved):
Thanks for taking the time to answer my questions. LGTM

PR #360, opened 2026-09-10
Register more Terraform evaluation functions

whitemerch:
## Motivation

Terraform evaluation still left common string, collection, CIDR, and hash calls unresolved, so rules saw interpolations such as `${startswith(...)}` instead of concrete values.

This continues the function registry work from [PR 357](https://github.com/DataDog/datadog-iac-scanner/pull/357) for functions that have no suitable go-cty stdlib implementation.

## Changes

**Terraform evaluation**

Register `startswith`, `endswith`, `strcontains`, `alltrue`, `anytrue`, `sum`, `one`, `matchkeys`, `transpose`, `base64decode`, `urlencode`, `md5`, `sha1`, `sha256`, `sha512`, `cidrhost`, `cidrnetmask`, `cidrsubnet`, and `cidrsubnets`. Implementations follow Terraform semantics, including empty-list `alltrue`/`anytrue`, find-by-value `matchkeys`, IPv4-only `cidrnetmask`, negative `cidrhost` host numbers, prefix-aligned `cidrsubnets`, form-style `urlencode` spaces, and UTF-8-only `base64decode`.

Unknown or failing calls still wrap as interpolations. Filesystem functions (`file`, `fileset`, `templatefile`) stay out of this change.

## Author Checklist

- [ ] I have reviewed my own PR.
- [x] I have added or updated relevant unit tests where necessary. If no tests are added, I've explained why.
- [x] All new and existing tests pass.
- [ ] I have tested my changes on staging (if applicable).
- [ ] I have updated any relevant documentation (if applicable).

## QA Instruction

1. `go test ./pkg/parser/terraform/functions/ ./pkg/parser/terraform/converter/ ./pkg/parser/terraform/tfeval/ ./pkg/parser/terraform/ -count=1`
2. Scan a Terraform file that uses `cidrsubnet` or `startswith` and confirm the attribute is a concrete value, not `${...}`.

## Blast Radius

This PR only changes Terraform expression evaluation in the scanner. Rules that inspect resolved string, collection, CIDR, or hash values may start matching configurations that previously stayed unresolved.

## Additional Notes

I submit this contribution under the Apache-2.0 license.

PR #362, opened 2026-09-10
Register filesystem and remaining Terraform evaluation functions

whitemerch:
## Motivation

Terraform evaluation still left filesystem, path, encoding, and template calls unresolved, so rules saw interpolations such as `${file("note.txt")}` instead of the file contents.

This continues the function registry work from [PR 360](https://github.com/DataDog/datadog-iac-scanner/pull/360). Filesystem functions read through the scan `vfs.FS` and stay confined to the `.tf` file's directory tree.

## Changes

**Terraform evaluation**

Register path helpers (`abspath`, `basename`, `dirname`, `pathexpand`), encoding helpers (`base64gzip`, `base64sha256`, `base64sha512`, `textencodebase64`, `textdecodebase64`), `timecmp`, `uuidv5`, and `templatestring` on the static function map.

`EvalFuncs` adds `file`, `fileexists`, `fileset`, `filebase64`, the file hash helpers, and `templatefile`, rooted at the configuration directory and the scan filesystem (including MemFS). Paths that escape that tree fail closed and wrap as interpolations. Nested `templatefile`/`templatestring` calls are depth-limited.

The parser now passes the real file directory and `vfs.FS` into conversion so `file()` can resolve in both CLI and server scans.

Intentionally still unregistered: `timestamp`, `plantimestamp`, `uuid`, `bcrypt`, `rsadecrypt`, `sensitive`/`nonsensitive`/`issensitive`/`ephemeralasnull`, deprecated `list`/`map`, and `convert`.

## Author Checklist

## QA Instruction

1. `go test ./pkg/parser/terraform/functions/ ./pkg/parser/terraform/converter/ ./pkg/parser/terraform/tfeval/ ./pkg/parser/terraform/ -count=1`
2. Scan a Terraform file that uses `file("sidecar.txt")` next to the `.tf` file and confirm the attribute is the file contents, not `${file(...)}`.
3. Confirm `file("../outside")` stays wrapped as `${file("../outside")}`.

## Blast Radius

This PR only changes Terraform expression evaluation in the scanner. Rules that inspect resolved file, path, encoding, or template values may start matching configurations that previously stayed unresolved.

## Regression detector

The finding-count increase on one regression target is expected. This change evaluates `file()` and `templatefile()`, so rules can inspect sidecar policy documents that previously stayed as `${file(...)}`. Sampled new hits fire only when the inlined JSON actually matches the rule (for example `Action` and `Resource` both `*`). Scoped policies and unresolved `${path.module}/...` paths still wrap and do not invent findings. Peak RSS and wall time are unchanged.

## Additional Notes

### DIFF
--- a/pkg/parser/terraform/converter/default.go
+++ b/pkg/parser/terraform/converter/default.go
@@ -15,10 +15,12 @@ import (
 	"github.com/DataDog/datadog-iac-scanner/pkg/logger"
 	"github.com/DataDog/datadog-iac-scanner/pkg/model"
 	"github.com/DataDog/datadog-iac-scanner/pkg/parser/terraform/functions"
+	"github.com/DataDog/datadog-iac-scanner/pkg/vfs"
 	"github.com/hashicorp/hcl/v2"
 	"github.com/hashicorp/hcl/v2/hclsyntax"
 	"github.com/zclconf/go-cty/cty"
 	ctyconvert "github.com/zclconf/go-cty/cty/convert"
+	"github.com/zclconf/go-cty/cty/function"
 	ctyjson "github.com/zclconf/go-cty/cty/json"
 )
 
@@ -28,10 +30,20 @@ type VariableMap map[string]cty.Value
 // This file is attributed to https://github.com/tmccombs/hcl2json.
 // convertBlock() is manipulated for combining the both blocks and labels for one given resource.
 
-// DefaultConverted an hcl File to a toJson serializable object
-// This assumes that the body is a hclsyntax.Body
-var DefaultConverted = func(ctx context.Context, file *hcl.File, inputVariables VariableMap) (model.Document, error) {
-	c := converter{bytes: file.Bytes, inputVars: inputVariables}
+// Options controls filesystem-backed function evaluation during conversion.
+type Options struct {
+	BaseDir string
+	FS      vfs.FS
+}
+
+// Convert an hcl File to a toJson serializable object.
+// This assumes that the body is a hclsyntax.Body.
+func Convert(ctx context.Context, file *hcl.File, inputVariables VariableMap, opts Options) (model.Document, error) {
+	c := converter{
+		bytes:     file.Bytes,
+		inputVars: inputVariables,
+		funcs:     functions.EvalFuncs(opts.BaseDir, opts.FS),
+	}
 	body, err := c.convertBody(ctx, file.Body.(*hclsyntax.Body), 0)
 
 	if err != nil {
@@ -45,9 +57,27 @@ var DefaultConverted = func(ctx context.Context, file *hcl.File, inputVariables 
 	return body, nil
 }
 
+// DefaultConverted an hcl File to a toJson serializable object
+// This assumes that the body is a hclsyntax.Body
+var DefaultConverted = func(ctx context.Context, file *hcl.File, inputVariables VariableMap) (model.Document, error) {
+	return Convert(ctx, file, inputVariables, Options{})
+}
+
 type converter struct {
 	bytes     []byte
 	inputVars VariableMap
+	funcs     map[string]function.Function
+}
+
+func (c *converter) evalContext() *hcl.EvalContext {
+	funcs := c.funcs
+	if funcs == nil {
+		funcs = functions.TerraformFuncs
+	}
+	return &hcl.EvalContext{
+		Variables: c.inputVars,
+		Functions: funcs,
+	}
 }
 
 const (
@@ -232,10 +262,7 @@ func (v *converterExprVisitor) VisitFunctionCall(e *hclsyntax.FunctionCallExpr) 
 	return v.c.evalFunction(e)
 }
 func (v *converterExprVisitor) VisitConditional(e *hclsyntax.ConditionalExpr) (interface{}, error) {
-	val, err := e.Value(&hcl.EvalContext{
-		Variables: v.c.inputVars,
-		Functions: functions.TerraformFuncs,
-	})
+	val, err := e.Value(v.c.evalContext())
 	if err != nil || !val.IsWhollyKnown() {
 		return v.c.wrapExpr(e)
 	}
@@ -486,10 +513,7 @@ func (c *converter) convertTemplateFor(expr *hclsyntax.ForExpr) (string, error) 
 }
 
 func (c *converter) tryEvalExpression(expr hclsyntax.Expression) (interface{}, error) {
-	val, _ := expr.Value(&hcl.EvalContext{
-		Variables: c.inputVars,
-		Functions: functions.TerraformFuncs,
-	})
+	val, _ := expr.Value(c.evalContext())
 	if val.IsWhollyKnown() && !checkDynamicKnownTypes(val) {
 		return ctyjson.SimpleJSONValue{Value: val}, nil
 	}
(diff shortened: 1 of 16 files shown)


pkg/parser/terraform/functions/filesystem.go
```diff
@@ -0,0 +1,332 @@
...
+	}
+	sort.Strings(out)
+	return out, nil
+}
+
+func matchGlob(pattern, name string) (bool, error) {
+	pattern = filepath.ToSlash(pattern)
+	name = filepath.ToSlash(name)
+	if !strings.Contains(pattern, "**") {
+		return path.Match(pattern, name)
```
barbiergaspard, line 267: path.Match does not support patterns you can find in terraform, such as fileset(path.module, "files/{hello,world}.txt") (example direclty taken from https://developer.hashicorp.com/terraform/language/functions/fileset)

pkg/parser/terraform/functions/filesystem.go
```diff
@@ -0,0 +1,332 @@
...
+	}
+	return re.MatchString(name), nil
+}
+
+func globToRegexp(pattern string) (*regexp.Regexp, error) {
+	var b strings.Builder
+	b.WriteByte('^')
+	for i := 0; i < len(pattern); i++ {
+		switch {
+		case i+1 < len(pattern) && pattern[i] == '*' && pattern[i+1] == '*':
```
barbiergaspard, line 334: this accepts allows for ** invalid terraform patterns such as example/path**/file.tf (which in terraform should be interpreted the same as example/path*/file.tf)

** should be an entire path component https://developer.hashicorp.com/terraform/language/functions/fileset#-1

barbiergaspard (commented):
Does this implementation manage the path.module, path.root... explained [here](https://developer.hashicorp.com/terraform/language/expressions/references#filesystem-and-workspace-info)?

whitemerch:
> Does this implementation manage the path.module, path.root... explained [here](https://developer.hashicorp.com/terraform/language/expressions/references#filesystem-and-workspace-info)?

@barbiergaspard Nice spot! I will add it

pkg/parser/terraform/converter/default.go
```diff
@@ -33,16 +33,24 @@ type VariableMap map[string]cty.Value
 // Options controls filesystem-backed function evaluation during conversion.
 type Options struct {
 	BaseDir string
+	RootDir string
```
barbiergaspard: we should add this RootDir to terraform.go:320, because we create a new converter.Options there in the Parse function

pkg/parser/terraform/functions/path.go
```diff
@@ -0,0 +1,104 @@
...
+	root := filepath.Clean(baseDir)
+	if root == "" {
+		return AbsPathFunc
+	}
+	return function.New(&function.Spec{
+		Params: []function.Parameter{{Name: "path", Type: cty.String}},
+		Type:   function.StaticReturnType(cty.String),
+		Impl: func(args []cty.Value, retType cty.Type) (cty.Value, error) {
+			p := args[0].AsString()
+			if !filepath.IsAbs(p) {
```
barbiergaspard, line 55: Can't we just use joinUnderRoot, which already has this guard + another one?

pkg/parser/terraform/functions/template.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+func nestTemplateFuncs(
+	baseDir, rootDir string,
+	fsys vfs.FS,
+	funcsCb func() map[string]function.Function,
+	depth int,
+) map[string]function.Function {
+	funcs := maps.Clone(funcsCb())
+	funcs["templatefile"] = makeTemplateFileFuncDepth(baseDir, rootDir, fsys, funcsCb, depth+1)
```
barbiergaspard, line 117: I'm not sure I understand the whole implementation, but according to [the templatefile function documentation](https://developer.hashicorp.com/terraform/language/functions/templatefile), recursive calls to `templatefile` are not allowed. Why are we taking a recursive approach? From what I see , this can accept configurations that should be rejected by terraform, and there might be some performance concerns (especially since the recursive calls are in both the type evaluation and the actual implementation.

pkg/parser/terraform/functions/encode.go
```diff
@@ -68,9 +72,85 @@ func hashHexFunc(newHash func() hash.Hash) function.Function {
...
+var Base64GzipFunc = function.New(&function.Spec{
+	Params: []function.Parameter{{Name: "str", Type: cty.String}},
+	Type:   function.StaticReturnType(cty.String),
+	Impl: func(args []cty.Value, retType cty.Type) (cty.Value, error) {
+		var buf bytes.Buffer
+		gz := gzip.NewWriter(&buf)
+		gz.ModTime = time.Unix(0, 0)
+		if _, err := gz.Write([]byte(args[0].AsString())); err != nil {
+			return cty.NilVal, err
+		}
```
barbiergaspard, line 108: This implementation is missing the flush the [terraform implementation](https://github.com/hashicorp/terraform/blob/v1.14.3/internal/lang/funcs/encoding.go#L168-L183) has

pkg/parser/terraform/functions/template.go
whitemerch: The deliberate divergence here is that the scanner is best-effort: we want to resolve as much as possible rather than hard-fail configs Terraform would reject, so we support recursion with a depth cap instead of a hard error. Same philosophy as the rest of the evaluator (e.g. we tolerate unresolved vars by wrapping expressions).
The perf concern was valid though so it's fixed

pkg/parser/terraform/functions/filesystem.go
```diff
@@ -0,0 +1,415 @@
...
+		return nil, err
+	}
+	sort.Strings(out)
+	return out, nil
+}
+
+func matchGlob(pattern, name string) (bool, error) {
+	pattern = filepath.ToSlash(pattern)
+	name = filepath.ToSlash(name)
+	for _, alt := range expandBraces(pattern) {
```
barbiergaspard, line 273: this computes the braces expansion once per file for the pattern. This might add up to a lot of computations, especially if there are multiple braces expansion in 1 file. Couldn't we compute this once for the pattern and then use it for all the files?

pkg/parser/terraform/functions/filesystem.go
```diff
@@ -0,0 +1,415 @@
...
+	suffix := pattern[end+1:]
+	var out []string
+	for _, alt := range splitBraceAlts(pattern[start+1 : end]) {
+		out = append(out, expandBraces(prefix+alt+suffix)...)
+	}
+	return out
+}
+
+func findBraceGroup(pattern string) (start, end int, ok bool) {
+	start = strings.IndexByte(pattern, '{')
```
barbiergaspard: This does not take context into account, for example, "is the '{' escaped?" or "is the '{' in brackets?"

barbiergaspard: Not sure how common "{" are in file names that we would have to escape them in an actual pattern though

barbiergaspard (commented):
Nice! I left 2 comments for things I missed during my last review

pkg/parser/terraform/functions/filesystem.go
whitemerch: let's keep it safe

barbiergaspard (dismissed):
LGTM

barbiergaspard (approved):

PR #363, opened 2026-09-11
Add Docker Compose scanning support

whitemerch:
## Motivation

The default-rules repo ships Docker Compose detections, but the scanner cannot classify or parse Compose files yet. This PR adds end-to-end platform support so Compose can be scanned when selected while preserving the publication gate used by backend rulesets and local rule checkouts.

## Changes

**Platform detection and parsing**

Compose YAML is recognized structurally (versionless `services` with `image`, `build`, or `extends`, plus canonical filenames). Files route through the shared YAML parser under the `dockercompose` platform type and appear in `platforms.Supported`.

**Parser platform scoping**

`parser.Build()` intersects each parser's supported types with the requested platform list when building `Parser.Platform`. Requesting Kubernetes no longer silently attaches Docker Compose to the same YAML parser instance.

**Publication consistency**

Filesystem rule sources now skip metadata with `published: false`, matching backend rule filtering. Authors can explicitly include those rules with `--include-unpublished`. Catalog visibility remains controlled by `is_published`, and hosted scan enablement remains behind the existing platform feature flag.

> [!NOTE]
> Regression Detector will flag more findings as we enable the dockercompose platform, and the detector reads from the datadog-iac-scanner-default-rules repo so it reads from the old rules in the repository. However, this is not prod's behavior, as prod reads from the backends' rules, which won't get published for now, as the rules can be seen [here](https://github.com/DataDog/datadog-iac-scanner-default-rules/pull/271) with `"published": false`, so this PR is safe even though the detector is misleading.

## QA Instruction

1. `go test ./pkg/analyzer/... ./pkg/parser/... ./pkg/platforms/... ./pkg/engine/source/... ./pkg/scan/... ./cmd/scanner/...`
2. `golangci-lint run -c .golangci.yml --timeout 20m`
3. Scan a Compose fixture with `--type DockerCompose` and confirm published rules run.
4. Scan local metadata with `published: false`; confirm it is skipped by default and included with `--include-unpublished`.
5. Repeat with `--type Terraform` and confirm the Compose file is not scanned.

## Impact

CLI scans, custom-rule evaluation, and analyze requests that include `DockerCompose` can process Compose files. Unpublished local and backend rules stay inactive unless explicitly requested.

## Additional Notes


whitemerch:
@codex review

whitemerch:
Addressed the P1 review comment: the `Build` function in `pkg/parser/parser.go` now intersects the parser's `SupportedTypes` with the requested type list before populating `Parser.Platform`. When a scan requests only `kubernetes` (or any other YAML platform), the YAML parser's `Platform` slice contains only that platform — not `dockercompose` or the other types the same parser supports. The wildcard case (`types[0] == ""`) is preserved. Added a regression test `TestBuild_PlatformIsIntersectedWithRequestedTypes` that pins this boundary.

pkg/analyzer/analyzer.go
```diff
@@ -937,7 +934,14 @@ func storeHelmResults(hc *sync.Map, dirs []string, result bool) bool {
 	return result
 }
 
-func checkYamlPlatform(ctx context.Context, content []byte, path string) string {
+func dockerComposeExplicitlyRequested(typesFlag []string) bool {
```
robertohuertasm-datadog, line 937: **blocking**

I am unsure this helper means what the name suggests.

It returns true whenever `dockercompose` appears anywhere in `typesFlag`. The only exceptions are "no types" and the wildcard `{""}`.

That matters because a default scan is not Compose-only. The CLI `--type` default (in `cmd/scanner/scan.go`, not in this diff) and a server request with no `platform` both pass `platforms.Supported`:

`Ansible, CICD, CloudFormation, DockerCompose, Dockerfile, Kubernetes, Terraform`

Compose is on that list, so `dockerComposeExplicitlyRequested` is true on every ordinary scan, including `datadog-iac-scanner scan -p .` with no `-t`.

When that flag is true, `dockerComposeFromYAMLNode` accepts any YAML whose root `services:` is a non-empty mapping. It does not require `image`, `build`, `extends`, or a compose filename.

I reproduced this with `ClassifyFile` on a file named `prod.yaml` whose only Compose-like content is:

```yaml
services:
  web:
    ports:
      - "8080:80"
    environment:
      APP_ENV: production
```

No `image`, `build`, or `extends`. Results (input type list → classified platform):

- default list (`platforms.Supported`) → `"dockercompose"` (this is the production path, and the surprise)
- wildcard `{""}` → `""` (not classified; this is what the new test checks)
- `{DockerCompose}` only → `"dockercompose"` (this I expect, for `-t DockerCompose`)
- `{Terraform}` only → `""` (Compose was not requested, so we skip it)

I also passed `platforms.Supported` a generic `application.yaml` (`services.billing.endpoint`, no compose filename). It came back `"dockercompose"` too. `Analyze` with `Types: platforms.Supported` then treated `prod.yaml` as a Compose file.

So the problem is not "someone passed `-t Terraform,DockerCompose` on purpose." It is that the default scan already is that kind of list, and the loose `services:` matcher turns on for the whole repo.

Would it be worth treating "explicit" as "the only selected type is DockerCompose"? Default multi-platform scans would then keep the stricter heuristic (`image` / `build` / `extends`, or a compose-style filename). `-t DockerCompose` could still pick up override files like `prod.yaml`. Happy to be told the loose matcher on default scans is intentional.

```suggestion
func dockerComposeExplicitlyRequested(typesFlag []string) bool {
	if len(typesFlag) != 1 {
		return false
	}
	return typesFlag[0] == dockercompose
}
```

pkg/analyzer/analyzer_test.go
```diff
@@ -457,51 +457,78 @@ func TestClassifyFile_SwaggerOpenAPI(t *testing.T) {
...
+      - "8080:80"
+    environment:
+      APP_ENV: production
+`)
+	path := filepath.Join(t.TempDir(), "prod.yaml")
+	require.NoError(t, os.WriteFile(path, content, 0o600))
+
+	require.Equal(t, dockercompose,
+		ClassifyFile(context.Background(), nil, path, content, []string{"DockerCompose"}))
+	require.Equal(t, "",
```
robertohuertasm-datadog, line 485: **non-blocking**

This test uses two type lists for the same `prod.yaml` (ports/environment only, no compose filename):

1. `{DockerCompose}` → expect `"dockercompose"` (Compose-only scan, loose matcher on)
2. `{""}` → expect `""` (treats the wildcard as "default / not Compose-only", loose matcher off)

`{""}` is a real internal "all platforms" sentinel, but it is not what the CLI or the server send on a default scan. Those send `platforms.Supported`, which now includes `DockerCompose`.

I ran the same `prod.yaml` through `ClassifyFile` with `platforms.Supported` and got `"dockercompose"`, so this assertion does not cover the path customers actually hit.

Would it be worth adding a third case with `platforms.Supported` (expect `""` if the loose matcher should stay Compose-only)? I could be missing why `{""}` was used as the stand-in for default.

pkg/analyzer/yaml_classify.go
```diff
@@ -12,6 +12,7 @@ import (
 // when no regex-based type matched during classification.
 var yamlPlatformRootKeys = []string{
 	listKeywordsGoogleDeployment[0], // resources
+	"services",                      // Docker Compose
```
robertohuertasm-datadog: **non-blocking**

`yamlPlatformRootKeys` is not the Compose detector. It is the shared cheap gate in `checkYamlPlatform`:

```go
if !ansibleVarsPath && !yamlRootHasAnyKey(content, yamlPlatformRootKeys...) {
    return ""
}
```

If that call is false, we never parse the file and we never reach `dockerComposeFromYAMLNode`. Putting `"services"` on this list is how Compose gets through the gate today.

The same list is also the gate for GDM (`resources`) and Ansible (`playbooks`, `all`, `ungrouped`). `services` is a very common root key (Spring `application.yaml`, Helm values, generic app config), so the gate becomes true for a lot of YAML that is not IaC. We then fully unmarshal it, walk GDM/Ansible/Compose, and usually return `""`.

Compose still needs a way past that `return ""`. Would it be worth taking `"services"` off `yamlPlatformRootKeys` and adding a Compose-local condition next to `ansibleVarsPath`, instead of reusing the shared list?

Something like:

```go
maybeCompose := isDockerComposeFileName(path) || yamlRootHasAnyKey(content, "services")
if !ansibleVarsPath && !maybeCompose && !yamlRootHasAnyKey(content, yamlPlatformRootKeys...) {
    return ""
}
```

Same Compose files still enter phase 2. Unrelated `services:` YAML would not ride in because the GDM/Ansible keys happened to share a list. Totally understand if keeping one list was the point, I just wanted to flag that this key is doing gatekeeping for every YAML platform, not only Compose.

pkg/analyzer/analyzer.go
whitemerch: I missed that default scans already include Compose, so the loose matcher was on everywhere

pkg/analyzer/analyzer_test.go
whitemerch: I missed that {""} isn’t the real default path, added a platforms.Supported case.

pkg/analyzer/yaml_classify.go
whitemerch: I missed that this gate is shared across platforms, moved Compose’s services check to a Compose-local gate instead

robertohuertasm-datadog (approved):
LGTM! 🚀

PR #368, opened 2026-09-14
Enable Terraform local module evaluation in server mode

whitemerch:
## What

Enables local Terraform module resolution for the IDE/server (`serve`) scan path, closing the CLI-vs-IDE gap for local modules, and serializes module attribution on the analyze response so the IDE can anchor instantiated findings on the call site (the convention the CLI SARIF report already uses).

### How (scanner side)

- **vfs threading in tfeval**: `parseDir` / `LoadRootVars` / `CalledModuleDirs` read through the injected `vfs.FS` instead of `os.*`. In server mode that is the request's `MemFS` (pushed content only); in the CLI it is `DiskFS` (a passthrough. behavior-preserving).
- **Path-shape fix**: `EvaluateModule` resolves module dirs via `fsys.Abs` instead of `filepath.Abs`, which would relativize pushed paths against the server's working directory and break key matching. `repoRelativeFile` and `normalizedModuleSource` keep the pushed shape for content-push scans (`repoPath == ""`).
- **Flag flip**: `IacEnableLocalModuleEval` pinned **on** in `serverFlagEvaluator`. With every read confined to pushed content, the arbitrary-directory-read that motivated the pin is structurally gone: an unpushed module directory is a missing file reported for escalation, never a disk read. The guarantee also rests on the remote-module pre-scan staying off (analyze params never set `TerraformModules`).
- **Wire addition**: `ModuleAttribution` (previously `json:"-"`) serializes on the analyze response as `moduleAttribution` with `call_site` (the caller's module declaration: the anchor) and `code_location` (defining location, module-relative), the same locations the SARIF report anchors on. Verified this reaches only the server's analyze response: CLI reporters go through `VulnerableFile`, whose attribution stays private and is emitted as the SARIF `properties.module` payload.

### Escalation contract (unchanged on the wire, newly load-bearing)

An unpushed module directory is reported in `missing_files` as a **directory path**; the IDE stats it, expands it to its direct supported files, and the next round evaluates the module and discovers the level beneath it. `LoadRootVars` Stat-probes tfvars candidates so the always-probed `terraform.tfvars` does not surface as a false missing file.

## Verification

- Unit: `pkg/server`, `pkg/engine`, `pkg/model`, `pkg/parser/terraform/tfeval`, `pkg/scan`, `pkg/vfs`, `pkg/runner` — all green.
- E2E suite (`./test/e2e`) green; CLI smoke scan of `test/modules/simple_input` produces findings identical to the pre-change baseline (fingerprints and module attributions byte-identical).
- Live round-trip against a built binary: round 0 reports `missing_files: ["modules/networking"]` (directory shape), round 1 (module files pushed) resolves and returns the instantiated finding with `call_site` = the caller's module block. Same fixture scanned via the CLI confirms SARIF primary-location parity.
- New server tests: instantiation + call-site attribution, root-variable binding, tfvars override, absent-tfvars-not-reported (Stat-probe), absolute pushed-path shape preservation.

## Companion PR

https://github.com/DataDog/datadog-vscode/pull/2514: the extension consumes `moduleAttribution.call_site` to anchor module findings on the call site in both the interactive editor and the batch/folder scan. No protocol versioning needed: an older extension with a newer scanner (or vice versa) degrades to the previous behavior (defining-file anchoring / no attribution).

## Notes for reviewers

- CLI/DiskFS passthrough: `New()` = `NewWithFS(vfs.DiskFS{})`, `DiskFS.Abs == filepath.Abs`; the new `repoPath == ""` branches are server-only.
- The `summary_test` flip pins the new serialization contract (Vulnerability serializes for content-push; the report-side `VulnerableFile` stays clean).


whitemerch:
@codex review

pkg/server/analyze.go
```diff
@@ -275,23 +273,23 @@ func validateFilePath(p string) error {
...
+// analyze params never set TerraformModules): moduleprepare and modulegraph
+// still read the real disk.
 //
 // Parallel file parsing fans the per-file parse across CPUs; enabled by default
 // and can be disabled with --x-parallelparsing=false.
 func serverFlagEvaluator(parallelParsing bool) featureflags.FlagEvaluator {
 	return featureflags.NewLocalEvaluatorWithOverrides(map[string]bool{
 		featureflags.IacEnableKicsHelmResolver:        false,
-		featureflags.IacEnableLocalModuleEval:         false,
+		featureflags.IacEnableLocalModuleEval:         true,
```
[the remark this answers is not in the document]
whitemerch: Not a bug, the missing-module signal doesn't come from the evaluation phase, so the declaresTargetedResource shortcut can't suppress it

pkg/engine/module_resolve.go
```diff
@@ -360,7 +362,7 @@ func evaluateRootModules(
 	contextLogger := logger.FromContext(ctx)
 	for _, dir := range roots {
 		evaluator.ResetSpeculativeBudget()
-		resources, _, childDirs, err := evaluator.EvaluateModule(ctx, dir, tfeval.LoadRootVars(dir))
+		resources, _, childDirs, err := evaluator.EvaluateModule(ctx, dir, tfeval.LoadRootVars(dir, fsys))
```
robertohuertasm-datadog: **non-blocking**

`NewWithFS(fsys)` at `pkg/engine/module_resolve.go:477` already stores this filesystem on `evaluator`. After that, reads can go through the evaluator (that is what `Evaluator.CalledModuleDirs` does: it takes only `dir` and uses `e.fsys`).

Here we still pass `fsys` into `evaluateRootModules` as a separate argument, and the only use is this call: `LoadRootVars(dir, fsys)`. Same handle, twice, because `LoadRootVars` is a package function instead of a method on `Evaluator`.

That extra parameter shows up as a trailing `nil` on every `resolveModuleDocuments` test call (~15 sites). `parseDir` and `LoadRootVars` also nil-fall-back to `vfs.Default()`, which `NewWithFS` already does.

Would it be simpler to make `LoadRootVars` a method (`evaluator.LoadRootVars(dir)`) so `evaluateRootModules` can drop `fsys` and the injection point stays `NewWithFS` only? Totally understand if plumbing it through was the smallest way to get MemFS in.

pkg/model/module_attribution.go
```diff
@@ -23,15 +23,17 @@ type ModulePathHop struct {
...
+// (SARIF primary location, IDE code bit); ModuleCodeLocation is the defining
+// location inside the module, relative to the module root.
 type ModuleAttribution struct {
 	Name               string          `json:"name,omitempty"`
 	Source             string          `json:"source,omitempty"`
 	SourceType         string          `json:"source_type,omitempty"`
 	Version            string          `json:"version,omitempty"`
 	DependencyType     string          `json:"dependency_type,omitempty"`
-	CodeLocation       SourceLocation  `json:"-"`
+	CodeLocation       SourceLocation  `json:"call_site,omitempty"`
```
robertohuertasm-datadog: **nit**

`ModulePathHop.CodeLocation` still serializes as `code_location`, while this field is `call_site`. [datadog-vscode#2514](https://github.com/DataDog/datadog-vscode/pull/2514) already pins the JSON key (`IacModuleAttribution.call_site`), so I would not change the tag.

Take it or leave it: would renaming the Go field to `CallSite` (keeping `json:"call_site,omitempty"`) make the wire name and the tests that already say `call_site` line up?

```suggestion
	CallSite           SourceLocation  `json:"call_site,omitempty"`
```

robertohuertasm-datadog (dismissed):
LGTM! 🚀

pkg/model/module_attribution.go
whitemerch: makes sense

pkg/engine/module_resolve.go
whitemerch: yes!!

robertohuertasm-datadog (approved):

PR #371, opened 2026-09-15
fix(scan): some more optimizations

whitemerch:
## Motivation

Large IaC corpora used to keep several copies of every file's bytes and parsed tree live through eval, so peak RSS tracked corpus size rather than the working set. This change shares identical content and subtrees, converts OPA payloads in one pass, and drops data as soon as a phase no longer needs it.

## Changes

**Document lifetime**

Line info is computed lazily and original file bytes are released after each scan phase, so files that never produce a finding do not keep a split-line copy through eval.

**YAML parsing**

YAML documents are assigned directly instead of round-tripping through JSON. Repeated scalar strings are interned without a length or entry cap.

**Sharing across files**

Files with identical content reuse one parsed tree (except Terraform, reference-resolved, and Ansible parses, which are path-dependent). Structurally identical sanitized subtrees are hash-consed. Sharing is not gated on file size.

**OPA payload build**

All documents convert in one shared-memo pass and source trees are released as they convert. Nesting is not depth-capped; cyclic documents are still rejected earlier in sanitization.

**Eval GC**

Periodic GC during eval is gated on measured live heap rather than file count.

**Resolution shortcuts**

YAML reference resolution is skipped when no reference literal is present. OpenAPI regexes run only after a cheap literal pre-filter.

## QA Instruction

1. `go test ./pkg/engine ./pkg/runner ./pkg/model -count=1`
2. `golangci-lint run -c .golangci.yml --timeout 20m ./pkg/engine/... ./pkg/runner/... ./pkg/model/...`
3. Scan a large YAML corpus (for example `community-operators`: 74,271 files, 226 rules) and confirm the finding count matches `main`.

Local A/B on that corpus (same 31,570 findings on both sides):

| | `main` | this branch |
|---|---|---|
| Peak RSS | 22.3 GiB | 5.9 GiB |
| Wall | 15m13s | 2m10s |

## Impact

CLI and hosted scans of large multi-file repositories, especially YAML-heavy trees with duplicated manifests. Finding semantics are unchanged; peak RSS and wall time drop sharply on those corpora.


pkg/model/model.go
```diff
@@ -241,6 +305,21 @@ func (f *FileMetadata) EnsureLineInfoDocument(ctx context.Context) error {
...
+	}
+	// For non-Terraform kinds, OriginalData is only needed by the line-info
+	// loader (now nil'd) and by Lines() (which uses the now-populated
+	// LinesOriginalData). Detectors for these kinds don't read OriginalData
+	// directly, so release it to progressively reclaim ~the raw input content
+	// as each file's line info is resolved during eval.
+	if f.releaseOriginalDataAfterLineInfo &&
+		f.Kind != KindTerraform && f.Kind != KindTerraformPlan {
+		f.OriginalData = ""
+	}
```
MikaYuoadas: ❓ **Question:** Shouldn’t this also be guarded by `f.linesLazy.once` too to avoid race conditions with `Lines()` above?

pkg/model/model_yaml_test.go
MikaYuoadas: 🤏 **Nitpick:** Comment is now out of date as you drop this JSON round-trip.

MikaYuoadas (dismissed):
LGTM, just posted a nit about comment and one question about a possible race.

pkg/model/model.go
whitemerch: yes. it shouldnt happen in the current state of things, but we would be better gating it. Thank you for noticing this

MikaYuoadas (approved):

PR #374, opened 2026-09-18
Collect per-module resolution stats for hosted-scan telemetry

whitemerch:
## What

Collect one terminal resolution stat per unique remote module identity, for hosted-scan telemetry (K9CODESEC-5303).

`modulegraph.Resolve` now returns `Result.Stats`, one `ModuleStat` per unique module identity (source+version), never per declaration or per frontier:

- `Source`: resolver origin (`registry`, `git`, `git_http`, `git_local`, `local`, `dot_terraform`, `prefetched`), never the raw module source
- `Outcome`: `resolved` / `unresolved` / `unfetched` (budget-shed)
- `Cache`: on-disk module cache `hit`/`miss`/unknown
- `DurationMS`: the single resolution attempt
- `FailureCode: bounded failure taxonomy from the new `resolver.ClassifyFailure`: `timeout`, `budget_exceeded`, `allowlist_denied`, `registry_not_found`, `http_4xx`, `http_5xx`, `network`, `unpack_failed`, `unknown`. Raw error strings never become tags.

`moduleprepare.Result` carries the stats through so sandboxed/hosted consumers can aggregate them.

## Design notes

- **One sample per identity**: dedup is by resolution identity + resolution cache. A module resolved once and declared by several callers, or reached again in a deeper frontier, is counted exactly once. Resolution results are cached unconditionally (including under an expired context, where the entry carries its error) so a context-expiry retry cannot emit a duplicate sample.
- **Origin precedence**: a resolver-provided `Origin` wins over the detected source type; git sources resolved over HTTP are labeled `git_http` to distinguish them from BareGit's `git`.
- **Classification order**: typed errors first (`BudgetExceededError`, context deadline/cancel, `io.EOF`/`io.ErrUnexpectedEOF`), then stable substrings, then an `HTTP NNN` regexp for the 4xx/5xx ranges. Anything unrecognized maps to `unknown` rather than leaking error text. Substrings are deliberately specific (`allowlist`, `HTTP 404`, `no versions published`) rather than generic (`destination`, bare `not found`): `ChainResolver` joins errors from every attempted resolver, so a generic fragment from a benign local-cache miss would hijack the classification of the real failure that follows it in the joined message.
- Origins and failure codes are separate enums; `OriginUnknown` is deliberately not aliased to `FailureUnknown`.

## Telemetry consumption guidance

Stats are deliberately **not** metrics on their own, per-module source/outcome/cache/org cardinality would be a custom-metric bomb. Consumers should aggregate to scan-level counts by failure code (bounded, ~9 values) and leave per-module detail in structured logs. That is how dd-source consumes it (see the dd-source PR in this stack).

## Testing

- `stats_test.go`: one stat per identity (duplicate declaration), resolved/unresolved/unfetched outcomes, origin precedence (resolver origin beats detected type), HTTP failure classification.
- `classify_test.go`: full taxonomy table including wrapped errors.

Stacked with the dd-source replay PR; scanner merges first.


MikaYuoadas (approved):
LGTM 👍

PR #375, opened 2026-09-18
Retire --x-local-module-eval flag now that local module evaluation is unconditionally enabled

whitemerch:
## Motivation

Terraform local module evaluation shipped unconditionally in PRs #367 and #368 (CLI default and server mode). The `--x-local-module-eval` flag, its feature-flag constant `IacEnableLocalModuleEval`, and the remote-killswitch gate in the engine are now dead code.

## Changes

**Flag retirement**

`--x-local-module-eval` is moved to `retiredScanOnlyFlags()` alongside `--timeout`. Passing it logs a deprecation warning and is otherwise a no-op. A `retiredBool` helper is added to `retired_flags.go` to mirror the existing `retiredInt`.

**Engine gate removed**

The conditional in `Inspector.Inspect` that called `EvaluateWithOrg(IacEnableLocalModuleEval)` before running `instantiateLocalModules` is removed. Local module instantiation now runs unconditionally whenever `shouldInstantiateLocalModules` is true.

**Feature flag constant removed**

`IacEnableLocalModuleEval` is removed from `featureflags/evaluator.go` and its default entry from `local_evaluator.go`. The `serverFlagEvaluator` comment and override entry that pinned it `true` are removed from `analyze.go`.

**Tests and helpers cleaned up**

`TestLocalModuleEvalEnabled` and the `localModuleEvalEnabled()` helper in `scan.go` are deleted. The `moduleEvalEnabled()` test helper is removed from the engine integration tests; all call sites are replaced with `featureflags.NewLocalEvaluator()`. Stale `// CxSAST query under review` comments in `helpers.go` and `service.go` are dropped.

## QA Instruction

```
go test ./cmd/scanner/ ./pkg/engine/ ./pkg/server/ ./pkg/scan/ ./pkg/runner/ -count=1
```

All 6 packages pass. Local module evaluation behaviour is unchanged — the feature runs unconditionally as it has since main was cut.

## Impact

CLI and server scan paths are unaffected. Any caller passing `--x-local-module-eval` (or `--x-local-module-eval=false`) now receives a deprecation warning instead of silently toggling behaviour.


barbiergaspard (approved):
LGTM

PR #378, opened 2026-09-21
feat: render pushed helm charts in the serve path (IDE helm support)

whitemerch:
## Motivation

Helm is the most widespread way to define Kubernetes manifests, yet the IDE extension never reported findings from a chart: the serve endpoint it talks to had the Helm resolver pinned off, because chart rendering only read the real filesystem, which content-push mode cannot materialize. This PR closes that gap, a chart pushed as plain files now renders and scans in serve mode, the same way the CLI has always scanned it.

## Changes

- Helm charts load through the scan filesystem instead of the real disk, so the request's in-memory chart content renders with no disk access. The CLI keeps its exact loader and behavior.
- The serve path discovers pushed charts, renders each one once, and anchors findings on the template files (with template-mapped lines). A chart that fails to render, usually an incomplete push, asks the IDE for the rest of the chart via `missing_files`.
- Binary chart dependencies (`charts/*.tgz`) ride the wire base64-encoded, so packaged subcharts render too.
- `supported-files` advertises `Chart.yaml` with a new `chart` strategy, which tells the IDE to push a whole chart directory.

## Author Checklist

- [x] I have added or updated relevant unit tests where necessary. If no tests were added, I've explained why.

## QA Instruction

```
go test ./... -count=1
golangci-lint run -c .golangci.yml --timeout 20m
```

By hand: build, run `serve`, and POST a chart as plain files to `/ide/v1/iac/analyze` with k8s rules. What to look for: a finding whose `FileName` is the pushed template path (e.g. `chart/templates/deployment.yaml`) with a line mapped back into the template not one on the rendered manifest. A CLI scan over a directory containing charts produces the same findings as before this PR.

## Blast Radius

- CLI: no behavior change.
- Serve/IDE: pushed charts now render and report; supported-files gains one additive entry.
- No rules, libraries, or engine changes.

## Additional Notes

- Companion extension PR: https://github.com/ddoghq/datadog-vscode/pull/2524 (chart strategy on the IDE side; degrades gracefully against a binary without these changes).
- Activation for IDE users requires a scanner binary release containing this PR.
- Charts larger than the IDE's 200-file request cap are truncated and escalation cannot recover them.


pkg/engine/provider/filesystem.go
```diff
@@ -363,11 +363,13 @@ func (s *FileSystemSourceProvider) BuildInventoryFromPrebuilt(ctx context.Contex
...
 
-func isUnderChartRoot(path string, chartRoots []string) bool {
+// IsUnderChartRoot reports whether path lies at or below one of chartRoots.
+// A "." root covers everything: a chart at the workspace (or scan) root.
+func IsUnderChartRoot(path string, chartRoots []string) bool {
 	path = filepath.ToSlash(path)
 	for _, root := range chartRoots {
 		root = filepath.ToSlash(root)
-		if path == root || strings.HasPrefix(path, root+"/") {
+		if root == "." || path == root || strings.HasPrefix(path, root+"/") {
```
robertohuertasm-datadog: **blocking**

A pushed `Chart.yaml` becomes chart root `"."` (`MemorySourceProvider.ChartRoots` uses `filepath.Dir`). After a successful render, `PrepareMemorySources` (`prepare.go:132`) skips every path this helper accepts, and `root == "."` accepts every path. `filesystem_test.go:814` locks that in for `infra/main.tf`.

`dispatchMemoryChart` passes `afterResolve: nil`. `dispatchChart` in the same file excludes `resFiles.Excluded` instead, so only chart files drop out of the disk walk.

On the extension side, `groupIacScanPaths` puts a workspace-root chart over the whole workspace. The grouping test expects `infra/main.tf` in that chart group and in no directory group (`grouping.test.ts`, "groups a workspace-root chart over the whole workspace"). The interactive path walks up to that same root (`findEnclosingChartRoot`) and pushes the chart together with the open file. I traced this against `chakib.hamie/iac_helm_chart_support`. I did not run the two binaries together.

Those directory-strategy files (Terraform, raw YAML, Bicep) would come back with no findings, and the IDE would treat the targets as clean. Nested charts (`chart/...`) stay inside their prefix, so this is the root-chart case.

The IDE [review on datadog-vscode#2524](https://github.com/ddoghq/datadog-vscode/pull/2524#discussion_r4073093955) asks to leave `.tf`, `.tfvars` and `.bicep` in their own directory groups, and not to replace them with the chart when the file is opened. A CLI scan of a repo with `Chart.yaml` at the root still hits this skip, because `"."` covers every path. Shallow-first rendering should collapse a nested chart into its parent, not drop Terraform that sits next to a root `Chart.yaml`.

Should a successful root render skip only the Helm excluded set, and still dispatch `infra/main.tf`?

pkg/server/analyze_test.go
```diff
@@ -165,6 +229,258 @@ func TestAnalyze_ContentPush_TerraformFinding(t *testing.T) {
...
+		},
+		Ruleset:  ruleset(syntheticK8sRule()),
+		Platform: []string{"kubernetes"},
+	}
+
+	out, _ := postAnalyzeK8s(t, s, req)
+
+	var found bool
+	for _, f := range out.Findings {
+		if f.QueryID == syntheticK8sRuleID && f.FileName == "chart/charts/sub/templates/service.yaml" {
```
robertohuertasm-datadog, line 336: **blocking**

This assertion pins findings on `chart/charts/sub/templates/service.yaml`, which is never a path in the request (only `chart/charts/sub.tgz` is pushed). I read the companion extension in [datadog-vscode#2524](https://github.com/ddoghq/datadog-vscode/pull/2524): the batch provider drops findings whose anchor is not in `sentPaths` (`iacScanProvider.ts`, the `unattributed` filter), and interactive code bits keep a finding only when the anchor equals the open file. An unpacked `charts/sub/**` on disk still attributes, because those paths are real files. This is the `.tgz`-only case, which is what `helm dependency update` leaves behind. I did not run the two PRs together.

The IDE cannot guess the archive from the extracted path. `sub.tgz` lines up with `charts/sub/...` in this test. `nginx-1.2.3.tgz` does not, when the chart inside is named `nginx`. Attaching every such finding to line 1 of the parent `Chart.yaml` would stack unrelated comments on a file people edit.

The IDE review asks to report the finding on the `.tgz` that was pushed, with the inner path and line kept in the message, so Problems and File Insights can show it. A squiggle inside the archive is not possible. That remap needs this side to say which pushed file the finding was rendered from.

Would the scanner also return the `.tgz` path, and keep `FileName` as the path inside the archive? The IDE can then attach the diagnostic to that `.tgz` instead of dropping it.

pkg/runner/prepare.go
```diff
@@ -71,13 +88,98 @@ func PrepareSharedWalk(ctx context.Context,
...
+	files := mp.EligibleFiles(union)
+
+	renderedRoots := make([]string, 0)
+	for _, root := range mp.ChartRoots(union) {
+		if provider.IsUnderChartRoot(root, renderedRoots) {
+			continue
+		}
+		if dispatchMemoryChart(ctx, mp, services, root, scanID, openAPIResolveReferences, maxResolverDepth) {
+			renderedRoots = append(renderedRoots, root)
+		}
```
robertohuertasm-datadog: **blocking**

On a successful render we add the parent to `renderedRoots`, which skips nested `Chart.yaml` roots and later skips dispatch for every path under that parent. Escalation (`RecordMissing`) only runs when `Resolve` returns an error.

`TestResolveMemFSPartialChartRendersWithoutSpuriousEscalation` pins the other half: a chart that still renders after subchart templates are deleted records no `missing_files`. I did not push a chart that was cut off mid-tree.

The IDE review on [datadog-vscode#2524](https://github.com/ddoghq/datadog-vscode/pull/2524) asks to stop splitting charts at 200 files and send the whole tree in one request. The scanner's own file cap is 50,000. The limit that remains on the wire is the JSON body, 32 MiB (`--max-request-mib`). I looked in Datadog over the last month and found neither `error while analyzing IaC files` nor a 413, so I have not seen that body cap hit. A push can still be incomplete for other reasons (a dropped `charts/` file, a body the server refuses). Parent templates often remain, so Helm still renders, nested charts are not rendered on their own, and the IDE has nothing to escalate.

The PR body says an incomplete push that fails to render asks the IDE for the rest of the chart. A render that succeeds from a partial tree does not, so the result looks clean. Should a chart that rendered from an incomplete push still escalate (or fail) so the IDE can ask for the rest?

pkg/scanner/scanner.go
```diff
@@ -54,6 +54,27 @@ func PrepareAndScan(
 		return StartScan(ctx, scanID, services)
 	}
 
+	// The in-memory (server) dispatch: one pass over the pushed files, rendering
+	// each pushed Helm chart once. It runs regardless of the parallel-parsing
+	// flag — the flag only fans the file dispatch out — so charts never silently
+	// fall back to raw-template scanning.
+	if mp, ok := runner.SharedMemoryProvider(services); ok {
```
robertohuertasm-datadog, line 56: **non-blocking**

A scan prepares files in two places, and both are meant to do the same job.

`PrepareSharedWalk` is the disk path (the CLI). It lists the files, renders each Helm chart once, and does not scan that chart's templates again as raw YAML.

`PrepareMemorySources` is the new path for files the IDE pushed, and this branch is what calls it. It repeats that loop: find chart roots, render the outer chart first, skip files the render already covered, then hand the rest to the parsers. It had to be separate because `MemorySourceProvider.GetSources` never runs the Helm resolver, so the disk function could not be reused as it stands.

Would it be simpler to have one prepare function for both? The memory provider would return the same kind of file list `BuildInventoryFromPrebuilt` already builds, and disk and IDE would differ only in where the bytes are read from. A Helm change would then land in one place.

The two loops already disagree, which is the part I would rather not leave split even if a shared function is too much for this PR:

- On disk, `dispatchChart` drops only `resFiles.Excluded`. On this path, `dispatchMemoryChart` has no such callback, and `IsUnderChartRoot` treats a root of `"."` as every file (`filesystem.go:372`). A `Chart.yaml` at the workspace root then also drops `infra/main.tf`.
- `dispatchFile` keeps the parser for that file's platform (`prepare.go:255`). `dispatchMemoryFile` sends the file to every parser that claims the extension (`prepare.go:219`).

Totally understand if folding the two functions together is out of scope here. If so, would you still want the IDE loop to exclude the same files and pick the same parser as the disk loop?

pkg/runner/prepare.go
```diff
@@ -103,6 +208,41 @@ func dispatchChart(ctx context.Context,
...
+// parser supports its extension — dispatchFile's in-memory counterpart.
+func dispatchMemoryFile(ctx context.Context,
+	mp *provider.MemorySourceProvider,
+	services []*Service,
+	filePath, scanID string,
+	openAPIResolveReferences bool,
+	maxResolverDepth int) error {
+	if len(services) == 0 {
+		return nil
+	}
```
robertohuertasm-datadog, line 181: **non-blocking**

`.yaml` is claimed by two parsers, the general YAML one (Kubernetes, CloudFormation, Ansible, and the rest) and the CI one. `dispatchFile` keeps only the parser for the file's platform (`prepare.go:255`). This function hands the file to every parser that lists the extension. I did not produce a wrong-platform finding from that.

Copying the `servicesForPlatform` line on its own would not change anything here. In content-push, `initScan` leaves `FilePlatform` empty (`scan.go:60`), and `servicesForPlatform` returns every service when the platform is empty (`prepare.go:322`). The disk filter only works because the CLI walk fills that map first. Server mode skips that walk on purpose: `analyzer.Analyze` opens files with `os.Open`.

`analyzer.ClassifyFile` already does the per-file half of that, and it takes the in-memory FS plus the raw bytes (`analyzer.go:792`). The IDE already sends both. No new field from the extension.

Would it be worth, before this dispatch, classifying each pushed file with `ClassifyFile`, storing the result on the `FilePlatform` map the services already carry, and then calling `servicesForPlatform` here? That is a pass over the pushed files plus this filter. I would leave `analyzer.Analyze` off.

What that changes is limited to `.yaml` and `.yml`. Every other extension in this server set already belongs to one parser, so `.tf`, `Dockerfile`, and `.bicep` are unaffected.

Each YAML file is parsed twice today, and the CI parser then also walks `run` blocks and `${{ }}` expressions. A chart push is mostly YAML, so that second pass is the cost that would drop. I have not timed it. Query eval may still dominate, and the old per-service server path already parsed YAML twice, so this is not a regression of this PR.

The two parses are stored as separate files (`FileMetadatas.ToMap` keys by id, and `Combine` keeps both). I have not seen a rule fire twice on the same path. If one does, the filter would drop that second copy.

The risk is a wrong `ClassifyFile` result: the parser that should have seen the file never does, and a finding disappears. An empty classification keeps both parsers, which is today's behavior.

One gap in that helper: `checkHelm` still `os.Stat`s `Chart.yaml` (`analyzer.go:915`), so this pass will not see a chart that exists only in the push. The split above is CI versus the other YAML platforms, which `classifyByContent` decides from the bytes.

If the single prepare in `scanner.go:61` lands, this filter belongs in that one dispatch, still fed by `ClassifyFile` on the pushed bytes. Totally understand if the double parse is the server behavior you want to keep.

pkg/scan/scan.go
```diff
@@ -379,10 +379,11 @@ func (c *Client) createService(
...
 	}
 
-	// combinedResolver to be used to resolve files and templates
-	builder := resolver.NewBuilder()
+	// The helm resolver reads chart files through the scan FS, so server-mode
+	// requests render pushed chart content without disk access.
+	builder := resolver.NewBuilder().WithFS(c.fsys)
 	if flagEvaluator.EvaluateWithOrg(featureflags.IacEnableKicsHelmResolver) {
-		builder = builder.Add(ctx, &helm.Resolver{})
+		builder = builder.Add(ctx, helm.NewResolver(c.fsys))
```
robertohuertasm-datadog, line 386: **question**

`WithFS(c.fsys)` is used by `GetType` to sniff `Chart.yaml`, and `helm.NewResolver(c.fsys)` is used to load the chart. I have not hit a mismatch: both call sites currently pass the same FS.

I'm wondering whether `GetType` could live on `helm.Resolver` so there is a single injection site. Passing `WithFS(memfs)` with `&helm.Resolver{}` would classify from memory and render from disk. Or is the split intentional because `GetType` has to run even when the Helm flag is off?

pkg/resolver/helm/helm.go
```diff
@@ -656,10 +743,18 @@ func crdChartRelativePath(name string) string {
...
+// resolvedChartFilePath maps a chart-relative path to a path beside chartPath.
+// The output keeps the chart path's separator style: a pushed (content-push)
+// chart path is forward-slash, and its resolved files must be too so findings
+// match the pushed path shape on every platform; a native disk chart path
+// (the CLI on Windows) keeps its native separators.
 func resolvedChartFilePath(chartPath, chartRelative string) string {
 	subFolder := filepath.Base(chartPath)
-	return filepath.Join(filepath.Dir(chartPath), subFolder, filepath.FromSlash(chartRelative))
+	joined := filepath.Join(filepath.Dir(chartPath), subFolder, filepath.FromSlash(chartRelative))
+	if !strings.Contains(chartPath, `\`) {
```
robertohuertasm-datadog: **nit**

A rendered finding has to use the same separators as the path that was scanned, or the IDE will not match it to the file it pushed. The IDE always sends forward slashes (`chart/templates/deployment.yaml`), including on Windows. The CLI on Windows keeps backslashes. `filepath.Join` on Windows emits backslashes, so this function forces `filepath.ToSlash` when `chartPath` contains no `\`.

That treats "no backslash in the string" as "this chart was pushed". It matches what we send today, and I have not seen a mismatch. A disk path that is already slash-separated, or a pushed path that contains a backslash, would take the other branch.

`NewResolver` already knows which case it is: the server passes a `*vfs.MemFS`, the CLI passes `vfs.DiskFS`. Would it be clearer to remember that there, and have this function take the flag instead of inspecting the string?

```go
type Resolver struct {
	fsys       vfs.FS
	slashPaths bool // pushed charts: always "/". CLI: the OS separator.
}

func NewResolver(fsys vfs.FS) *Resolver {
	_, pushed := fsys.(*vfs.MemFS)
	return &Resolver{fsys: fsys, slashPaths: pushed}
}

func resolvedChartFilePath(chartPath, chartRelative string, slashPaths bool) string {
	subFolder := filepath.Base(chartPath)
	joined := filepath.Join(filepath.Dir(chartPath), subFolder, filepath.FromSlash(chartRelative))
	if slashPaths {
		joined = filepath.ToSlash(joined)
	}
	return joined
}
```

The call in `resolver.go` (`resolvedChartFilePath(filePath, chartRelative)`) would pass `slashPaths` from the resolver. No new field on the IDE request. Take it or leave it.

pkg/engine/provider/memory.go
```diff
@@ -69,6 +69,37 @@ func (m *MemorySourceProvider) eligibleFiles(extensions model.Extensions) []stri
...
+
+// ReadFile reads a pushed file through the provider's FS.
+func (m *MemorySourceProvider) ReadFile(p string) ([]byte, error) {
+	return m.fsys.ReadFile(p)
+}
+
+// RecordMissing records an escalation request when the provider's FS tracks
+// missing paths (the in-memory one does; the real disk cannot).
+func (m *MemorySourceProvider) RecordMissing(path string) {
+	if r, ok := m.fsys.(interface{ RecordMissing(string) }); ok {
```
robertohuertasm-datadog: **nit**

When a pushed chart fails to render, the server asks the IDE for the rest of that directory. That list lives on `MemFS.RecordMissing`. The CLI disk has no such list: a missing file is just missing. `MemorySourceProvider` stores the filesystem as `vfs.FS`, which only reads (`ReadFile`, `Stat`, and the rest), so `RecordMissing` is not on it. This asks at the call:

```go
if r, ok := m.fsys.(interface{ RecordMissing(string) }); ok {
	r.RecordMissing(path)
}
```

`MemFS` has the method, so the path is recorded. `DiskFS` does not, so the check fails and nothing happens. That is the right split, and I am not asking to change it.

The interface has no name, so there is nothing to search for except the method text. Would a named interface be easier to follow?

```go
// vfs
type MissingRecorder interface {
	RecordMissing(name string)
}
```

`MemFS` already has `RecordMissing`, so it would satisfy this with no change to its body. `DiskFS` would not. The call becomes:

```go
if r, ok := m.fsys.(vfs.MissingRecorder); ok {
	r.RecordMissing(path)
}
```

Same behavior. Take it or leave it.

robertohuertasm-datadog (commented):
I left a few comments inline for your consideration.

The three I would want to look at before merge: a workspace-root `Chart.yaml` skipping co-pushed files such as `infra/main.tf`, findings anchored on a path inside a `.tgz` that was never pushed, and a chart that still renders from an incomplete push so the IDE is never asked for the rest. I traced these against the companion extension. I did not run the two binaries together.

Two more are optional. One is classifying a pushed `.yaml` before it goes to every parser that claims the extension. The other is whether the memory prepare loop and the disk one could be a single function. The last two, path separators and a name for `RecordMissing`, are take it or leave it.

Happy to be told any of these are out of scope, or that I missed some context.

pkg/engine/provider/filesystem.go
whitemerch: Fixed, but not by skipping only Excluded. Helm's loader puts every non-ignored file under the chart root into `chart.Raw`, and `getExcluded` returns all of them, so Excluded would still hide infra/main.tf. Now a rendered chart withholds only its Helm files (Chart.yaml, Chart.lock, values*.yaml, values.schema.json, requirements.*, templates/, crds/, charts/), through one helper used by the in-memory dispatch and all three CLI walks. The CLI had the same gap through SkipDir 😢 , so this also changes CLI results for repos with IaC next to a chart. Nested charts are now skipped only under a parent's charts/; a standalone chart elsewhere under a rendered chart is rendered on its own

pkg/runner/prepare.go
whitemerch: Yes the scanner can't tell a partial chart from a complete one, so this is fixed on the extension side. Chart pushes are no longer capped anywhere (grouping, chart listing, batch fill, escalation merge, interactive cap), so a chart is always sent whole and the request body limit is the only bound. The description overstated the escalation behavior; I've corrected it

pkg/server/analyze_test.go
whitemerch: Agreed, the response now includes archive_files, mapping each such finding path to the pushed charts/*.tgz (chart name from the archive, so nginx-1.2.3.tgz → charts/nginx/). The extension uses that in iacFindingAnchorFile / withIacArchiveLocation so the IDE anchors on the .tgz at line 1

pkg/scanner/scanner.go
whitemerch: disk and pushed content share prepareSources with a small preparedSource (diskSource / memorySource) for list/read/route/chart callbacks. PrepareSharedWalk and PrepareMemorySources are thin wrappers

pkg/runner/prepare.go
whitemerch: for pushed files, memorySource.platform calls analyzer.ClassifyFile before servicesForPlatform, so YAML goes only to the platform’s parser. We skip classification when only one parser handles the extension (e.g. .tf)

pkg/scan/scan.go
whitemerch: Moved onto helm.Resolver.GetType reading through the same FS as render, and i removed Builder.WithFS. When the Helm resolver isn’t registered, GetType stays KindCOMMON. same effective behavior as before with the flag off 👍

pkg/resolver/helm/helm.go
whitemerch: Replaced the “no \ in path” heuristic with an explicit flag derived from !vfs.IsDisk(fsys) when resolving paths, so pushed (MemFS) always uses / and CLI disk keeps OS separators

pkg/engine/provider/memory.go
whitemerch: Added vfs.MissingRecorder, so MemorySourceProvider.RecordMissing type-asserts to that instead of an inline interface

pkg/server/archive_sources.go
```diff
@@ -0,0 +1,114 @@
...
+	for i := 0; i+2 < len(segs); i++ {
+		if segs[i] != chartsDirName {
+			continue
+		}
+		if _, unpacked := dirs[strings.Join(segs[:i+2], "/")]; unpacked {
+			continue
+		}
+		chartsDir := strings.Join(segs[:i+1], "/")
+		for _, archive := range archivesByDir[chartsDir] {
+			if chartName(archive) == segs[i+1] {
```
robertohuertasm-datadog: **blocking**

`sourceArchive` treats the packaged chart's `Chart.yaml` name as the directory Helm rendered under. A dependency alias breaks that: Helm rewrites the subchart name before it emits `# Source:`.

I reproduced this with Helm v3.18.4. Parent chart dependency `name: redis`, `alias: cache`, archive already at `charts/redis-0.1.0.tgz`:

```
# Source: parent/charts/cache/templates/svc.yaml
```

`helm show chart` on that archive still reports `name: redis`. `ArchiveChartName` (`pkg/resolver/helm/helm.go:181`) returns that name via `loader.LoadArchive`, and the comparison here is `chartName(archive) == segs[i+1]`, so `redis == cache` fails. The function then returns `""` on that first `charts/` hit (`pkg/server/archive_sources.go:111`), and `archive_files` gets no entry.

The IDE anchors a packaged-subchart finding on `archiveFile` (`iacFindingAnchorFile`). Without it, the rendered path was never pushed (only the `.tgz` was) and `iacScanProvider.ts` drops the finding as unattributed. I did not POST this chart through `/ide/v1/iac/analyze`. The miss follows from that equality once the source path uses the alias. The non-alias case (`nginx-1.2.3.tgz` rendering under `charts/nginx/`) still matches, which is what the tests cover.

Would it be worth recording the archive path when the chart is loaded, under the name Helm actually renders (the alias, once dependencies have been processed), and copying that into `archive_files`? Happy to be told aliases are out of scope, though the drop is silent.

pkg/runner/prepare.go
```diff
@@ -103,26 +168,64 @@ func dispatchChart(ctx context.Context,
...
+}
+
+// diskSource is the CLI's source: the walked (or analyzer-prebuilt) inventory,
+// read from disk or the analyzer's content cache, routed by the analyzer's
+// per-file platform.
+type diskSource struct {
+	*provider.FileSystemSourceProvider
+}
+
+func (d diskSource) chartRendered(ctx context.Context, chartPath string, resFiles model.ResolvedFiles) {
```
robertohuertasm-datadog: **non-blocking**

`diskSource.chartRendered` still registers an exclude list, while the inventory walk already drops those same paths with `IsHelmChartFile` (`pkg/engine/provider/filesystem.go:332`). `memorySource.chartRendered` is a no-op for that reason. `HelmChartFiles` only keeps paths `IsHelmChartFile` would skip, so this exclude looks like a subset of the new policy.

`resolveChartDir` does the same pair: `ExcludePaths` at `pkg/engine/provider/filesystem.go:740`, then it appends the chart root that `IsHelmChartFile` consults.

Would it be simpler to make the disk callback a no-op too, and drop the `ExcludePaths` call in `resolveChartDir` if nothing else reads that map for Helm files? I could be missing a walk that still depends on it.

pkg/server/analyze_test.go
```diff
@@ -165,6 +229,326 @@ func TestAnalyze_ContentPush_TerraformFinding(t *testing.T) {
 	}
 }
 
+// TestAnalyze_ContentPush_HelmChartFinding verifies the full content-push helm
+// path: a chart pushed as plain files renders through the in-memory FS, its
+// manifests are scanned with the Kubernetes rules, and the finding anchors on
+// the pushed template path.
+func TestAnalyze_ContentPush_HelmChartFinding(t *testing.T) {
```
robertohuertasm-datadog: **nit**

Take it or leave it. `analyze_test.go` was already 1365 lines on `main` and this branch takes it to 1750. Most of the growth is the Helm content-push cases that start at this test (the insertion here is about 320 lines).

Would it be worth moving the Helm, archive, and base64 cases next to `archive_sources.go`, for example `analyze_helm_test.go`, and leaving this file for the HTTP and lifecycle tests?

pkg/server/archive_sources.go
whitemerch: Yeah, that’s the bug we were keying off the name inside the .tgz, but Helm renders under the alias, my bad

pkg/runner/prepare.go
whitemerch: Yes! dropped chartRendered and the resolveChartDir ExcludePaths call

pkg/server/analyze_test.go
whitemerch: Moved the Helm, archive, and base64 content-push tests into analyze_helm_test.go so this file stays focused on the HTTP and general analyze paths

pkg/runner/prepare.go
```diff
@@ -86,11 +150,9 @@ func dispatchChart(ctx context.Context,
 		for _, s := range services {
 			s.logResolverResolveError(ctx, kind, chartPath, err)
 		}
+		src.chartFailed(chartPath)
```
robertohuertasm-datadog: **blocking**

The chart is already fully pushed. Rendering fails because a deploy-time value is absent (`required`, or an unset `.Values` key). `src.chartFailed` still adds that chart directory to `missing_files`, so the IDE is asked to send the same folder again. Sending it again cannot supply the missing value.

`missing_files` is documented as paths the engine referenced that were not pushed (`analyzeResponse` in `pkg/server/analyze.go`, and `MemFS.MissingFiles`). `isExpectedHelmRenderError` already knows these failures (`nil pointer evaluating`, `map has no entry for key`, `can't evaluate field`, `error calling required:`), and on the memory path `chartFailed` still calls `RecordMissing(chartPath)` (`prepare.go` line 259). I traced this from `resolveAndStoreChart`. I did not execute a chart that fails this way.

With `Chart.yaml` at the workspace root, `chartPath` is `"."`, and the IDE reads that as the workspace folder (`toSafePathSegments` in the companion extension). It re-pushes the chart once (`escalateAnalyze` escalates each path at most once), then stops as `noProgress` and records the analysis as incomplete. A CLI scan leaves `diskSource.chartFailed` empty and keeps the raw templates, with no second request.

`TestAnalyze_HelmChartRenderFailureEscalates` only covers an `include` whose helper was never pushed. That error is outside the expected-error list, and asking for the folder there still makes sense.

Would it be worth skipping `chartFailed` for these expected value failures, and still returning false so the raw templates are scanned? 

```suggestion
		if kind != model.KindHELM || !isExpectedHelmRenderError(err) {
			src.chartFailed(chartPath)
		}
```

pkg/server/archive_sources.go
```diff
@@ -0,0 +1,168 @@
...
+			if err != nil {
+				continue
+			}
+			name, version, loadErr := helm.ArchiveChartIdentity(data)
+			if loadErr != nil {
+				contextLogger := logger.FromContext(ctx)
+				contextLogger.Debug().Msgf("could not read chart name from pushed archive %s: %v", archive, loadErr)
+				continue
+			}
+			out[chartsDir+"/"+renderedChartName(name, version, deps)] = archive
```
robertohuertasm-datadog: **non-blocking**

Two packaged copies of the same chart can sit in `charts/`, for example `nginx-1.16.0.tgz` (the one the parent depends on) and a leftover `nginx-1.2.3.tgz`. Helm renders the one whose version matches. `archive_files` is supposed to point the finding at that tarball, so the IDE can show it on a file that was actually pushed.

Both archives want the same map key, `charts/nginx`. The leftover fails `IsCompatibleRange` and still falls through to `return name` in `renderedChartName` (line 147), which is that same key. This assignment writes the key every time. `MemFS.Paths()` sorts with `sort.Strings`, and `nginx-1.16.0.tgz` comes before `nginx-1.2.3.tgz` (`1` before `2`), so the leftover is written last and replaces the match. The rendered chart is the right one. The IDE would anchor the finding on the leftover. I traced this from the sort and from `renderedChartName`. I did not run it.

Would it be worth writing that fallback name only when the key is still empty, so a match can replace a leftover and a leftover cannot replace a match? Two archives that both satisfy the range would still share the key. Is that narrower case worth a different key, or is the fallback guard enough?

pkg/server/archive_sources.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+	return out
+}
+
+func chartDependencies(memfs *vfs.MemFS, parent string) []chartDependency {
+	chartYAML := "Chart.yaml"
+	if parent != "." && parent != "" {
+		chartYAML = parent + "/Chart.yaml"
+	}
+	data, err := memfs.ReadFile(chartYAML)
```
robertohuertasm-datadog: **non-blocking**

A push that already produced findings, and that also includes a packaged chart such as `vendor/charts/dep.tgz` with no `vendor/Chart.yaml` next to it, asks the IDE to send `vendor/Chart.yaml`. The scan did not need that file. `missing_files` then reads as "please send this" for a path opened only while mapping archives.

`archiveSources` runs when the response has findings (`pkg/server/analyze.go`, the `ArchiveFiles` assignment). `renderedArchives` calls `chartDependencies`, which reads the parent `Chart.yaml` with `MemFS.ReadFile`. `ReadFile` records every miss for escalation (`pkg/vfs/memfs.go`, the `m.missing[key]` write). `Stat` does not. The chart loader already stats `.helmignore` before reading it, for this same reason (`pkg/resolver/helm/helm.go`, the `HelmIgnore` stat). I traced this in the code. I did not send a request that reproduces it.

The suggestion below replaces this read and the `err != nil` return under it (lines 127 to 130).

```suggestion
	if _, statErr := memfs.Stat(chartYAML); statErr != nil {
		return nil
	}
	data, err := memfs.ReadFile(chartYAML)
	if err != nil {
		return nil
	}
```

Would that `Stat` be enough? Or would it be simpler to record the rendered path to archive path while the chart renders, so this pass does not open `Chart.yaml` at all? Happy to be told a stray archive next to other findings is out of scope.

pkg/engine/provider/filesystem.go
```diff
@@ -605,33 +709,25 @@ func pathWithinBase(base, path string) bool {
...
+// the chart's Helm files are skipped for the rest of the walk so they are not
+// scanned again as raw, unrendered templates (which would yield bogus names like
+// name: {{ .Release.Revision }}); other files under the chart root still are.
 // On failure it returns nil to fall back to scanning the raw files.
 func (s *FileSystemSourceProvider) resolveChartDir(ctx context.Context, path string,
 	resolverSink ResolverSink, resolvedChartPaths *[]string) error {
 	contextLogger := logger.FromContext(ctx)
-	excluded, errRes := resolverSink(ctx, strings.ReplaceAll(path, "\\", "/"))
+	normPath := toSlash(path)
+	_, errRes := resolverSink(ctx, normPath)
```
robertohuertasm-datadog: **non-blocking**

After a chart renders, the walk already skips its Helm files with `IsHelmChartFile`. `ExcludePaths` is the old registration for that same skip, and nothing calls it. The method itself is not in this diff (`pkg/engine/provider/filesystem.go` line 147).

This line discards the sink's excluded paths. `resolveAndStoreChart` dropped the call that used to pass them in (the deleted `fsp.ExcludePaths(ctx, resFiles.Excluded)` lines in `pkg/runner/prepare.go`). A search for `.ExcludePaths(` finds no callers. The `Excluded` list is separate and still used: `getExcluded` fills it, and `pkg/resolver/helm/memfs_load_test.go` compares the disk render with the in-memory one. I did not re-run the Helm tests.

Would it be worth deleting `ExcludePaths` and leaving that list? Totally understand if the method stays as a hook.

robertohuertasm-datadog (dismissed):
LGTM! 🚀 Just a couple of minor comments for your consideration.

pkg/server/archive_sources.go
whitemerch: Stat + test, gift from me

pkg/engine/provider/filesystem.go
whitemerch: removed the unused ExcludePaths method ✨

robertohuertasm-datadog (approved):

